Issue #66, opened 2012-02-08
orientation adaptation


dhardy92:
Hello,

For tablets and smartphone it could be useful to make width and height adaptive against the image orientation.
The image must resize well if you turn the table or smartphone on the same orientation as the image.

So an image source with size of 80x40 with a resize request of /fit-in/30x60/ must not be outed as 30x15 but at 60x30

I suppose filter could not be used for that because it run after resize and maybe it require some change in the API like "/fit-in/o30x60/" (o for oriented :) or something else

Cheers,
## 

Damien

PR #67, opened 2012-02-10
/adaptive-fit-in/ fit image independently of its orientation landscape/portrait

dhardy92:
Hello,

This is a contrib in API to address the #66 issue
As you suggested I create a new entry in API : 'adaptative-fit-in'
witch do the same as 'fit-in' but try to invert requested width and height if it permit a better image definition.

I add a test case in FIT_IN_CROP_DATA to test it.
## Cheers,

Damien


AizazAZ:
How to use it? Please give a URL example. Thanks :)

heynemann:
Exactly the same way you use fit-in, but with adaptive- in front of it. Check the usage page at https://github.com/thumbor/thumbor/wiki/Usage

AizazAZ:
In from it doesn't return anything, I used it at it's back. Let me know if I am doing it right

http://thumb.xufu.com/unsafe/adaptive-fit-in/350x100/smart/http://p1.lobrata.com/image/a/2015/04/2108382445_thumb.jpg

PR #321, opened 2014-07-09
Thumbor support for gif and animated gif images.

heynemann:
Supporting gifs and animated gifs by automating gifsicle.

Some things to improve here:

a) Use tornado.subprocess to asynchronously call gifsicle, thus releasing tornado's ioloop to keep procesing requests. The issue with this is we would have to change EVERY call to the engine's methods to be async (or coroutine);
b) Support filters;
c) Support face detection.

What do you guys think (@cezarsa, @robolson, @albohlabs, @allafa70, @rfloriano, @fabiomcosta, @guilhermef, @ollierattue)?

This implementation with gifsicle supports:
- Manual Crop
- Resizing
- Flip
- Rotation

Of both gifs and animated gifs. I've also found that the quality of the generated gifs is FAR superior to our current solution.

The current solution for gifs:

old

-rw-r--r--@   1 heynemann  staff   578K  9 Jul 15:37 old.gif

The new solution result:

proper

-rw-r--r--@   1 heynemann  staff   511K  9 Jul 15:35 proper.gif

As you can see, the new solution is far superior (and smaller).

Hope to get some comments so we can merge this to the master branch. Also worth noting that the USE_GIFSICLE_ENGINE default is False, thus this change is backwards-compatible.

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -37,7 +37,7 @@ class Context:
 
 
 class ServerParameters(object):
-    def __init__(self, port, ip, config_path, keyfile, log_level, app_class, fd=None):
+    def __init__(self, port, ip, config_path, keyfile, log_level, app_class, fd=None, gifsicle_path=None):
         self.port = port
         self.ip = ip
         self.config_path = config_path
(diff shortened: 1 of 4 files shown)


guilhermebr:
Great! worked \0

PR #322, opened 2014-07-10
Add `full-fit-in` feature

moeffju:
Refresh of #243 by @gockxml. Similiar to fit-in, while scaling the image to exactly fill a given box to the full without cropping.

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -84,7 +84,7 @@ class RequestParameters:
                  crop_bottom=None,
                  crop=None,
                  adaptive=False,
-                 full = False,
+                 full=False,
                  fit_in=False,
                  width=0,
                  height=0,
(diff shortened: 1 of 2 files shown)


victornpb:
I just think `full-fit-in` is a awfully named feature.

I think the motivation for "fit-in" is supposed to mean fit in the canvas W x H
Why not `fit-out` fit on outer boundaries of the canvas.

Or just `full-fit` or `fit-full` or just `full` or just `fit`. So many options.

Or follow the css spec background-sizing: `cover`; should be more intuitive for developers.

I had a really bad time trying to find this full-fit-in trying every possible keyword I could think of, it isn't even documented anywhere.

heynemann:
Good points @victornpb 
Can you help us out and contribute the docs for this feature?

PR #376, opened 2014-10-23
PIL: convert palette to RGBA before resize()

pda:
Before resizing 8-bit indexed images, convert to RGBA to avoid the terrible results of nearest-neighbour resampling:

> The filter argument can be one of NEAREST, BILINEAR, BICUBIC, or ANTIALIAS. **If omitted, or if the image has mode “1” or “P”, it is set to NEAREST.**
— http://effbot.org/imagingbook/image.htm#tag-Image.Image.resize

Thumbor's PIL engine does specify `ANTIALIAS`, but as stated above, PIL disregards that for `1` and `P` images.

This patch converts indexed `P` images before resize, but not 1-bit black-and-white `1` images. This is partly because I'm not sure if antialiasing is desirable given a 1-bit image (1-bit is uncommon, and perhaps it was 1-bit for a reason?), and partly because it's irrelevant to my use-case. I'm happy to add `1` images too.

Note that the output PNG was RGBA anyway, but that conversion was happening *after* the resize, resulting in worst of both worlds — nearest neighbor resampling of a highly constrained palette for poorest image quality, and then re-encoded into 32-bit color space for largest file size.

Here's the before and after of the actual image thumbnail that flagged this issue:

palette rgb

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -68,6 +68,9 @@ class Engine(BaseEngine):
         del d
 
     def resize(self, width, height):
+        if self.image.mode == 'P':
+            logger.debug('converting image from 8-bit palette to 32-bit RGBA for resize')
+            self.image = self.image.convert('RGBA')
         self.image = self.image.resize((int(width), int(height)), Image.ANTIALIAS)
 
     def crop(self, left, top, right, bottom):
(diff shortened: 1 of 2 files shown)


heynemann:
Sorry for taking so long to merge this.

PR #395, opened 2014-12-11
Add CUSTOM_HANDLERS

wandenberg:
The idea is to make possible add new routes to the default ThumborServiceApp instead of replace it for a new one using a simple configuration.

The custom handler class must be called HttpHandler and has a static method called url_regexp which has to return the url pattern to be associated with this handler.

To enable the handler is necessary to set it on the thumbor.conf file like
CUSTOM_HANDLERS = [ 'module.where.the.handler.is' ]

Is possible to set multiple handlers, they will be added in order to the top of thumbor app.

### DIFF
--- a/thumbor/app.py
+++ b/thumbor/app.py
@@ -30,6 +30,14 @@ class ThumborServiceApp(tornado.web.Application):
             (r'/healthcheck', HealthcheckHandler),
         ]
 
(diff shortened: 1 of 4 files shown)


heynemann:
As was explained by @guilhermef we won't be merging this PR.

PR #396, opened 2014-12-11
Cleanup handler

wandenberg:
Using the CUSTOM_HANDLERS configuration, this optional handler make possible remove the original and resultants images from the respective storages.

To remove an original image a DELETE must be done to a signed url following the pattern
/original/<delete hash>/<url containing the original image>

To remove a resultant image a DELETE must be done to a signed url following the pattern
/result/<delete hash>/<url of result image>

The delete_hash is obtained using the security_key to sign a url composed by the desired delete type and the target url, which is the same as the previous examples just removing the "/<delete_hash>"

### DIFF
--- a/thumbor/app.py
+++ b/thumbor/app.py
@@ -30,6 +30,14 @@ class ThumborServiceApp(tornado.web.Application):
             (r'/healthcheck', HealthcheckHandler),
         ]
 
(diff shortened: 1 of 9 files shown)


heynemann:
Same as the other PR.

PR #403, opened 2014-12-23
Add Statsd facilities for better instrumentation

clifff:
screen shot 2014-12-23 at 12 21 23 pm
_Part of our statsd driven Thumbor dashboard_

As Vox Media has scaled out our Thumbor cluster, we've had to troubleshoot some performance issues around our storage and loaders. We have been using statsd for this kind of performance monitoring for a while, so it was a natural choice when we wanted to get some more visibility into Thumbor internals - now, we have a set of statsd dashboards that let us know exactly how things are running.

This PR adds a stastd_client to the Context object, and a few starting hooks that were useful to us. For now, it is mostly instrumenting the HTTP loader and result storage. The default behavior of the `stastsd_client` object is just to write to the debug log, and it will only actually send UDP data out if you set a `STATSD_HOST` value in your config. 

I also added the HTTP_LOADER_CURL_ASYNC_HTTP_CLIENT config option because the curl async client provides a lot more timing data that is very useful to log.


heynemann:
I can't begin to tell you how happy I am to merge this! :)

@clifff would you be willing to write the docs on this on the wiki? If you can't I'll try to write it later on.

Thanks once again for being a valuable part of thumbor community.

PR #414, opened 2015-02-11
[feature] id max lenth should be an option

dhardy92:
We use plain uuid with dash '-' but thumbor truncate it as non dash uuid length (32 char)
This is a new config entry to allow changes: MAX_ID_LENGTH


dhardy92:
ping @ncolomer

dhardy92:
@heynemann any thought on this ?
I suppose everybody is dancing at carnaval ;)

heynemann:
Sounds sensible. Backwards compatible as well. Merging.

ncolomer:
Thanks a lot @heynemann!

Issue #377, opened 2014-10-23
Optional syslog logging


wking:
I'm not sure that this has been submitted upstream (i.e. here), but @miracle2k has a [patch][1] by @elprans that swaps in Python's `SysLogHandler` based on a `THUMBOR_USE_SYSLOG` environment variable.  The patch seems like it's [also under the MIT license][2], so we should be able to include it without trouble.  It would be nice to get something like that incuded by default, although I'd probably treat the environment variable as a string which would be:

* empty if you don't want syslog handling
* `tcp://HOST:PORT` or `udp://HOST:PORT` if you want to talk to the syslog daemon over the network
* anything else would be interpreted as the absolute path to the syslog Unix socket.

[1]: https://github.com/miracle2k/dockerfiles/blob/master/thumbor/setup/patches/thumbor/0001-Add-syslog-support.patch
[2]: https://github.com/thumbor/thumbor/blob/master/LICENSE

heynemann:
@wking any news on this?

wking:
On Mon, Jan 12, 2015 at 04:20:18AM -0800, Bernardo Heynemann wrote:
> @wking any news on this?

Ah, sorry, it had slipped off my radar.  I'll queue it for the
upcoming weekend.

guilhermef:
I've released the 4.9.0 with @dhardy92 pull request.

wking:
On Thu, Feb 19, 2015 at 10:09:10AM -0800, Guilherme Souza, A.k.a.: Galinho wrote:
> Closed #377.

Thanks :).

PR #415, opened 2015-02-13
[log] add logging configuration capabilities

dhardy92:
This permit a custom log policy declared in thumbor.conf
Much more flexible than proposed in #377
Full python logging capabilities is so available and extendable without new change on Thumbor code base.
https://docs.python.org/2/library/logging.config.html
log in a file, in syslog, tcp, udp, HTTP endpoint or even custom sink is now available ...

Here to write in a file : /var/log/thumbor.log 
```python
THUMBOR_LOG_CONFIG = {
              'version': 1,
              'disable_existing_loggers': False,
              'formatters': {
                'standard': {
                  'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
                },
              },
              'handlers': {
                'info_file_handler': {
                  'class': 'logging.FileHandler',
                  'level': 'DEBUG',
                  'formatter': 'standard',
                  'filename': '/var/log/thumbor.log',
                  'encoding': 'utf8'
                }
              },
              'root': {
                'handlers': ['info_file_handler'],
                'level': 'DEBUG'
              }
            }
```
Here is some how-to  http://victorlin.me/posts/2012/08/26/good-logging-practice-in-python

Even thumbor.error_handlers is not usefull anymore. you can write your own handler out of Thumbor and use it in configuration.

if THUMBOR_LOG_CONFIG is defined, THUMBOR_LOG_FORMAT, THUMBOR_LOG_DATE_FORMAT configs and -l parameter are not  used any more.

:warning: I don't know how to tests this as vows tests only applies at most in thumbor.app module not thumbor.server.
May I have some help on this point ?

### DIFF
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -16,6 +16,7 @@ from derpconf.config import Config
 
 from thumbor import __version__
 
+Config.define('THUMBOR_LOG_CONFIG', None, 'Logging configuration as json', 'Logging')
 Config.define(
     'THUMBOR_LOG_FORMAT', '%(asctime)s %(name)s:%(levelname)s %(message)s',
     'Log Format to be used by thumbor when writing log messages.', 'Logging')
--- a/thumbor/server.py
+++ b/thumbor/server.py
@@ -9,7 +9,7 @@
 # Copyright (c) 2011 globo.com timehome@corp.globo.com
 
 import sys
-import logging
+import logging, logging.config
 import os
 import socket
 from os.path import expanduser, dirname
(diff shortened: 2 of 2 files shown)


dhardy92:
I didn't noticed but ```logging.config.dictConfig()``` is only available since python2.7 /o\
We can use ```logging.config.fileConfig()``` but this implies another config file for logs.
What do you think ?

heynemann:
Hmm... Not sure if we need to be python 2.6 compatible anymore. Let's see what other committers think: @cezarsa @guilhermef @fabiomcosta @rafaelcaricio

guilhermef:
I think we could drop python 2.6 support.

rafaelcaricio:
From me we could just drop support to py2.6 too.

guilhermef:
@dhardy92 I've added python 2.6 support because we still had a few machines running it. I'll just do a commit removing it.

guilhermef:
Merged!
Thanks @dhardy92  ;)

PR #421, opened 2015-02-26
Changin the way image format is being checked so that it doesnt conflict...

krynble:
... with MPO format for JPG files


krynble:
Please review @guilhermef

guilhermef:
Simple but super effective ;)
Thank you @krynble

PR #422, opened 2015-03-02
[mime] mimetype should be mimetypes... and thumbor must handle Bad Content-Type from client

dhardy92:
From time to time, client try to upload image with : 
```Content-Type: image/jpeg;charset=UTF-8```
This result in 500 error : 
```
cannot concatenate 'str' and 'NoneType' objects
<type 'exceptions.TypeError'>
File "/usr/lib/python2.7/dist-packages/tornado/web.py", line 1021,
  in _execute getattr(self, self.request.method.lower())(*args, **kwargs) 
File "/usr/lib/python2.7/dist-packages/thumbor/handlers/upload.py", line 48, 
  in post filename = self.context.config.UPLOAD_DEFAULT_FILENAME + extension 
```
Method engine.get_mimetype() was used sometime to get mime and other to get extension.
I choose to fix it as it's named : return mime

Now GET /image should return correct Content-Type: 

BTW there is declared dependency on python-magic but used nowhere.
By the time I choose not using it to avoid new dependency regarding the few format supported.
Maybe we should use it now if not too heavy, as dependency already exists ?
This implies some refactoring to make the code base clearer.

### DIFF
--- a/thumbor/engines/__init__.py
+++ b/thumbor/engines/__init__.py
@@ -57,23 +57,23 @@ class BaseEngine(object):
 
     @classmethod
     def get_mimetype(cls, buffer):
-        extension = None
+        mime = None
 
         #magic number detection
         if buffer.startswith('GIF8'):
-            extension = '.gif'
+            mime = 'image/gif'
         elif buffer.startswith('\x89PNG\r\n\x1a\n'):
-            extension = '.png'
+            mime = 'image/png'
         elif buffer.startswith('\xff\xd8'):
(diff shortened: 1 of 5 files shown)


heynemann:
This seems like a sensible change. As soon as someone (maybe @guilhermef may be able to help) takes a look and sees if this does not break stuff, we should be fine to merge it.

guilhermef:
@dhardy92 you're awesome, this PR is really good.

Issue #423, opened 2015-03-02
Allow re-evaluating the content-type after optimizations


masom:
The content-type of a response is evaluated before optimizers are executed.

There is some code to handle gifv/mp4 but if a different optimizer ( let say, gif -> flash ) is executed, Thumbor will always return the content-type of the pre-optimized image.

See https://github.com/thumbor/thumbor/blob/09ec9c9d0c4f10486d5c85a5cab5ae4ca2ebcd40/thumbor/handlers/__init__.py#L251)

It would be awesome if the optimizers could let Thumbor know if the content-type property should be evaluated again.

heynemann:
That's a nice change. Do you think you can submit a PR?

masom:
@heynemann There's a very simple PR that fixed the issue we are facing. It is based on the experimental gifv optimizer ( it sets the format at the end of the transformation ) and the change allow other kind of transformations ( mp4 -> gif, gif -> swf, etc. )

I couldn't get opencv to work properly on my dev machine, hopefully CI passes.

PR #425, opened 2015-03-03
Allow optimizers to change the response Content-Type

masom:
This PR allow optimizers to change the response format if the content type has been defined in `thumbor/handlers/__init__.py`

Fixes https://github.com/thumbor/thumbor/issues/423


guilhermef:
Thanks @masom

Issue #357, opened 2014-10-07
Convert animated GIF to non-animated (with an image-format filter)


illarionov:
Now the image-format filter can not be applied to animated GIFs. The output image remains in a gif format.
[gif](http://thumbor0.tasty0.ru/unsafe/0x0/filters:format(jpg)/att/be/5b/16143410_39649_19345266_9e00cdc9d718dd1648ce70dc)
It would be great if the image will be converted to the non-animated (using the first frame).
Or maybe add a filter to cut a specified frame.

guilhermef:
After experimental gifv(), I think it's time to deal with this issue.
@kplaube

guilhermef:
Done on https://github.com/thumbor/thumbor/wiki/Extract-Cover

PR #426, opened 2015-03-06
Extracting the first frame of a GIF

kplaube:
According to the issue number #357, we needed to deliver a GIF image as a static image. I did this using the first frame of the image, using Gifsicle. I didn't do it for PIL Engine, unfortunately it's not trivial as using Gifsicle.

I'm not pretty sure about the tests coverage. Is the content of this pull request ok? If isn't, someone could help me with this?

My very thanks to @guilhermef!


guilhermef:
Thank you @kplaube

PR #430, opened 2015-03-09
Add max clients setting

kevin1024:
If you're making lots of big upstream requests, you can run out of async HTTP clients, causing lots of upstream HTTP request timeouts.  This makes a setting so it can be tweaked.


masom:
Awesome!

kevin1024:
Oh, I should mention that 10 is tornado default for max_clients, so this should be a backwards-compatible change.

guilhermef:
Thanks @kevin1024

PR #433, opened 2015-03-09
Add config entry to choose allowed image dimensions.

debonzi:
When ALLOW_DIMENSIONS is set as list of image dimensions (e.g ['100x100', '200x200']=] it will return 404 for all dimensions but the ones defined.

### DIFF
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -75,7 +75,9 @@ Config.define(
 
 Config.define(
     'ALLOW_DIMENSIONS', [],
-    'Indicates whether thumbor should process certain images dimensions.', 'Imaging')
+    'Indicates whether thumbor should only process certain image dimensions. ' +
+    'It can be used to avoid url tampering that would overload the service. ' +
+    'If ALLOW_DIMENSIONS are set, filter will be forbiden.', 'Imaging')
 
 Config.define(
     'LOADER', 'thumbor.loaders.http_loader',
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -74,6 +74,10 @@ Config.define(
     'Indicates whether thumbor should enable blacklist functionality to prevent processing certain images.', 'Imaging')
 
 Config.define(
+    'ALLOW_DIMENSIONS', [],
+    'Indicates whether thumbor should process certain images dimensions.', 'Imaging')
+
+Config.define(
     'LOADER', 'thumbor.loaders.http_loader',
     'The loader thumbor should use to load the original image. This must be the full name of a python module ' +
     '(python must be able to import it)', 'Extensibility')
--- a/tests/fixtures/allow_dimensions_conf.py
+++ b/tests/fixtures/allow_dimensions_conf.py
@@ -1,0 +1,41 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/globocom/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com timehome@corp.globo.com
+
+# the domains that can have theyre images resized
+ALLOWED_SOURCES = ['s.glbimg.com', 'www.globo.com']
+
+# the max width of the resized image
+MAX_WIDTH = 1280
+
+# the max height of the resized image
+MAX_HEIGHT = 800
+
+ENGINE = 'thumbor.engines.pil'
+
+LOADER = 'thumbor.loaders.http_loader'
+
+STORAGE = 'thumbor.storages.file_storage'
+
+FILE_STORAGE_ROOT_PATH = "/tmp/thumbor/storage"
+
+# this is the security key used to encrypt/decrypt urls.
+# make sure this is unique and not well-known
+# This can be any string of up to 24 characters
+SECURITY_KEY = "MY_SECURE_KEY"
+
+# if you enable this, the unencryted URL will be available
+# to users.
+# IT IS VERY ADVISED TO SET THIS TO False TO STOP OVERLOADING
+# OF THE SERVER FROM MALICIOUS USERS
+ALLOW_UNSAFE_URL = True
+
+## Indicates whether thumbor should process certain images dimensions.
+## Defaults to: []
+ALLOW_DIMENSIONS = ['100x100', '300x600']
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -87,5 +87,22 @@ class MainHandlerFitInImagesTest(ImageTestCase):
         assert width == 200, width
         assert height == 151, height
 
+
+class MainHandlerAllowDimensions(AsyncHTTPTestCase):
+
+    def get_app(self):
+        return ThumborServiceApp(get_conf_path('allow_dimensions_conf.py'))
+
+    def test_allowed_dimension(self):
+        self.http_client.fetch(self.get_url('/unsafe/100x100/www.globo.com/media/globocom/img/sprite1.png'), self.stop)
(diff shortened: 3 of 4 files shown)


guilhermef:
@debonzi Why do you need to limit the images size ?
If you generate a signed url, nobody who doesn't have the key will be able to generate another size.

debonzi:
@guilhermef 
I intend to use it on mobile clients and, as far as I can see, I would need to ship the app with the KEY embedded which could make some trouble if someone manage to retrieve it and/or I need for any reason to change it on server side. Also, I will have a full JS web interface that would expose the KEY unless I create a proxy server to hold it and implement the dimensions filter at it.
I considered to use the blacklist that is already implemented but the functionality I need is more like a whitelist.

guilhermef:
@dhardy92 @heynemann What do you guys think ?
It's looking like a valid scenario.

@debonzi do you mind adding some test cases for this functionality ?

debonzi:
@guilhermef 
No problem. Working on it.

guilhermef:
@scorphus we need your opinion too.

heynemann:
I am for accepting this PR. Looks like a valid use. Couple things I'd add before accepting this PR:

* Change the default thumbor.conf to include this configuration commented out with the default value and the comment explaining how to use it;
* We need to define who's going to update the wiki with this new info. Do you think you can do it @debonzi? There are a couple pages to update.

scorphus:
I imagine @debonzi's intent with this PR is to prevent URL tampering. But how about other available options such as filters? One could still overload the service with `fill()`, for instance:

    http://example.com/unsafe/200x200/filters:fill(ffffff,true)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(ffffff,false)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(ffffff,1)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(ffffff,0)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(fffffe,true)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(fffffe,false)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(fffffe,1)/path/to/picture.png
    http://example.com/unsafe/200x200/filters:fill(fffffe,0)/path/to/picture.png

thumbor/config.py
```diff
@@ -74,6 +74,10 @@
     'Indicates whether thumbor should enable blacklist functionality to prevent processing certain images.', 'Imaging')
 
 Config.define(
+    'ALLOW_DIMENSIONS', [],
+    'Indicates whether thumbor should process certain images dimensions.', 'Imaging')
```
crccheck: To me, it seems like this should read "certain image dimensions"

dhardy92:
I agree with @scorphus this is not a real protection (just like spreading secret key, better using /unsafe/ actually)
Would be better filter at load balancing level (apache, haproxy,nginx, F5...) with rules allowing Location containg only patterns matching what is really wanted and 403 for other.

guilhermef:
I agree with them, that wouldn't avoid url tampering.
@debonzi how about @dhardy92 solution ?

debonzi:
Hi all,
@guilhermef  I didn't realize that it would still be possible to tamper using fill() because I am not using filtering.
First plan was to use @dhardy92 solution, but it seamed to me that it would be a nice option to be available on Thumbor. Another motivation is that I will use AWS Deis (Thumbor is running in a Docker container) and nginx, LB, etc are already on Deis environment working out of the box. I am not sure I can do those custom configurations easily on those services but been able to configure it on Thumbor, a simple 'git push' on deis infrastructure would do the magic.

I don't see an easy way to solve the fill() (and possible other) way to tamper with Thumbor url so I guess I will need to find a way to use on of the proposed solutions.

Thank you all anyway.

RTodorov:
+1 on this. Whitelist is a very useful feature, although I can't see a clear solution to the problem with the filters. To me, if someone wants to use the whitelist for security, he should also block or limit the use of filters.

debonzi:
I did another commit where it will return 403 when using filters with ALLOW_DIMENSIONS set. Also, described it on configuration.
If filters are the only case where it would be a flaw, I think it would be covered now.
Do you guys think it is acceptable?

guilhermef:
@debonzi I think It would be better to have a separated config like, DISABLE_FILTERS.

It's kinda strange to have ALLOW_DIMENSIONS disabling filters.

debonzi:
@guilhermef that is a way to go too, but my intention with that was to avoid someone to set ALLOW_DIMENSIONS and believe it safe when it is not.
If you guys think that doing a DISABLE_FILTERS and just suggest its use with ALLOW_DIMENSIONS on config comment and future documentation for this is ok, that would work for me too.

masom:
@debonzi Why not just disable filters with an empty list in FILTERS? Same goes with Optimizers or any future features that could potentially be "unsafe".

`ALLOWED_DIMENSIONS` is quite handy if you want to limit what image sizes web developers are allowed to use vs allowing all sizes. I don't think it should prevent filters from being used.

debonzi:
@masom That is fine for me too. All I need is to be able to limit the image sizes allowed. I just did this change based on @scorphus comment.
I am happy making any changes you guys think is best to get it merged.

felipemorais:
+1 ALLOWED_DIMENSIONS
+1 ALLOWED_FILTERS

masom:
@felipemorais `ALLOWED_FILTERS` is already implement as `FILTERS`

jeremyk:
Another suggestion to solve this problem (that would be a larger change) is to have WHITELISTED_OPERATIONS where you could have a name like 'default_thumbnail' that defines the size/filters part of the url. The url instead of /unsafe/200x200/filters/path_to_image would just be /named/default_thumbnail/path_to_image. The thumbor would just replace /named/default_thumbnail/ with the transformation. Has some caching implications when changing operation on the server but seems secure.

guilhermef:
The problem is that even with the FILTERS conf empty,
some attacker would still be able to request something like this:

http://example.com/unsafe/200x200/filters:fill(ffffff,true)/path/to/picture.png

The url regex still treats this url as valid.

heynemann:
@guilhermef is right. It's not a problem of actually executing the filters, since thumbor is not the issue here. It will just ignore filters that do not exist. The actual problem is caching. Since the URLs generated can be infinite, an attacker will be able to:

a) Fill the cache with infinite objects (all the same image, though, since filters will be ignored);
b) Overload thumbor since the caching mechanism will be 100% misses.

Both are equally bad for any given site, or app for that matter.

This is a hard problem, guys. Do not be disheartened because we haven't found a solution for it. Most imaging solutions haven't.

One possible solution would be to have a third route, something like `/preset/small-thumb/source-image.jpg`. This is a route that does not require the private key. This is the way imgix and cloudinary support both Front-end and Apps images. The presets need to be, well, pre set, as in you have to have them in thumbor's configuration. 

The key thing to understand here is that these presets include all available options to thumbor. You won't be able to use a preset with an URL specified filter. You will be able to use presets in more aggressive environments though (browser, apps, etc).

What do you guys think?

jeremyk:
@heynemann - I think this is what I was suggesting above. Right?

heynemann:
Yep. Sorry for repeating it. We just need to ensure it uses a different route than the signed urls. I think we should close this PR and open a new one with this change. What do you think?

jeremyk:
No worries, just wanted to make sure. Anyway, that sounds good to me!

scorphus:
It's been almost a year since we last discussed about this pull request. Some nice ideas came out of the initial suggestion. Should anyone like to see them in Thumbor, just open a new PR and we'll kindly review it. I'm closing this one for now.

PR #435, opened 2015-03-11
Fix StatsD client so it doesn't do a lot of DNS lookups

kevin1024:
The DNS lookups in the StatsD client use a blocking getaddrinfo call.

This change keeps the instance of the StatsD client around so it won't
have to look up DNS over and over again, so it won't block the IOLoop.

Also cleaned up some whitespace in the ThumborStatsClient class.

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -17,24 +17,35 @@ import statsd
 
 class ThumborStatsClient(statsd.StatsClient):
 
-  def __init__(self, config):
-    self.config = config
-    if config and config.STATSD_HOST:
-      self.enabled = True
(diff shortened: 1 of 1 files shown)


kevin1024:
If you guys merge this and my other pull request, I'll do another pull request to start sending some more stats, including:
 
  * Number of HTTP upstream workers queued (backlog)
  * Busy thread pool workers
  * Number of thread pool workers queued (backlog)

I think a total request time metric would be useful too.

masom:
:+1:

guilhermef:
Thank you @kevin1024 you're awesome

kevin1024:
:cake:

PR #436, opened 2015-03-11
Change optimizers to use subprocess.call instead of os.system

kevin1024:
os.system is deprecated and blocks the GIL.  Let's change the
optimizers to use subprocess.call instead.  This will fix the
optimizers blocking the IOLoop when using the thread pool, and also
helps eliminate a deprecated library from the codebase.

### DIFF
--- a/thumbor/optimizers/gifv.py
+++ b/thumbor/optimizers/gifv.py
@@ -8,7 +8,7 @@
 # http://www.opensource.org/licenses/mit-license
 # Copyright (c) 2011 globo.com timehome@corp.globo.com
 
-import os
+import subprocess
 
 from thumbor.optimizers import BaseOptimizer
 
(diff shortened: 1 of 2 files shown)


kevin1024:
@masom if you found a way way to call the optimizers with the subprocess module that keeps your terminal from getting messed up, feel free to open a pull request of your own and I'll close this one.  I wasn't able to reproduce the terminal oddness on my machine.

thumbor/optimizers/gifv.py
```diff
@@ -28,7 +28,7 @@ def optimize(self, buffer, input_file, output_file):
             command_params,
             output_file,
         )
-        os.system(command)
+        subprocess.call(command, shell=True)
```
masom: `subprocess.call(command, shell=True, stdin=open(os.devnull))`

ffmpeg somehow tries to read from stdin and messes up the shell.

kevin1024:
OK cool, I made the change as suggested, but used the with statement to make sure the file handle to /dev/null gets closed when the command finishes.

PR #439, opened 2015-03-16
Add Last-Modified header handling.

acaciocenteno:
This is not a Thumbor issue per se, but many people use it behind Nginx, and Nginx won't properly handle the ETag header if the response lacks a Last-Modified header. The result is that if a browser tries to revalidate an image, Nginx will send a 200 response instead of a 304.

This minor patch passes the Last-Modified header received from the origin if one is available or adds one with the current time otherwise.

Off course, the problem could be solved by forcing the header on Nginx itself, but doing it on Thumbor seemed more elegant, as we can respect what the origin has set.

Also, as I'm not acquainted to the code, the way I did it might not be the best one. Should you think this patch is worth but would like to take another approach, just let me know and I'll fix it.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -373,13 +373,20 @@ class BaseHandler(tornado.web.RequestHandler):
         else:
             self.context.statsd_client.incr('storage.miss')
 
-            def handle_loader_loaded(buffer):
+            def handle_loader_loaded(response):
+                if response is None:
+                    callback(False, None)
+                    return
+
+                buffer = response.body
+
                 if buffer is None:
                     callback(False, None)
                     return
 
                 original_preserve = self.context.config.PRESERVE_EXIF_INFO
                 self.context.config.PRESERVE_EXIF_INFO = True
+                self.set_header('Last-Modified', response.headers['Last-Modified'])
 
                 try:
                     mime = BaseEngine.get_mimetype(buffer)
--- a/thumbor/loaders/http_loader.py
+++ b/thumbor/loaders/http_loader.py
@@ -47,11 +47,14 @@ def return_contents(response, url, callback, context):
         logger.warn("ERROR retrieving image {0}: Empty response.".format(url))
         callback(None)
     else:
+        if not 'Last-Modified' in response.headers:
+            response.headers['Last-Modified'] = datetime.datetime.utcnow()
+
         if response.time_info:
           for x in response.time_info:
               context.statsd_client.timing('original_image.time_info.' + x, response.time_info[x] * 1000)
           context.statsd_client.timing('original_image.time_info.bytes_per_second', len(response.body) / response.time_info['total'])
-        callback(response.body)
+        callback(response)
 
 
 def load(context, url, callback):


masom:
Thumbor somewhat supports the `Last-Modified` header depending on the result storage:

````
# Sends If-Modified-Since & Last-Modified headers; requires support from result storage
SEND_IF_MODIFIED_LAST_MODIFIED_HEADERS = True
```

Does that work for you?

acaciocenteno:
Yes, we did try that and it does work when using the result storage, but
that is not our case, as we use Nginx for caching. Nginx is really the
issue here, as it will only validate etag if a Last-Modified header is
present.

As mentioned the patch we did might not be the best one, if someone has a
better ideia, we can change it.

2015-03-17 12:47 GMT-03:00 Martin Samson <notifications@github.com>:

> Thumbor somewhat supports the Last-Modified header depending on the
> result storage:
>
> # Sends If-Modified-Since & Last-Modified headers; requires support from result storage
> SEND_IF_MODIFIED_LAST_MODIFIED_HEADERS = True
>
> Does that work for you?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/thumbor/thumbor/pull/439#issuecomment-82425700>.
>



-- 
Atenciosamente,
Luiz Acácio Centeno Cordeiro Jr
(51) 8118-9947

thumbor/handlers/__init__.py
```diff
@@ -373,13 +373,20 @@ def _fetch(self, url, extension, callback):
         else:
             self.context.statsd_client.incr('storage.miss')
 
-            def handle_loader_loaded(buffer):
+            def handle_loader_loaded(response):
+                if response is None:
+                    callback(False, None)
+                    return
+
+                buffer = response.body
```
phoet: @acaciocenteno i think this is not working with the other handlers and because of that lots of the tests fail on travis with `AttributeError: 'str' object has no attribute 'body'`

```
[0mERROR:tornado.access:500 GET /unsafe/filters:max_age(30)/image.jpg (127.0.0.1) 4.73ms
[0mERROR:thumbor:ERROR: Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/tornado/web.py", line 1320, in _stack_context_handle_exception
    raise_exc_info((type, value, traceback))
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/tornado/web.py", line 1503, in wrapper
    result = method(self, *args, **kwargs)
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/imaging.py", line 100, in get
    self.check_image(kw)
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/imaging.py", line 96, in check_image
    return self.execute_image_operations()
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/__init__.py", line 97, in execute_image_operations
    self.filters_runner.apply_filters(thumbor.filters.PHASE_PRE_LOAD, self.get_image)
  File "/home/travis/build/thumbor/thumbor/thumbor/filters/__init__.py", line 81, in apply_filters
    callback()
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/__init__.py", line 122, in get_image
    self._fetch(self.context.request.image_url, self.context.request.extension, callback)
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/__init__.py", line 415, in _fetch
    self.context.modules.loader.load(self.context, url, handle_loader_loaded)
  File "/home/travis/build/thumbor/thumbor/thumbor/loaders/file_loader.py", line 22, in load
    callback(f.read())
  File "/home/travis/build/thumbor/thumbor/thumbor/handlers/__init__.py", line 381, in handle_loader_loaded
    buffer = response.body
AttributeError: 'str' object has no attribute 'body'
```

phoet:
i would also like to see something like this as we can't use the result storage for the last-modified headers.

my workaround was a bit different though: https://github.com/thumbor/thumbor/commit/c6fd880678bd1cfac48310a4af481214ef976458

PR #448, opened 2015-03-24
Prevent the gif engine from returning an invalid image.

masom:
In certain scenario it is possible that gifsicle return an empty or invalid STDOUT output.

This patch uses PIL to ensure the `buffer` value is a valid image file.


guilhermef:
@masom can you create a test to ensure that this problem doesn't come back ?

masom:
@guilhermef Working on that right now

masom:
@guilhermef added tests.

guilhermef:
Thanks @masom ;)

PR #450, opened 2015-03-25
Wiki documentation ported to read the docs (sphinx)

ricardodani:


ricardodani:
@masom This is the part of docs that I worked on, it contains all the docs written in Thumbor's wiki, plus the read the docs theme, a Makefile target (make docs) to setup and generate them, and a organized menu of topics.

I think you can patch your PR's in the codebase of this docs, making sure that will be in the correct menu item. I Think we can call this menu item of auto generated docs as "API Reference" or "Module Reference". What do you think?

About your PR's, I checkout them on my machine and experienced some build issues like no modules available. Maybe you need to test this docs in an empty virtualenv to ensure that all dependecies of the docs are in docs/requirements.txt file. (I write a Makefile target called setup_docs to install them).

masom:
Awesome. I'll rebase once this is merged.

As for the module errors it seems like the installation of opencv isn't recognized by your virtual env.

masom:
"API Reference" might be slightly less confusing to end-users vs developers.

ricardodani:
@masom I think API Reference is better than Module Reference too...
What do you think @heynemann @guilhermef ?

masom:
@ricardodani @guilhermef @heynemann it would be awesome to get this merged as it would allow the other PRs to be rebased and some documentation work started.

guilhermef:
Done @masom

PR #453, opened 2015-03-27
Fix error log on load then store

dhardy92:
My thumbor Ids have no extension so this generate lots of ERROR logs.
This patch handle extension is None to prevent unnecessary flood in logs.
Excotic extentions still generate a log...

Tests are not really relevant in this case but increase coverage :) I don't know how to test if log is emitted (any idea ?)

Side note: looking a the code it seams to me we are always trying to determine mime and extension from modules to modules ... may this be centralized ?


guilhermef:
Thanks @dhardy92

PR #457, opened 2015-03-30
Callback-based storage.

### LINKED ISSUE #452: BaseStorage interface is missing callback parameter to optimize for Tornado Runloop
I noticed a huge bottleneck when using the thumbor storages and result storages:

The BaseStorage interface in
https://github.com/thumbor/thumbor/blob/master/thumbor/result_storages/__init__.py#L22
```python
class BaseStorage(object):
    ...
    def put(self, bytes):
        ...

    def get(self):
        ...
```
supplies only synchronous methods with return values instead of callbacks like the loader interface in
https://github.com/thumbor/thumbor/blob/master/thumbor/loaders/file_loader.py
```python
```
(perfectly fine)

https://github.com/thumbor/thumbor/blob/master/thumbor/result_storages/__init__.py
Should look more like 

```python
#!/usr/bin/python
# -*- coding: utf-8 -*-
...
import os
from os.path import exists

class BaseStorage(object):
    def __init__(self, context):
        self.context = context

    def put(self, bytes, callback):
        raise NotImplementedError()

    def get(self, callback):
        raise NotImplementedError()

    def last_updated(self, callback):
        raise NotImplementedError()

    def ensure_dir(self, path, callback):
        raise NotImplementedError()
```

masom:
Should fix #452

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -265,9 +265,15 @@ class ThreadPool(object):
             self.pool = None
 
     def _execute_in_foreground(self, operation, callback):
-        result = Future()
-        result.set_result(operation())
-        callback(result)
+
+        def process_callback():
+            result = Future()
+            result.set_result(operation())
+            callback(result)
+
+        tornado.ioloop.IOLoop.instance().add_callback(
+            functools.partial(process_callback)
+        )
 
     def _execute_in_pool(self, operation, callback):
         task = self.pool.submit(operation)
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -265,15 +265,9 @@ class ThreadPool(object):
             self.pool = None
 
     def _execute_in_foreground(self, operation, callback):
-
-        def process_callback():
-            result = Future()
-            result.set_result(operation())
-            callback(result)
-
-        tornado.ioloop.IOLoop.instance().add_callback(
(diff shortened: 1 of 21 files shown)


masom:
@guilhermef We use file and redis so we can't really see if this improve performance ( both storage don't benefit from this ).

masom:
The branch will be rebased if it looks ok.

guilhermef:
@masom take a look at the gen.coroutine implementation http://tornado.readthedocs.org/en/latest/gen.html

Maybe we can avoid a little the callback hell.

masom:
Oh how silly...

I'll start a new branch that use `tornado.gen.maybe_future`

guilhermef:
Just be careful because not all features are available since we're on
tornado 3.2

masom:
@guilhermef I wonder if the storages should just return a `Future` instance? That way we can use the coroutine.

heynemann:
Sounds reasonable @masom.

masom:
Moving toward https://github.com/thumbor/thumbor/pull/459 instead of callback hell. I wish I would have figured out to use Tornado instead of a bunch of little callbacks.

heynemann:
You are doing great @masom Can't even stress how valuable your contributions are to the whole community. Thanks!

PR #458, opened 2015-03-31
engines.pil to invoke get_default_extension() on self.

watsonrelay:
This happens when image file extension was not determined, then PIL engine needs to invoke self.get_default_extension() on self.


guilhermef:
Thanks @watsonrelay, @heynemann already fixed this on #451

watsonrelay:
Thanks!!

PR #459, opened 2015-03-31
Storage use tornado.concurrent.Future

### LINKED ISSUE #452: BaseStorage interface is missing callback parameter to optimize for Tornado Runloop

masom:
Should fix #452

Two vows are currently failing due to some Async operation timing out after 5 seconds. Not sure why.


masom:
@guilhermef @heynemann is it possible the failing Vows are hitting http://stackoverflow.com/a/14779326/1014879 ?

masom:
image

Not sure why this times out. The methods used should not have been impacted by the added futures.

guilhermef:
@masom, you're a genius!
I've made this change on tornado_pyvows, 
looks like we can finally update to tornado 4.1.

masom:
@guilhermef Seems like this PR has the same issue that was reported in https://github.com/rafaelcaricio/tornado_pyvows/issues/12

masom:
I just tried the latest tornado_pyvows and it crashes with the following error when testing thumbor:

```
  File "/Users/martin/venvs/thumbor/bin/pyvows", line 9, in <module>
    load_entry_point('pyVows==2.0.6', 'console_scripts', 'pyvows')()
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/pyvows/cli.py", line 180, in main
    result = run(path, pattern, verbosity, arguments.progress, prune)
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/pyvows/cli.py", line 137, in run
    Vows.collect(path, pattern)
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/pyvows/core.py", line 155, in collect
    __import__(module_name)
  File "/Users/martin/dev/os/thumbor/vows/transformer_vows.py", line 13, in <module>
    from vows.transformer_test_data import TESTITEMS, FIT_IN_CROP_DATA, TestData, MockSyncDetector, MockErrorSyncDetector
ImportError: No module named transformer_test_data
```

guilhermef:
@masom, I think that this error is related to the code, I'll create a tornado4 branch so we can use to compare.

masom:
@guilhermef works now.

There's still 2 vows that fail with a timeout of 5 seconds ( even if pyvows marks the tests as completed)

guilhermef:
@masom, I've changed _fetch to use gen.coroutine https://github.com/thumbor/thumbor/commit/9179435fcd133beccff63a4815985f024fd25551

thumbor/filters/frame.py
```diff
@@ -88,7 +90,7 @@ def frame(self, callback, url):
         self.nine_patch_engine = self.context.modules.engine.__class__(self.context)
         self.storage = self.context.modules.storage
 
-        buffer = self.storage.get(self.url)
+        buffer = yield self.storage.get(self.url)
```
guilhermef: @masom I think this one could be
```
buffer = yield gen.maybe_future(self.storage.get(self.url))
```
this way if the storage doesn't implement the gen. coroutine, it won't raise an error.

masom: It would be great to use `maybe_future` but it is not part of tornado 3.2 :(

guilhermef: I've just updated to tornado 4.1 ;)

masom: sweet!

masom:
@guilhermef Working with maybe_future. Now all that's borking are the two requests to the upload handler.

guilhermef:
@masom it's looking good

guilhermef:
is it done ?

masom:
https://github.com/thumbor/thumbor/pull/459#issuecomment-88245594 is still happening.

guilhermef:
did you comment these test ?
because travis looks ok.

masom:
Travis say they are ok but the two vows are still timing out in the build logs.

Also I just got this error when moving to maybe_future:
```
2015-04-01 16:32:23 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/gen.py", line 812, in run
    yielded = self.gen.send(value)
  File "/Users/martin/dev/os/thumbor/thumbor/handlers/__init__.py", line 106, in get_image
    self.context.request.extension
TypeError: 'NoneType' object is not iterable
2015-04-01 16:32:23 thumbor:ERROR [REDIS_STORAGE] Error 61 connecting to localhost:6379. Connection refused.
2015-04-01 16:32:23 thumbor:ERROR ERROR: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/web.py", line 1302, in _stack_context_handle_exception
    raise_exc_info((type, value, traceback))
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/stack_context.py", line 314, in wrapped
    ret = fn(*args, **kwargs)
  File "/Users/martin/dev/os/thumbor/thumbor/loaders/http_loader.py", line 54, in return_contents
    callback(response.body)
  File "/Users/martin/dev/os/thumbor/thumbor/handlers/__init__.py", line 415, in handle_loader_loaded
    raise gen.Return([normalized, None, self.context.request.engine])
```

guilhermef:
the method check_resource on https://github.com/masom/thumbor/blob/x-452-futures/thumbor/handlers/image_resource.py
is missing a self.finish() call.

scorphus:
Amazing! It's looking really good, @masom! What do you think about squashing all these commits into a single one once it's ready?

masom:
@scorphus yep. right now there's some weird stuff happening but it's looking good.

masom:
This is happening right now:

```
WARNING:tornado.access:404 GET /image/an/unknown/image (127.0.0.1) 1.46ms
ERROR:tornado.application:Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/gen.py", line 812, in run
    yielded = self.gen.send(value)
  File "/Users/martin/dev/os/thumbor/thumbor/transformer.py", line 166, in do_smart_detection
    detectors[0](self.context, index=0, detectors=detectors).detect(self.after_smart_detect)
  File "/Users/martin/dev/os/thumbor/vows/transformer_test_data.py", line 73, in detect
    raise Exception('x')
Exception: x
ERROR:tornado.application:Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/gen.py", line 812, in run
    yielded = self.gen.send(value)
  File "/Users/martin/dev/os/thumbor/thumbor/transformer.py", line 166, in do_smart_detection
    detectors[0](self.context, index=0, detectors=detectors).detect(self.after_smart_detect)
  File "/Users/martin/dev/os/thumbor/vows/transformer_test_data.py", line 73, in detect
    raise Exception('x')
Exception: x
```

guilhermef:
I think that this error is expected

masom:
Ok so right now the only blocking is a 500 when the result_storage does not have the file:
```
2015-04-01 16:53:27 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/gen.py", line 812, in run
    yielded = self.gen.send(value)
  File "/Users/martin/dev/os/thumbor/thumbor/handlers/__init__.py", line 106, in get_image
    self.context.request.extension
TypeError: 'NoneType' object is not iterable
2015-04-01 16:53:27 thumbor:ERROR ERROR: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/web.py", line 1302, in _stack_context_handle_exception
    raise_exc_info((type, value, traceback))
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/stack_context.py", line 314, in wrapped
    ret = fn(*args, **kwargs)
  File "/Users/martin/dev/os/thumbor/thumbor/loaders/http_loader.py", line 54, in return_contents
    callback(response.body)
  File "/Users/martin/dev/os/thumbor/thumbor/handlers/__init__.py", line 415, in handle_loader_loaded
    raise gen.Return([normalized, None, self.context.request.engine])
Return
```

Seems to be related to changes in `_fetch`

guilhermef:
which test is broken ?

masom:
No test but when running thumbor with the http_loader the first time the file is fetched this error is raised.

guilhermef:
oh, that's bad, we should have a test for this.

masom:
@guilhermef Agreed. At least I can reproduce but I can't figure out why this is happening.


First request when the storage is empty fails and the second one succeeds:
```
2015-04-01 17:07:27 thumbor:DEBUG STATSD: result_storage.incoming_time:0|ms
2015-04-01 17:07:27 thumbor:DEBUG STATSD: result_storage.miss:1|c
2015-04-01 17:07:27 thumbor:DEBUG STATSD: storage.miss:1|c
2015-04-01 17:07:27 thumbor:DEBUG STATSD: original_image.status.200:1|c
2015-04-01 17:07:27 thumbor:ERROR ERROR: Traceback (most recent call last):
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/web.py", line 1302, in _stack_context_handle_exception
    raise_exc_info((type, value, traceback))
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/stack_context.py", line 314, in wrapped
    ret = fn(*args, **kwargs)
  File "/Users/martin/dev/os/thumbor/thumbor/loaders/http_loader.py", line 54, in return_contents
    callback(response.body)
  File "/Users/martin/dev/os/thumbor/thumbor/handlers/__init__.py", line 417, in handle_loader_loaded
    raise gen.Return([normalized, buffer, self.context.request.engine])
Return

2015-04-01 17:07:27 tornado.access:ERROR 500 GET /unsafe/80x80/filters:quality(80)/gif-1188749c-ed13-45b1-997d-8dc8a22b2df6.gif (127.0.0.1) 488.23ms
2015-04-01 17:07:36 thumbor:DEBUG STATSD: result_storage.incoming_time:0|ms
2015-04-01 17:07:36 thumbor:DEBUG STATSD: result_storage.miss:1|c
2015-04-01 17:07:36 thumbor:DEBUG STATSD: storage.hit:1|c
2015-04-01 17:07:36 thumbor:DEBUG No image format specified. Retrieving from the image extension: .gif.
2015-04-01 17:07:36 thumbor:DEBUG Content Type of image/gif detected.
2015-04-01 17:07:36 tornado.access:INFO 200 GET /unsafe/80x80/filters:quality(80)/gif-1188749c-ed13-45b1-997d-8dc8a22b2df6.gif (127.0.0.1) 151.00ms
2015-04-01 17:07:36 thumbor:DEBUG [REDIS_RESULT_STORAGE] putting `result:/unsafe/gif-1188749c-ed13-45b1-997d-8dc8a22b2df6.gif/test_f-quality-80_s-80x80.gif` with ttl `600`
2015-04-01 17:07:36 thumbor:DEBUG STATSD: result_storage.bytes_written:90313|c
2015-04-01 17:07:36 thumbor:DEBUG STATSD: result_storage.outgoing_time:0|ms
```

guilhermef:
That's probably happening because It's returning after this: https://github.com/masom/thumbor/blob/x-452-futures/thumbor/handlers/__init__.py#L417

masom:
ah. I'll see what I can do. Probably do the maybe_future and have the loader use `return_future` that way the API stays relatively the same and we can yield.

guilhermef:
I think I've introduced this bug, with the gen.coroutine on _fetch.
The loader should be a gen.coroutine too.

masom:
@guilhermef Just updated, removing a callback.

guilhermef:
@masom I'm impressed, you're really fast.

masom:
@guilhermef :) I actually started to work on the yield when you mentioned the code reached the end of the function.

masom:
Seems like we also broke the gif engine.
```
            File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pyvows/runner/abc.py", line 63, in run_vow
    result = vow(ctx_obj, topic)
            File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/pyvows/decorators.py", line 76, in wrapper
    ret = method(*args, **kw)
            File "/home/travis/build/thumbor/thumbor/vows/gif_engine_vows.py", line 45, in should_throw_an_exception
    expect(topic).to_be_an_error_like(IOError)
            File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/preggy/core.py", line 232, in _assert_topic
    return _registered_assertions[method_name](self.topic, *args, **kw)
            File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/preggy/core.py", line 59, in wrapper
    func(*args, **kw)
            File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/preggy/assertions/types/errors.py", line 25, in to_be_an_error_like
    raise err
AssertionError: ("Expected topic(unpack requires a string argument of length 4) to be an error of type <type 'exceptions.IOError'>, but it was a <class 'struct.error'>", error('unpack requires a string argument of length 4',), <type 'exceptions.IOError'>, <class 'struct.error'>)
```

guilhermef:
That's strange, because that only breaks on travis

masom:
Yep also working here.

guilhermef:
We change this test to 
```python
expected = "cannot identify image file <_io.BytesIO"
expect(topic.message).to_include(expected)
```
and hope for the best

masom:
@guilhermef has pyvows / tornado_pyvows changed how  `@Vows.capture_error`
 works?

guilhermef:
Probably not, we're still using the same version.
I've started a new env and it passed.

masom:
The error is thrown by Pillow.

masom:
@guilhermef Travis is pulling in Pillow 2.8 which returns a different error. I just upgrade to 2.8 in my env and it fails the same way.

guilhermef:
At least the error is consistent,
But that's really strange since I've started my env from scratch.

guilhermef:
@masom, can I merge ?

masom:
@guilhermef Looks good to me.

guilhermef:
Thanks for this amazing PR @masom

scorphus:
Superb, @masom. It was/is a lesson. Thank you.

RobertBiehl:
Thanks for working on this! Will this be a part of the next 5.0 release?

guilhermef:
@RobertBiehl currently there's a pre release, https://github.com/thumbor/thumbor/releases .
If everything is ok, we'll release as 5.0.

banks:
This looks great, but I'm left a bit confused, the changelog warning on https://github.com/thumbor/thumbor/releases for 5.0.0 claims 

> If you use your own version of a Storage, ResultStorage or Imaging Engine, all of those are async now and should return a Future.

But in this diff and the current master code I see no change to the Engine interface at all and no mention of futures.

This is important to us as we use opencv engine which also has no futures in latest version and no commits that make it clear it is compatible with new thumbor release. We have other custom engines that pick best engine based on criteria. And it's unclear if we need to update any of those - code says no, but the warning specifically says we should...

Can you confirm if that warning message is misleading and perhaps update it in change log if in fact image engines didn't require any code changes?

guilhermef:
@banks that's the right diff you should take a look: https://github.com/thumbor/thumbor/compare/4.12.2...5.0.0

These commits to be more specific: 
https://github.com/thumbor/thumbor/commit/9179435fcd133beccff63a4815985f024fd25551
https://github.com/thumbor/thumbor/commit/95308c87187915bd1c3fcd2777e2369aba1a38e4
https://github.com/thumbor/thumbor/commit/719878331dcbed6b63788d651f62b2d41da6fc26
https://github.com/thumbor/thumbor/commit/18570417805e497402279ee6a60f9ddf1d356a81

banks:
@guilhermef thanks although I guess I'm still missing the answer to my question.

All of those commits seem to indicate that the new async handling of engines is a wrapper/dispatcher layer ABOVE the actual Engine itself.

In other words the actual PIL engine or base engine interface did not change and hence opencv ending e in a different repo that was not touched should still work right? And these custom engines we made that extend BaseEngine also should work right?

If so then the comment on the changelog is confusing - I assumed it meant we had to re-write out custom engines and port OpenCV as I saw no changes to it but when trying to figure out how I found that actually no changes to engines are apparently required...

You'll notice me in another issue today pointing out a bug and wanting to upgrade to see if it's fixed and this seemed like a barrier for upgrading but now I'm not sure... When I get time I can test it I guess, but it would be useful to confirm that I'm right in suspecting the comment was misleading rather than missing some subtle breaking change for Engines?

masom:
@banks at PopKey we locked ourselves to 4.x until some of our custom extensions are ported to use futures.

Using `maybe_future` should kinda prevent breaking changes in engines and most loaders although you should test your setup with 5.x before making the switch in production.

banks:
@masom so the idea is that maybe_future is backwards compatible with non-future engines/storages but can elverage async where those do work? If so it seems like the warning is quite overblown in the changelog and "breaking change" version bump.

Also the bundled PIL engine appears to still return sync responses not futures so I guess it must be a well tested code path!

Thanks anyway, we'll test and see what happens.

masom:
@banks The API did change a little bit for a few items and changing from sync code to async with generators + a new version of tornado kinda makes this upgrade a lot more risky.

We had a CI machine automatically install thumbor ( no version check ) and some of our custom extensions broke.

guilhermef:
@banks We had the same problem at Globo, we've changed our loaders to fit the new API.

banks:
OK thanks all. In case anyone reading this later was confused the definitive answer as given here: https://github.com/thumbor/thumbor/issues/490#issuecomment-110376046 seems to be: custom engines probably don't need to change code since 4.12 introduced thread pool anyway and maybe_future allows non future results (but you really should test as with any production upgrade), any other exensions like loaders (or storages?), probably warrant closer inspection.

PR #465, opened 2015-04-12
jpegtran optimizer copy comments only

afjlambert:
Reduce file size by only copying markers that are essential for image display. This leaves out JFIF thumbnails, Exif data, and Photoshop settings, see http://linux.die.net/man/1/jpegtran


heynemann:
Thanks!!! :)

kkopachev:
Why keep comments?

PR #479, opened 2015-05-17
Update pymongo dependency

marcelometal:
Tested with 3.0.1


marcelometal:
See thumbor/thumbor/issues/477

marcelometal:
If you apply this patch, please close #479

PR #482, opened 2015-05-22
Https loader and strict https loader bugfix

lukaselmer:
Sorry, I've accidentally made a mistake programming the https loaders. Here's the fix, including tests.

Explanation: [Here](https://github.com/thumbor/thumbor/blob/master/thumbor/handlers/__init__.py#L385) the loader is called without a callback. I've also written a test for the http loader.


guilhermef:
thanks @lukaselmer

PR #531, opened 2015-08-13
fix #528 - not image request

### LINKED ISSUE #528: Thumbor deamon segFault on false url
Hello,

During some of our test on thumbor, we realised that when calling the thumbor server, with an uri that is not an image (and respond with code 200), the deamon crashes without error.

You can easily reproduce that by calling:
curl -I "http://localhost:8888/unsafe/300x200/smart/google.fr"

You should get an error like:
(52) Empty reply from server

and after that the daemon is gone.

Regards,
Antoine

lfalcao:
if can't identify the mime type, returns 404. 

(Still i dont know if it's the best way to fix it.)

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -372,6 +372,9 @@ class BaseHandler(tornado.web.RequestHandler):
         if mime is None:
             mime = BaseEngine.get_mimetype(buffer)
 
+        if mime is None:
+            raise gen.Return([False, None, None])
+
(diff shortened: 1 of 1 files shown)


masom:
Should probably be a `400 Bad Request`

heynemann:
Yep, we have a different request to change the response when thumbor can't find a source image to a 504 http code. I do think here we should also change thumbor to return 400. If we return 404 we'll be cache poisoning any intermediaries between thumbor and the client.

Issue #537, opened 2015-08-18
exceptions not logged to custom error handler


phoet:
we want to use sentry for custom error reporting, but it does not receive any errors, eventhough we see some in the logs:

```
2015-08-18T08:58:34.870570+00:00 app[web.6]: 2015-08-18 08:58:34 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
2015-08-18T08:58:34.870576+00:00 app[web.6]:   File "/app/.heroku/python/lib/python2.7/site-packages/tornado/gen.py", line 876, in run
2015-08-18T08:58:34.870578+00:00 app[web.6]:     yielded = self.gen.throw(*exc_info)
2015-08-18T08:58:34.870579+00:00 app[web.6]:   File "/app/.heroku/src/thumbor/thumbor/handlers/__init__.py", line 86, in get_image
2015-08-18T08:58:34.870580+00:00 app[web.6]:     self.context.request.image_url
2015-08-18T08:58:34.870582+00:00 app[web.6]:   File "/app/.heroku/python/lib/python2.7/site-packages/tornado/gen.py", line 870, in run
2015-08-18T08:58:34.870583+00:00 app[web.6]:     value = future.result()
2015-08-18T08:58:34.870585+00:00 app[web.6]:   File "/app/.heroku/python/lib/python2.7/site-packages/tornado/concurrent.py", line 215, in result
2015-08-18T08:58:34.870586+00:00 app[web.6]:     raise_exc_info(self._exc_info)
2015-08-18T08:58:34.870588+00:00 app[web.6]:   File "/app/.heroku/python/lib/python2.7/site-packages/tornado/gen.py", line 879, in run
2015-08-18T08:58:34.870589+00:00 app[web.6]:     yielded = self.gen.send(value)
2015-08-18T08:58:34.870590+00:00 app[web.6]:   File "/app/.heroku/src/thumbor/thumbor/handlers/__init__.py", line 389, in _fetch
2015-08-18T08:58:34.870592+00:00 app[web.6]:     self.context.request.engine.load(buffer, extension)
2015-08-18T08:58:34.870593+00:00 app[web.6]:   File "/app/.heroku/src/thumbor/thumbor/engines/__init__.py", line 96, in load
2015-08-18T08:58:34.870595+00:00 app[web.6]:     image_or_frames = self.create_image(buffer)
[...]

PR #541, opened 2015-08-20
Handle invalid images and return a 400 or 500 depending on the cause

masom:
Prevent thumbor from hanging when the engine or loader throw an exception.

Handle exceptions thrown by the `BaseHandler._fetch` method. If the exception is thrown by Pillow ( IOError  with a specific message ) the response code is set to 400 ( invalid request ).

Related to https://github.com/thumbor/thumbor/pull/531
Helps with https://github.com/thumbor/thumbor/issues/528, https://github.com/thumbor/thumbor/issues/537


heynemann:
Thank you VERY much @masom! You are doing an incredible job!

heynemann:
The downside here is that we need to check the other PR to make sure if we need to rebase before accepting.

vinczente:
awesome, thanks guys

heynemann:
We'll release a new version soon.

PR #542, opened 2015-08-20
Fixes #529 - cache poisioning issues with returning 404 when an image…

### LINKED ISSUE #529: Thumbor returns a 404 when it has a timeout retrieving the original image, should be a 504
I'm experiencing an issue similar to #208 in the same circumstances: Thumbor, under high load, is logging error messages about 599 timeouts when trying to fetch original source images so that it can create thumbnails of them.

Here's a log from one time that happened:

```
2015-08-12T00:52:20.488677+00:00 app[web.2]: 2015-08-12 00:52:20 thumbor:WARNING ERROR retrieving image https://example-s3-bucket.s3.amazonaws.com/uploads/listing-55/747.jpg: HTTP 599: Timeout
2015-08-12T00:52:20.786979+00:00 app[web.2]: 2015-08-12 00:52:20 tornado.access:WARNING 404 HEAD /THUMBOR_SECRET_REDACTED=/1800x1200/https://example-s3-bucket.s3.amazonaws.com/uploads/listing-55/747.jpg (10.169.215.33) 7136.35ms
```

When Thumbor experiences a timeout in this situation, it's returning a 404 status code. That's definitely not the right status code for this behavior, as it authoritatively means that file does not exist and tells upstream proxies to cache the result.

The correct status code in this scenario is a 5XX. In particular the 504 description perfectly summarizes this scenario:
 
> ## 10.5.5 504 Gateway Timeout
> The server, while acting as a gateway or proxy, did not receive a timely response from the upstream server specified by the URI (e.g. HTTP, FTP, LDAP) or some other auxiliary server (e.g. DNS) it needed to access in attempting to complete the request.

Is it possible to change Thumbor so it returns the correct HTTP status code when it experiences a timeout or other internal error?

heynemann:
… could not be loaded properly.

This also likely fixes the issue with hanging errors. The call was being made to context.log_error and should be self.log_error(*sys.exc_info()).

/cc @masom @bgentry #529

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -8,6 +8,7 @@
 # http://www.opensource.org/licenses/mit-license
 # Copyright (c) 2011 globo.com timehome@corp.globo.com
 
+import sys
 import functools
 import datetime
 import traceback
@@ -92,7 +93,7 @@ class BaseHandler(tornado.web.RequestHandler):
                 error=e
             )
 
-            self.context.log_exception(e)
+            self.log_exception(*sys.exc_info())
 
             if 'cannot identify image file' in e.message:
                 logger.warning(msg)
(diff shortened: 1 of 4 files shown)


masom:
:+1:

PR #554, opened 2015-09-11
FEAT: Use a LoaderResult object to support origin metadata and error handling

masom:
Following a discussion in https://github.com/thumbor/thumbor/issues/529 this PR implements a value object that holds loader response metadata.

Using a value object will allow thumbor to properly respond to upstream failures or errors.


heynemann:
Do you think this PR should also include the fix to return 404 when image not found and 504 otherwise?

heynemann:
Btw I wish I was half as productive as you are @masom... Thanks for the awesome job!

masom:
It should probably include the 404/504 fix although this means a little bit more refactoring. The `_fetch` method would need to return either a longer array or a value object to relay the information from the loader.

heynemann:
I agree we should return Value objects, but do you think it would be an acceptable compromise to return a larger array of values for this PR? Maybe we can get this refactor bit in another PR. If you feel you can do it all in the same PR, by all means go for it. I'm just worried that right now thumbor has a security leak (if people keep sending 404 images they can DDOS the server, since the edge caches won't cache it).

masom:
It depends on the edge node configuration. For instance AWS CloudFront will cache 404 / 5xx errors for 5 minutes unless configured otherwise.

masom:
The error handling behaviour might have to be changed to match https://github.com/thumbor/thumbor/pull/531 and other PRs/Issues.

@guilhermef  @heynemann @phoet What do you guys think of this approach?

The `metadata` property still has to be passed to the FetchResult object and conditional error handling needs to be improved.

phoet:
Cool THX. I did not understand why some vows were removed?

masom:
@phoet The deleted vows are a copy of the `http_loader` vows and they don't test anything from the `https_loader`

phoet:
@masom ah ok

masom:
@guilhermef @heynemann OK to merge?

guilhermef:
@masom, It looks really good

masom:
This merge introduce a way for loaders to return metadata and specify error conditions.

This will give thumbor a lot more flexibility for dealing with loader errors.

The change is backward-compatible with a `isinstance` check, allowing us to deprecated the old response format in thumbor 6.x.

PR #556, opened 2015-09-14
Adds a jpeg2000 optimizer to thumbor

okor:
So, as is, this works. And I feel decent about it. But it's shelling out to imagemagick which aligns well with the other optimizers. However, I believe PIL could also create jpeg2000 images directly. Open to feedback.

To try it out, uncomment the j2k optimizer and the imagemagick path in the config. Adjust j2k quality in config if you wish, I saw pretty good results at 41. And request an image from thumbor, throwing in j2k to a format filter. Something like this: `http://localhost:8888/unsafe/0x200/filters:format(j2k)/http://cdn0.vox-cdn.com/uploads/chorus_image/image/47144834/Screen_Shot_2015-09-09_at_8.55.52_PM.0.0.png`

You can test that in Safari.

### DIFF
--- a/thumbor/optimizers/j2k.py
+++ b/thumbor/optimizers/j2k.py
@@ -16,13 +16,14 @@ from thumbor.optimizers import BaseOptimizer
 class Optimizer(BaseOptimizer):
 
     def should_run(self, image_extension, buffer):
-        return 'j2k' in self.context.request.filters
+        return 'jpg' in image_extension or 'jpeg' in image_extension and 'j2k' in self.context.request.filters
 
     def optimize(self, buffer, input_file, output_file):
+        imagemagick_path = self.context.config.IMAGEMAGICK_PATH
         command = '%s %s -quality %s j2k:%s' % (
-            '/usr/local/bin/convert',
+            imagemagick_path,
             input_file,
-            self.context.config.J2K_QUALITY or '100',
+            self.context.config.J2K_QUALITY or '80',
             output_file,
         )
         subprocess.call(command, shell=True)
--- a/thumbor/optimizers/j2k.py
+++ b/thumbor/optimizers/j2k.py
@@ -1,0 +1,28 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/globocom/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com timehome@corp.globo.com
+
+import os
+import subprocess
+
+from thumbor.optimizers import BaseOptimizer
+
+class Optimizer(BaseOptimizer):
+
+    def should_run(self, image_extension, buffer):
+        return 'j2k' in self.context.request.filters
+
+    def optimize(self, buffer, input_file, output_file):
+        command = '%s %s -quality %s j2k:%s' % (
+            '/usr/local/bin/convert',
+            input_file,
+            self.context.config.J2K_QUALITY or '100',
+            output_file,
+        )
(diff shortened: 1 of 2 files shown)


heynemann:
Thanks a lot for the PR, but I think this should go to [thumbor-plugins](https://github.com/thumbor/thumbor-plugins).

We are trying to have thumbor's codebase be as small as possible.

okor:
@heynemann OK. That's highly irritating. Why was this repo not mentioned earlier when I was asking about how I should proceed and why wasn't it mentioned in the docs at all? The only plugins mentioned are storage related. You have a number of optimizers in the Thumbor core code and now you have a few sprinkled into this previously unmentioned repo for plugins (which looks a lot like a storage closet where you stuff things you don't care about but don't want to throw away).

If it's not going into core and you don't even mention plugins, why am I even bothering to share my code?

heynemann:
I'm very sorry you feel this way, @okor. 

We highly value the contributions each individual makes to thumbor, as I'm sure other committers / contributors can confirm.
 
Personally, I try to go out of my way to help people contribute. I'm even willing to get your PR and push it to thumbor-plugins myself. I do feel, though, that it would be best if you did it yourself.

The contribution to thumbor-plugins would be associated with your user, and that's very important IMHO.

I'm terribly sorry that we're failing at communicating the structure of thumbor's projects. We definitely need to get something in the docs that specify the projects we maintain as well as the ones maintained by thumbor community.

As for the other optimizers in thumbor's codebase, the rule-of-thumb here is that  thumbor should ALWAYS include a reference implementation for each new feature we implement. 

It's a consensus among committers, though, that thumbor's main codebase NEEDS to shrink. We removed the additional imaging engines (opencv and graphicsmagick) and we're in the process of removing additional storages (redis and mongodb). That's why I'm a little worried of adding any other optimizers to thumbor's main codebase.

There are a couple ways we can move forward:
a) You can create a repository for your own optimizer;
b) You can contribute the optimizer to thumbor-plugins;
c) You can contribute it as a project to thumbor-community (an org maintained by thumbor's users).

In any of the above, count me and @guilhermef in to help you get your contribution merged in.

If you don't have any more time to work in this PR, please let me know which way you'd like to go moving forward and I'll do anything in my power to get this done.

Once again, sorry for your less than optimal experience contributing with thumbor.

okor:
You know the project better than I and ultimately it's the call of the maintainers. And maintaining an open source project is a lot of work. So firstly, thanks for the huge amount of work you and the rest of the team have done and continue to do. It's a very awesome piece of software and a lot more people should be using it IMO. None of my criticisms change my opinion on any of that.

My concerns are:
  - encoders/optimizers feel like core functionality, low level stuff
  - general population will not benefit because the optimizers are hidden in a seperate repo
  - the plugin repo/package is not version synced, based on what I can see from the recommended install procedure listed on the plugins repo page.
    - this increases the likelyhood of breakage
    - this makes operations tasks like creating/scaling thumbor clusters more difficult and less predictable.
  - while data sets can vary, I believe it is possible to do some significant testing and select a "best" encoder/optimizer for each use case. Since Thumbor generats images for browsers and the current state of things there are exactly 5 encoder/optimizers that should be selected and no more.

Those are all solvable problems however and I understand the need to keep a project maintainable. So * I'll create a PR in the plugins repo * for each optimizer but I think we should at least consider a better way of keeping the plugins repo in sync with a specific version of Thumbor.

Thanks!

heynemann:
Thanks a lot for your criticism. I truly apreciate it. I believe we only grow in the face of better ways of doing what we do.

We are still experimenting with ways of organizing thumbor. It has grown fast and continues to grow. I completely agree with the versioning issue.

I'll give some more thought into it and we can create a separate issue to discuss thumbor's organization moving forward.

PR #557, opened 2015-09-14
Adds a pngquant optimizer to thumbor

okor:
Really straight forward pngquant optimizer for PNG's. You must have pngquant installed (use version 2).

Uncomment and adjust preferred quality setting. Uncomment pngquant optimizer. And now all PNG's will pass through pngquant.


heynemann:
Thanks a lot for the PR, but I think this should go to [thumbor-plugins](https://github.com/thumbor/thumbor-plugins).

We are trying to have thumbor's codebase be as small as possible.

PR #558, opened 2015-09-14
Adds a mozjpeg optimizer

okor:
You need to have mozjpeg installed. Uncomment mozjpeg stuff and adjust quality preference. Now all JPEG's will pass through mozjpeg.

### DIFF
--- a/thumbor/optimizers/mozjpeg.py
+++ b/thumbor/optimizers/mozjpeg.py
@@ -1,0 +1,31 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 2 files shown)


heynemann:
Why do we need imagemagick installed to use this?

okor:
@heynemann Because there is an intermediate step, where the image is converted to PPM format. Mozjpeg only supports a few obscure formats. So we convert to ppm then pipe to mozjpeg https://github.com/thumbor/thumbor/pull/558/files#diff-90e0753b56dd9e80644256a7ab21fafaR24

PR #560, opened 2015-09-18
Add exif orientation

OndrejIT:
Add rotate(exif) ...

### DIFF
--- a/thumbor/filters/rotate.py
+++ b/thumbor/filters/rotate.py
@@ -13,8 +13,10 @@ from thumbor.filters import BaseFilter, filter_method
 
 class Filter(BaseFilter):
 
-    @filter_method(BaseFilter.Number)
+    @filter_method(r'90|180|270|exif')
     def rotate(self, value):
-        if value % 90 == 0:
-            value = value % 360 # To optimize for engines
+        if value == 'exif':
+            self.engine.reorientate()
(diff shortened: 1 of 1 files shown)


masom:
Thanks @OndrejIT 
Would you mind adding a test for the `exif` value? Right now it tests for numbers but won't ensure something does not break if `exif` is passed.

sbaechler:
Does this make sense? A rotation filter and EXIF rotation are different things. One alters the image while the other just sets it right. I suggest they be kept separate. To apply exif rotation, simply set `config.RESPECT_ORIENTATION` to True.

There is another problem with the way Exif orientation is handled in Thumbor though: Thumbor reorientates the image somewhere in the middle of the process cycle. There is some metadata information such as "[mwg-rs:](http://www.metadataworkinggroup.com/specs/)Regions" which contains coordinates of image regions relative to the original image. A filter or processor that uses that data must know if the image has been rotated (or scaled) yet. 
So `engine.get_orientation()` should return a different value after `reorientate()` has been applied to the image.

sbaechler:
I suggest setting the default of `config.RESPECT_ORIENTATION` to True. Since most browsers don't support the embedded EXIF tag there is no reason not to rotate the image on the server. Also, the face detection algorithms only work with upright images.

PR #569, opened 2015-10-02
Set content type based on mime

okor:
This PR will set the `Content-Type` header based on the actual MIME type of whatever is returned after all processing. This will help greatly with a new optimizer I'm working on that sometimes returns JPEG's when you request a PNG's source image and apply the optimizer - if the source PNG doesn't contain any transparent pixels. Should also remove the burden of setting the correct MIME type from future plugins. Only caveat is that they must be defined in the `BaseEngine` class's `get_mimetype` method.

It's somewhat ugly code but seems to work. Open to suggestions, otherwise here it is.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -241,9 +241,9 @@ class BaseHandler(tornado.web.RequestHandler):
             )
         if not context.request.meta:
             results = self.optimize(context, image_extension, results)
-        if context.request.format:
-            # An optimizer might have modified the image format.
-            content_type = CONTENT_TYPE.get('.%s' % context.request.format, content_type)
+        # An optimizer might have modified the image format.
+        content_type = BaseEngine.get_mimetype(results)
+
         return results, content_type
 
     def _process_result_from_storage(self, result):
@@ -281,7 +281,7 @@ class BaseHandler(tornado.web.RequestHandler):
             context.config.RESULT_STORAGE_STORES_UNSAFE or not context.request.unsafe)
 
         def inner(future):
-            results, content_type = (future.result()[0], BaseEngine.get_mimetype(future.result()[0]))
+            results, content_type = future.result()
             self._write_results_to_client(context, results, content_type)
             if should_store:
                 self._store_results(context, results)
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -281,7 +281,7 @@ class BaseHandler(tornado.web.RequestHandler):
             context.config.RESULT_STORAGE_STORES_UNSAFE or not context.request.unsafe)
 
         def inner(future):
-            results, content_type = future.result()
+            results, content_type = (future.result()[0], BaseEngine.get_mimetype(future.result()[0]))
             self._write_results_to_client(context, results, content_type)
             if should_store:
                 self._store_results(context, results)


heynemann:
Let's wait to see what other committers think. I don't recal why we have a tuple being returned here...

okor:
Sounds good to me, I'm in no rush.

thumbor/handlers/__init__.py
```diff
@@ -281,7 +281,7 @@ def finish_request(self, context, result_from_storage=None):
             context.config.RESULT_STORAGE_STORES_UNSAFE or not context.request.unsafe)
 
         def inner(future):
-            results, content_type = future.result()
+            results, content_type = (future.result()[0], BaseEngine.get_mimetype(future.result()[0]))
```
guilhermef: Don't you guys think it's better to read the result content type here:
https://github.com/okor/thumbor/blob/master/thumbor/handlers/__init__.py#L244
since it runs inside a thread pool ?

okor: My only semi-strong opinion is that we should probably only set it once. So if it stays in the `inner` of `finish_request` we should remove https://github.com/okor/thumbor/blob/master/thumbor/handlers/__init__.py#L244 since it will be overridden anyways. 

Also, @guilhermef, not sure what you mean by pointing out that this runs in a thread pool ... seems like both the `finish_request` and `_load_result` execute within the context of the threadpool https://github.com/okor/thumbor/blob/master/thumbor/handlers/__init__.py#L290-L291 Is there some kind of thread safety or race condition concern here?

@guilhermef @heynemann You two have so much more experience with Thumbor code. I defer the choice of where to perform the content type setting to you two.

Also, I just realized that I'll need to add `jpeg2000` to the get_mimetype method, so the jp2 plugin will work as expected. This is the downside of setting the mimetype in one place in thumbor core. Then again, it's rare that we need to add a new MIME type so maybe not a big deal.

masom: It would be awesome to allow plugins to register new mime types. This was something we wanted to accomplish in thumbor-community/core

If we had a main static class for the mimetype library plugins could do something like:
```
from thumbor import MimeTypes

MimeTypes.register('image/gif', '.gif')
```

Instead of relying on a static list.

okor: @masom I almost suggested that but since it's so rare that this happens, I wasn't sure if it would make sense to do that work. But I also wouldn't stop anyone for doing it.

guilhermef: @okor by moving the BaseEngine.get_mimetype to inside the _load_results, you would avoid getting the content_type twice from different sources: https://github.com/okor/thumbor/blob/master/thumbor/handlers/__init__.py#L246
And I think that the callback runs synchronously.

okor: @guilhermef I'll keep digging, but I tried to put the BaseEngine call into the `if context.request.format` block right here https://github.com/okor/thumbor/blob/master/thumbor/handlers/__init__.py#L246 but it's not being executed. I added a debug print statement and `context.request.format` is returning None. But the plugin won't even run if that was the case (it's definitely running and modifying the MIME type). So something seems amiss. Any ideas?

okor: For the "auto" plugin, I have `'auto' in self.context.request.filters` in the `should_run` call, so it's somewhat different that just checking the `context.request.format`

okor: Ah, I see. 'auto' is not in `ALLOWED_FORMATS` so it's returning None. So, seems like we should drop the `if context.request.format` condition and just always set the MIME type. Seem logical?

okor:
@guilhermef Went ahead and made the adjustment based on your feedback. Let me know if that seems good.

guilhermef:
@heynemann @masom I'm happy with this PR, what do you guys think ?

masom:
:+1:

okor:
Thanks!

heynemann:
Good job everyone. Sorry for not contributing. I was moving to a new house yesterday. Things still kinda rough with a lot of unboxing today.

PR #572, opened 2015-10-05
Adds the jp2 aka jpeg2000 MIME type

okor:
Really straight forward.

See the RFC MIME spec http://www.rfc-editor.org/rfc/rfc3745.txt or consult your own source.

Alternatively, test locally from thumbor dir, do something like this:
```
(thumbor) ~/Projects/voxmedia/thumbor $ python
from thumbor.engines import BaseEngine
buffer = open('/Users/jason/Desktop/fromverge-small.jp2').read()
BaseEngine.get_mimetype(buffer)
'image/jp2'
```


masom:
:+1:

okor:
Thanks!

PR #573, opened 2015-10-06
Allow result_storages to be asynchronous

masom:
The `result_storage` are currently using synchronous operations.


Bladrak:
:+1:

masom:
@heynemann Seems like we forgot the `result_storages` when moving to async `storages` :(

This should be backward-compatible.

heynemann:
As soon as @guilhermef takes a look I'm fine with merging! Once again good job @masom

guilhermef:
Looks good

PR #574, opened 2015-10-07
Fixed async result storage with last modified header

Bladrak:
Following up #573

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -250,35 +250,35 @@ class BaseHandler(tornado.web.RequestHandler):
 
         return results, content_type
 
+    @gen.coroutine
     def _process_result_from_storage(self, result):
         if self.context.config.SEND_IF_MODIFIED_LAST_MODIFIED_HEADERS:
             # Handle If-Modified-Since & Last-Modified header
             try:
-                result_last_modified = self.context.modules.result_storage.last_updated()
+                result_last_modified = yield gen.maybe_future(self.context.modules.result_storage.last_updated())
+
+                if result_last_modified:
+                    if 'If-Modified-Since' in self.request.headers:
+                        date_modified_since = datetime.datetime.strptime(self.request.headers['If-Modified-Since'], HTTP_DATE_FMT)
+
+                        if result_last_modified <= date_modified_since:
+                            self.set_status(304)
+                            self.finish()
+                            return
+
+                    self.set_header('Last-Modified', result_last_modified.strftime(HTTP_DATE_FMT))
             except NotImplementedError:
                 logger.warn('last_updated method is not supported by your result storage service, hence If-Modified-Since & '
                             'Last-Updated headers support is disabled.')
 
-            if result_last_modified:
-                if 'If-Modified-Since' in self.request.headers:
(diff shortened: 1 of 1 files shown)


Bladrak:
@masom seems we forgot the last-modified option when putting the result_storage async (given it's optional, it's not in the default result_storage)

masom:
I wonder if we should update the `result_storage` to use a value object ( like storages )  that would contain some metadata.

Right now to get the last-modified value we make two calls to the result_storage when most backends support telling that value when the result is fetched.

Bladrak:
Yes, I was wondering that as well when working on async support for aws connector, where I put a local cache within the class to avoid doubling the calls (not committed yet) as a temporary solution. However, updating the result_storage to return a value object instead of a binary might lead to important BC Breaks, wouldn't it?

masom:
We would be able to support it with an `isinstance` check like in https://github.com/thumbor/thumbor/commit/7992b4a66b48b3f76acb3428b0f53a62efde7d01

If the return value isn't a `ResultStorageResult` then a call can be made for `last_modified`. That method could be flagged as `@deprecated`.

Bladrak:
At the moment Storages don't seem to handle ValueObject, maybe we should have a common class to handle both ``StorageResult`` & ``ResultStorageResult``? Which could be common to ``LoaderResult`` maybe?

masom:
I would assume storages and result_storages could use the same `StorageResult` object since there is a high chance they both provide the same kind of metadata.

`LoaderResult` should probably be separate from that ( the class is minimal and re-use not very efficient right now ).

As a start I would move the result_storage to return a `StorageResult` and also handle the current behaviour. A few `isinstance` will work until we can deprecate the old usage.

`StorageResult` could be based on the `LoaderResult` class.

Bladrak:
Ok, I'll try to work on that then :)

masom:
@Bladrak it should probably be a separate PR. This one should probably be merged.

Bladrak:
Fine with me

thumbor/handlers/__init__.py
```diff
@@ -250,16 +250,13 @@ def _load_results(self, context):
...
         if self.context.config.SEND_IF_MODIFIED_LAST_MODIFIED_HEADERS:
             # Handle If-Modified-Since & Last-Modified header
             try:
-                result_last_modified = self.context.modules.result_storage.last_updated()
-            except NotImplementedError:
-                logger.warn('last_updated method is not supported by your result storage service, hence If-Modified-Since & '
-                            'Last-Updated headers support is disabled.')
+                result_last_modified = yield gen.maybe_future(self.context.modules.result_storage.last_updated())
 
-            if result_last_modified:
```
masom: Removing this if might make 
`if result_last_modified <= date_modified_since` succeed/ fail even if it should not.

if `result_last_modified` is `None` then if clause is always true.

Bladrak:
@masom fixed

masom:
:+1: Tested locally and it works.

guilhermef:
@Bladrak @masom you guys are awesome

heynemann:
The issue was created today, discussed today, pull requested today and merged today... Jeez I hope I'll be able to keep up with you guys! Awesome job! Really impressed. Great to have you both onboard!

PR #578, opened 2015-10-14
clean up PILLOW_JPEG-options

seichner:
I was not happy with the confusing options for PIL-JPEG settings PILLOW_JPEG_SUBSAMPLING (but not PILLOW_JPEG_QTABLES), and PILLOW_COPY_JPEG_SETTINGS (which overrides both QUALITY and PILLOW_JPEG_SUBSAMPLING).

So i tried to clean it up:
* allows explicit setting or keeping for both subsampling and qtables (PILLOW_JPEG_SUBSAMPLING, PILLOW_JPEG_QTABLES)

* removes PILLOW_COPY_JPEG_SETTINGS, as it can now be achieved by setting both SUBSAMPLING and QTABLES to 'keep'


seichner:
happily looking fwd to any feedback! :)

seichner:
cc @phoet

PR #584, opened 2015-10-22
Loaders

hhankj2u:
I have a problem when using Thumbor with my project.Most images are load from same location of Thumbor. Another images are load from API (in another server). 
So if **/path/to/image.jpg** start with **http**, Thumbor will use http loader otherwise it will use file loader.


thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+        load_sync(context, path, callback)
+
+    else:
+        result.error = LoaderResult.ERROR_NOT_FOUND
+        result.successful = False
+        callback(result)
+
+
+def validate(context, url):
+    if url.startswith('http') == True:
```
masom: This line can just be `if url.startswith('http'):`

thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+            result.successful = True
+            result.buffer = f.read()
+
+            result.metadata.update(
+                size=stats.st_size,
+                updated_at=datetime.utcfromtimestamp(stats.st_mtime)
+            )
+        callback(result)
+
+    elif path.startswith('http') == True:
```
masom: This line can just be `if url.startswith('http'):`

thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+
+@return_future
+def load(context, path, callback):
+    file_path = join(context.config.FILE_LOADER_ROOT_PATH.rstrip('/'), unquote(path).lstrip('/'))
+    file_path = abspath(file_path)
+    inside_root_path = file_path.startswith(context.config.FILE_LOADER_ROOT_PATH)
+
+    result = LoaderResult()
+
+    if inside_root_path and exists(file_path):
```
masom: Here in the if block you could call the `file_loader.load` method instead of copy/pasting it here.

thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+
+    else:
+        result.error = LoaderResult.ERROR_NOT_FOUND
+        result.successful = False
+        callback(result)
+
+
+def validate(context, url):
+    if url.startswith('http') == True:
+        res = urlparse(url)
```
masom: This portion inside the `if` block should call the `http_loader.validate` to prevent code duplication

thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+        for pattern in context.config.ALLOWED_SOURCES:
+            if re.match('^%s$' % pattern, res.hostname):
+                return True
+
+        return False
+
+    return True
+
+
+def return_contents(response, url, callback, context):
```
masom: It think this entire method could be removed if `load_sync` proxy the call to `http_loader.load_sync`

thumbor/loaders/file_fallback_http_loader.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        if response.time_info:
+            for x in response.time_info:
+                context.metrics.timing('original_image.time_info.' + x, response.time_info[x] * 1000)
+            context.metrics.timing('original_image.time_info.bytes_per_second', len(response.body) / response.time_info['total'])
+        result.buffer = response.body
+
+    callback(result)
+
+def load_sync(context, url, callback):
```
masom: This method could be removed given it is a copy of `http_loader`. The call to `load_sync` can be changed to `http_loader.load_sync`

hhankj2u:
Thank you for checking my pull request. I don't know Python. This is the first time with Python.

masom:
@hhankj2u Thanks for your contribution. It's great to see someone just starting with a new language being able to propose additions to an existing project!

I think you are doing great and there are just a few little changes that need to be made.

We've used a similar pattern in different places within thumbor with `mixed_result_storage` and other components.

hhankj2u:
@masom Thanks for your suggestion.

heynemann:
@masom @hhankj2u I think this is a great addition to thumbor, but not for the main codebase. I'd rather see this as a plugin in the contrib area. We are trying hard to get thumbor's codebase to be very thin, so I'm wary of adding more loaders, storages, result storages or engines to the main codebase.

heynemann:
And by all means @hhankj2u congratulations on tackling a new language as well as a new codebase head on. Keep up the good work!

PR #588, opened 2015-11-03
Feature/nosetests

heynemann:
The idea of this PR is to get rid of pyvows, while at the same time improving our test suite. We found some bugs, as well as several tests that didn't actually test anything.

These were the changes I could either remember or gather from the changes. Please help me get it right, so we can release thumbor 6.0.0 as soon as possible.

Breaking Changes:

* Removed Legacy Upload API;
* Removed Redis Storage (in favor of community maintained Redis storage);
* Removed Old style URLs;
* File loader does not unquote URLs before putting them in the filesystem (might lead to some storage misses for unicode URLs);

Minor Changes:
* Fixed issues with face, glasses and profile detection;
* thumbor server console app greatly simplified;
* Better usage of thumbor server console app (thumbor -v for version is an example);
* Introduced numpy as a dependency to help detection be faster and easier;
* Fixed some bugs with incompatible color schemes (CMYK to gif and some other edge cases);
* Fixed memcache storage and using hashes as keys for memcached now;
* Fixed bug with transformer that used coroutine inside non-coroutine;
* Greatly simplified URL composer (thumbor-url) code;
* Lots of PEP-8 fixes;
* Loads and loads of additional unit tests.


heynemann:
AHHHHHHHHHHH MERGED FINALLY! lol

dhardy92:
WOW :open_mouth:

heynemann:
What do you think @dhardy92?

dhardy92:
impressive work

heynemann:
Thanks man! Trying hard to get thumbor's codebase in shape. Still have some way to go, but it's getting there.

fabiomcosta:
Awesome work!
Moving pyvows to nose seems like a good idea!
It's hard to follow the whole thing because there are way too many changes included.
Did you also remove support for old opencv versions on this diff?

heynemann:
@guilhermef did some work concerning the old opencv. Right now it's not AWESOME, since we still require PIL to process images for recognition, but it's a step forward. 

We need to change to OpenCV 3, but that's going to require a lot more work since they changed the API.

Bladrak:
Great work!
It's nice to see code coverage has been greatly increased as well :) Really well done!

PR #590, opened 2015-11-04
Fix result_storage metrics by calling yield

masom:
If the result_storage is fully async the metrics will be inaccurate.


heynemann:
Ok by me!

PR #595, opened 2015-11-06
Return a 504 (instead of 404) for 599 timeout (for caching/CDN)

andrew-a-dev:
Our thumbor setup often generates `599` HTTP errors (connection time-out/reset) from upstream, and these then get passed downstream as `404` errors (because any error gets turned into a 404), as per the conversation here https://github.com/thumbor/thumbor/issues/529 , and this causes problems for our caching/CDN.

This diff (branched from the 5.2.1 tag, because that's the version we're using) adds a new `LoaderResult` type for "connection timeout", and sets it when it receives a `599` error, which then causes thumbor to send the client a `504` ("Gateway timeout") error, instead.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -110,6 +110,10 @@ class BaseHandler(tornado.web.RequestHandler):
                     # Return a Bad Gateway status if the error came from upstream
                     self._error(502)
                     return
+                elif result.loader_error == LoaderResult.ERROR_TIMEOUT:
+                    # Return a Gateway Timeout status if upstream timed out (i.e. 599)
+                    self._error(504)
+                    return
                 else:
                     self._error(500)
                     return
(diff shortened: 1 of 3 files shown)


masom:
Hi Andrew,

We've seen huge amount of 599 errors when the tornado async loop is being locked by some long running operation or the CPU being fully utilized.

Activating the threadpool helped us solve most of the 599 problems. From googling around it seems like this status code is mostly used internally by tornado when the http client could not complete within a reasonable timeout.

Have you verified the 599 timeouts came from the image source and not from within thumbor/tornado?

A little backstory on 599: https://github.com/tornadoweb/tornado/issues/1400


:+1: on the PR.

andrew-a-dev:
Hi Martin. Thanks for that!

Sorry, I should have made it more clear: yes, the 599s are coming from thumbor/tornado, *not* the actual image source (i.e., tornado is timing-out).

I think our conclusion was the same (that the async loop is blocked too-often, or for too-long), and based on a similar issue to the one you posted, we enabled the thread-pool, which seems to mitigate (but not remove) the problem. 

At least with this, we're able to stop the errors from being cached at the CDN level as 404s.

masom:
@andrew-a-dev I'll see if we can manage a separate branch for thumbor 5.2.x

Issue #598, opened 2015-11-11
URL escaping parens in filter string not supported


banks:
Hi we are using thumbor 5.2.1 (current stable version I think).

We have an issue which has been partially reported before e.g. #344 but the "fix" in that case was to go fix the other system that was adding additoinal escaping to the URL.

We've been running fine with thumbor in production for a while.

Problem is we now have requirement to send HTML emails and many mail clients (including Outlook.com (used to be called hotmail) appear to add the same additional escaping to URLs that the ticket above complained about.

Specifically in many languages all special chars are encoded not just those required to make the URL RFC compliant.

For example when we send an email with a thumb URL:

http://t01.deviantart.net/6ER_tA9w-k8lfWg4f-fB3h-Neh8=/300x200/filters:fixed_height(100,100):origin()/pre10/05c2/th/pre/i/2015/300/9/6/testing_by_petralfire-d9el86h.jpg

outlook.com adds additional escaping (presumably as part of a bad link filtering process that decodes and re-encodes the URLs):

http://t01.deviantart.net/6ER_tA9w-k8lfWg4f-fB3h-Neh8=/300x200/filters:fixed_height%28100%2c100%29:origin%28%29/pre10/05c2/th/pre/i/2015/300/9/6/testing_by_petralfire-d9el86h.jpg

The result is that obviously the HMAC security key doesn't validate and they get a 404.

If we proactively encode our URLs so the are already like the outlook mangled one, the hmac key is valid but thumbor doesn't decode them again and so fails not being able to parse the filter string (presumably).

http://t01.deviantart.net/wYl5ECvmlOIkCC88Kwk_mOJo5eQ=/300x200/filters:fixed_height%28100%2c100%29:origin%28%29/pre10/05c2/th/pre/i/2015/300/9/6/testing_by_petralfire-d9el86h.jpg

In that case we see messages in thumbor log like:

```
2015-11-11_17:46:46.91013 thumbor-1: 2015-11-11 09:46:46 thumbor:WARNING ERROR retrieving image http://filters:fixed_height(100,100):origin()/pre10/05c2/th/pre/i/2015/300/9/6/testing_by_petralfire-d9el86h.jpg: [Errno -2] Name or service not known
```

So thumbor is obviously not recognising the percent encoded filter string as a filter string at all.

This is quite an urgent issue for us so we will likely be patching thumbor to fix it. I'll add a PR when/if we do.

[...]

PR #599, opened 2015-11-12
Re-encode URL before validating HMAC to work around bad clients

banks:
As discussed on #598 this is an incomplete patch for the issue.

It's good enough for us for now so we will probably use this anyway but has issues noted in #598 that prevent this from being merged as it is. Lack of tests is one glaring issue.

It's shared as a PR for discussion.

### DIFF
--- a/thumbor/handlers/imaging.py
+++ b/thumbor/handlers/imaging.py
@@ -65,7 +65,7 @@ class ImagingHandler(ContextHandler):
         if url_signature:
             signer = self.context.modules.url_signer(self.context.server.security_key)
 
-            url_to_validate = Url.encode_url(url).replace('/%s/' % self.context.request.hash, '')
+            url_to_validate = Url.reencode_url(url).replace('/%s/' % self.context.request.hash, '')
             valid = signer.validate(url_signature, url_to_validate)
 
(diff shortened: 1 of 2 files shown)


masom:
Tricky.

Thanks for looking into the issue and explaining the behaviour.

A major issue with old urls is thumbor allows something like `http://localhost:8000/unsafe/300x200/http://localhost/some/image.jpg`. It is somewhat not RFC compliant. The source url should always be encoded to something like `http://localhost:8000/unsafe/300x200/http%3A%2F%2Flocalhost%2Fsome%2Fimage.jpg`.

heynemann:
@banks I think we are going in the direction @masom highlighted. We assume the URL is encoded and that's that. In the past we tried to be smart and every single time we got it wrong for some edge case. Now we are sticking by the RFC and it's working pretty well (as long as you encode the source URL, that is).

banks:
Fair enough although note that #598 is NOT about source URL encoding. The comment in this PR even says clearly:

> Takes the raw URL and decodes any percent encoded characters that occur before the source URL starts.

So just enforcing that source URL is encoded is not a solution to the actual issue here.

This PR is dead either way and that's fine, just wanted to be clear that your stated rationale is a different issue entirely ;)

heynemann:
Sorry for that. You are absolutely right.

PR #600, opened 2015-11-12
LastModified should be in UTC

Bladrak:
LastModified date should not depend on local timezones, this PR should fix it. However, this creates a BC Break with other result storages.

### DIFF
--- a/tests/handlers/test_base_handler.py
+++ b/tests/handlers/test_base_handler.py
@@ -13,6 +13,8 @@ import tempfile
 import shutil
 from os.path import abspath, join, dirname
 import os
+from datetime import datetime, timedelta
+import pytz
 
 import tornado.web
 from preggy import expect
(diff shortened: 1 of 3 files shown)


masom:
Great find!

Should there be a test that ensures the TZ info is returned to the browser?

Bladrak:
Maybe yes, I still need to figure out how to run thumbor's tests on my computer :) I'll try to add some.

masom:
Poor @Bladrak :D ( we see all the builds on a slack channel )

Bladrak:
Haha sorry to spam you, didn't manage to run the tests locally :(

masom:
It's all good.

Yeah the local testing environment is a little bit hard to setup right now. We are removing a bunch of plugins from thumbor to make that process easier. No more mongodb / redis / memcache will greatly help with that.

Bladrak:
Yup. I also noticed an issue with NFS sharing for the pil engine (given NFS is case insensitive, it confounds the import between the engine and the module). Maybe there's a rename to perform on the engine?

masom:
I see. We might have to normalize engine names like `pil_engine`, `gif_engine`.

Bladrak:
Yep, might be better :)

Bladrak:
@masom \o/

Bladrak:
@masom @heynemann could we have a release of thumbor so I can reference this fix as a dependency for the next ``tc_aws`` version?

masom:
We're looking at thumbor 6...

There is also a 5.x branch we could backport this fix. We'll have to update all the storages to return UTC.

Bladrak:
Yup you're right, thumbor 6 might be soon enough :) Do we know around when it will be released?

heynemann:
I'm worried about releasing 6.0.0 due to URL encoding issues.

We need to work a little further on this.

PR #601, opened 2015-11-12
Ignore escaping safe characters when quoting tornado URL

abaldwin1:
Escaping the Tornado URL breaks HTTP loading of remote images. For example, /unsafe/200x200/https://example.com/image.jpg becomes /unsafe/200x200/https%3A//example.com/image.jpg.

Answer at http://stackoverflow.com/a/845595 cites an example fix in Python's urllib. That solution has been copied in this commit.


masom:
The second portion of the url should be encoded due to unsafe characters.

The url supplied to the url signer should be like `http%3A%2F%2Fexample.com%2Fimage.jpg`, not the unencoded version.

masom:
Also, not encoding the url produces weird results if your backend expects query parameters or something else.

ex: `/unsafe/200x200/something/somehwere.jpg?key=458280x`
By the HTTP RFC the `?key=` portion should be handled by thumbor, not the backend loading the image. Encoding the entire source url prevent possible confusion on where parameters end up.

abaldwin1:
Thanks for your quick response. Perhaps I made my change in the wrong spot. I am using Thumbor Community AWS and libcurl. When supplying an escaped url to the url signer, I receive this error:

$ curl localhost/unsafe/200x200/https%3A%2F%2Fexample.com%2Fimage.jpg

2015-11-12 19:57:04 tornado.curl_httpclient: GET https%3A//example.com/image.jpg
2015-11-12 19:57:04 tornado.curl_httpclient: Could not resolve: https%3A (Domain name not found)
2015-11-12 19:57:04 tornado.curl_httpclient: Closing connection 0
2015-11-12 19:57:04 thumbor: METRICS: inc: original_image.status.599:1
2015-11-12 19:57:04 thumbor: ERROR retrieving image https%3A//example.com/image.jpg: HTTP 599: Could not resolve: https%3A (Domain name not found)
2015-11-12 19:57:04 tornado.access: 504 GET /unsafe/200x200/https%3A%2F%2Fexample.com%2Fimage.jpg (127.0.0.1) 82.50ms

It seems that in handlers/__init__.py BaseHandler.get_image, self.context.request.image_url is escaped. This seems to be causing the failure.

Would you please suggest where in the project I should focus my debugging?

masom:
cc @Bladrak

Bladrak:
Not sure that tc_aws is related to this issue, as we use directly the http_loader if need be.
Regarding your debugging, it seems that the column is not properly unencoded. Maybe you should add some logging between the two lines to see if the decoding is handled properly?

PR #618, opened 2015-11-27
Replaced the facial detection illustration

sbaechler:
I added a graphic which corresponds to the numbers in the text and also illustrates the reframing and cropping discussed in the next section.
https://github.com/sbaechler/thumbor/blob/docs/docs/detection_algorithms.rst

In this example though, one of the faces is going to be cropped.


heynemann:
Thank you very much!

PR #621, opened 2015-12-01
Add MAX_AGE False information

Bertg:
MAX_AGE can be set to False and describe the effect of this setting


guilhermef:
Thanks @Bertg

PR #626, opened 2015-12-07
Exceptions might happen inside engine.load

gi11es:
The example use case I'm running into is a failing subprocess command
happening in the engine's create_image method. If that happens, there
is no way to proceed further with the thumbnailing.


guilhermef:
@gi11es can you create a test case to ensure this behavior ?
It could be a mocked exception.

gi11es:
Will do!

gi11es:
Coverage added

guilhermef:
Thanks @gi11es

PR #632, opened 2015-12-27
Re-read storage value to allow for override

gi11es:
The use case I need this for is when a custom loader determines
that a specific kind of content (eg. videos of a certain size) shouldn't
be cached. By simply overriding the context's storage value, it can disable
storage of the original for that request.


masom:
Using a lock to prevent processing the same media concurrently is awesome!

Every requests waits for the first one to finish but benefits from the cached result, reducing the cpu load.

heynemann:
I`m late, but very good job! :)

PR #634, opened 2016-01-03
WIP - Media Object

masom:
This is a work in progress on refactoring buffers into a Media object.
- Deprecate `LoaderResult` and `ResultStorageResult` in favour of `Media`
- Pass the `Media` object instead of `buffer` to most `BaseHandler` methods.
- Backward compatible change by introducing `is_media_aware` class property on `storages` and `result_storages`
#### TODO
- [ ] Update optimizers to receive a `Media` object if they support it.
- [ ] Fix borked tests

### DIFF
--- a/thumbor/media.py
+++ b/thumbor/media.py
@@ -1,6 +1,8 @@
 # -*- coding: utf8 -*-
 
 from thumbor.engines import BaseEngine
+from thumbor.loaders import LoaderResult
+from thumbor.result_storages import ResultStorageResult
 
 class Media(object):
     def __init__(self, buffer=None, is_valid=True, metadata={}, errors=[]):
@@ -8,6 +10,30 @@ class Media(object):
         self.metadata = metadata
         self.errors = errors
         self.is_valid = True
+
+    @classmethod
+    def from_result(self, result):
+
+        if isinstance(result, Media):
+            media = result
+        elif isinstance(result, LoaderResult):
+            media = Media(result.buffer, result.successful)
+
+            if not media.is_valid:
+                media.errors.push(result.error)
+
+        elif isinstance(result, ResultStorageResult):
+            media = Media(result.buffer, result.successful)
+
+            if not media.is_valid:
+                media.errors.push(result.error)
+        else:
+            media = Media(result)
+
+        if not media.buffer:
+            media.is_valid = False
+
+        return media
 
     @property
     def content_type(self):
--- a/thumbor/media.py
+++ b/thumbor/media.py
@@ -79,4 +79,7 @@ class Media(object):
         return mime
 
     def __len__(self):
-        return len(self.buffer)
+        if self.is_valid:
+            return len(self.buffer)
+        else:
+            return 0
--- a/thumbor/media.py
+++ b/thumbor/media.py
@@ -5,11 +5,10 @@ from thumbor.result_storages import ResultStorageResult
 from thumbor.utils import logger, EXTENSION
 
 class Media(object):
-    def __init__(self, buffer=None, is_valid=True, metadata={}, errors=[]):
+    def __init__(self, buffer=None, metadata={}, errors=[]):
         self.buffer = buffer
         self.metadata = metadata
         self.errors = errors
-        self.is_valid = True
 
     @classmethod
     def from_result(self, result):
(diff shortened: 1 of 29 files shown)


masom:
@guilhermef @heynemann there is something weird with thumbor.

`master` fails a few test like this:
```
======================================================================
FAIL: test_should_be_ok_but_150x150 (tests.handlers.test_base_handler.ImageOperationsWithMaxWidthAndMaxHeight)
----------------------------------------------------------------------
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/tornado/testing.py", line 132, in __call__
    result = self.orig_method(*args, **kwargs)
  File "/Users/martin/dev/os/thumbor/tests/handlers/test_base_handler.py", line 668, in test_should_be_ok_but_150x150
    expect(response.headers['Content-Type']).to_equal('image/jpeg')
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/preggy/core.py", line 232, in _assert_topic
    return _registered_assertions[method_name](self.topic, *args, **kw)
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/preggy/core.py", line 59, in wrapper
    func(*args, **kw)
  File "/Users/martin/venvs/thumbor/lib/python2.7/site-packages/preggy/core.py", line 128, in test_assertion
    raise AssertionError(err_msg)
AssertionError: Expected topic('image/png') to equal 'image/jpeg'
-------------------- >> begin captured logging << --------------------
thumbor: DEBUG: METRICS: inc: storage.hit:1
thumbor: DEBUG: No image format specified. Retrieving from the image extension: .png.
thumbor: DEBUG: Content Type of image/png detected.
tornado.access: INFO: 200 GET /unsafe/200x200/wellsford.jpg (127.0.0.1) 6.11ms
```

Thumbor seems to think the image is a PNG. This is on OS X but passes in CI.

I'm trying to trace back the root cause on this branch and it seems to be related to the same bug.

thumbor/media.py
```diff
@@ -0,0 +1,85 @@
...
+        media = None
+
+        if isinstance(result, Media):
+            media = result
+
+        elif isinstance(result, LoaderResult) or isinstance(result, ResultStorageResult):
+            media = Media(result.buffer)
+
+            if not media.is_valid:
+                media.errors.push(result.error)
```
sbaechler: Did you mean `append()` ?

masom: Oh yes, yes.

sbaechler:
Currently the `/meta/`endpoint is broken because the Media.mime property expects an image in the buffer. I have fixed this in my branch, but I'm not that happy with the solution. I think it would be better if the JSON or Javascript responses would be separated from the media responses. So the Media object only holds media data. But I think that would mean a bigger refactor of the `handlers` module.

masom:
@sbaechler yep, I also found a few state issues depending on the code paths.

One issue is the `engine` instance keeps a state between the flows. Depending on the order of operations the engine sometimes thinks it is operating on a `PNG` when there is a GIF in the buffer.

The current shared-state approach with engine + context + handler makes it hard to figure out why thumbor is making some content decisions.

Filters are also directly tied to the engine they operate on and have access to the shared state.

The `meta` endpoint also makes the entire shared-state harder to debug because of the content-type override being passed to Pillow.

This refactor is harder than anticipated and also changes a lot more code than I would have expected.

masom:
@heynemann @guilhermef is there any current known use of the `/meta` endpoint?

heynemann:
We use it at globo.com for an image selection pop-in.

sbaechler:
There are quite a few objects that contain state. I just found a filter (redeye) that uses the request to store state. It would really help if there was a documentation of the architecture and the data flow within Thumbor. Ideally the Media object should be the only object that would be passed around and could alter its state.

masom:
@sbaechler that's the goal of this PR. Reduce shared state by using a value object.

What's great right now: Optimizers, loaders, storages.
What's not so great: Engines, Filters, Transformers.

The current shared-state of the engine (accessible by filters and transformers) allows for efficient transformations of the image buffer. Everything happens within a `pil.Image` object ( or some OpenCV construct).

masom:
@sbaechler Feel free to open PRs against this branch.

This will probably land as 7.x, allowing us to apply deeper changes to the internal structure and flow.

We'll try to release 6.x within the next two weeks ( we are accumulating a lot of changes ) without this branch.

PR #635, opened 2016-01-05
Media object

sbaechler:
I have continued working on the media-object branch.

I replaced all Result classes with the Media class. A DeprecationWarning is thrown if someone uses a Result class but you can still use them.

I removed the possibility to call Media.from_result() wit a buffer as argument. There were just to many possibilities to call this function and it wasn't clear what was going on. 

I also moved the code that adds the `application/json` mimetype to the object to the `_write_media_to_client` method. It did throw off the filters if it was set to early.

There are still a few unit tests that don't pass but all the BDD tests are passing.

I also added a few debugging statements. We can remove them once we get close to finishing this refactor.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -252,7 +252,7 @@ class BaseHandler(tornado.web.RequestHandler):
             context.request.meta_callback = context.config.META_CALLBACK_NAME \
                     or self.request.arguments.get('callback', [None])[0]
 
-        logger.debug('Metadata requested. Serving content type of %s.'
+            logger.debug('Metadata requested. Serving content type of %s.'
                      % content_type)
 
         if content_type:
@@ -496,7 +496,6 @@ class BaseHandler(tornado.web.RequestHandler):
             storage_result = yield gen.maybe_future(storage.get(url))
 
             if storage_result is not None:
-                self.release_url_lock(url)
 
                 media = Media.from_result(storage_result)
 
@@ -517,7 +516,6 @@ class BaseHandler(tornado.web.RequestHandler):
             else:  # storage_result is None
                 self.context.metrics.incr('storage.miss')
 
-            # FIXME: Diese Zeile knallt noch.
             loader_result = yield self.context.modules.loader.load(self.context, url)
             media = Media.from_result(loader_result)
 
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -237,6 +237,9 @@ class BaseHandler(tornado.web.RequestHandler):
             elif self.is_webp(context):
                 image_extension = '.webp'
                 logger.debug('Image format set by AUTO_WEBP as %s.' % image_extension)
+            else:
+                image_extension = context.request.extension
+                logger.debug('No image format specified. Retrieving from the image extension: %s.' % image_extension)
 
             content_type = CONTENT_TYPE.get(image_extension, CONTENT_TYPE['.jpg'])
 
@@ -272,7 +275,7 @@ class BaseHandler(tornado.web.RequestHandler):
 
         try:
             result = context.request.engine.read(image_extension, quality)
-            media = Media.from_result(result)
+            media = Media.from_result(result, content_type)
         except Exception as e:
             logger.error(e)
             self._error(500)
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -111,7 +111,6 @@ class BaseHandler(tornado.web.RequestHandler):
 
     @gen.coroutine  # NOQA
     def get_image(self):
-        media = 'Hallo'
 
         try:
             media = yield self._fetch(
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -13,6 +13,7 @@ import functools
 import datetime
 import pytz
 import traceback
+import warnings
 
 import tornado.web
 import tornado.gen as gen
@@ -35,8 +36,10 @@ HTTP_DATE_FMT = "%a, %d %b %Y %H:%M:%S GMT"
 
 
 class FetchResult(object):
-
     def __init__(self, normalized=False, media=None, engine=None, successful=False):
+        warnings.warn('The FetchResult class is deprecated. '
+                      'Use the Media class instead.',
+                      DeprecationWarning, stacklevel=2)
         self.normalized = normalized
         self.engine = engine
         self.media = media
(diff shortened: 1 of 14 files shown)


thumbor/handlers/__init__.py
```diff
@@ -128,25 +132,25 @@ def get_image(self):
...
             else:
+                logger.error('[BaseHandler] Media is invalid')
                 self._error(500)
                 return
 
-        normalized = fetch_result.normalized
-        media = fetch_result.media
-        engine = fetch_result.engine
+        normalized = media.normalized
+        engine = media.engine
```
masom: It's probably best to not mix the stateful `engine` with the `Media` object. This was the reason why we added `FetchResult`, allowing us to pass back an object instead of a list of values.

`FetchResult` was implemented to separate the operation concern ( fetching something ) and the result itself ( in this case a `Media` instance ).

Same with the `normalized` value.

thumbor/handlers/__init__.py
```diff
@@ -237,16 +242,22 @@ def define_image_type(self, context, result):
...
             content_type = CONTENT_TYPE.get(image_extension, CONTENT_TYPE['.jpg'])
 
         if context.request.meta:
-            context.request.meta_callback = context.config.META_CALLBACK_NAME or self.request.arguments.get('callback', [None])[0]
-            content_type = 'text/javascript' if context.request.meta_callback else 'application/json'
-            logger.debug('Metadata requested. Serving content type of %s.' % content_type)
+            context.request.meta_callback = context.config.META_CALLBACK_NAME \
+                    or self.request.arguments.get('callback', [None])[0]
+
+        logger.debug('Metadata requested. Serving content type of %s.'
```
masom: Should this line be unindented from the `if`?

thumbor/handlers/__init__.py
```diff
@@ -258,23 +269,22 @@ def _load_media(self, context):
...
-        media = None
-
         logger.debug('Reading image as {extension}'.format(
             extension=image_extension
         ))
 
         try:
             result = context.request.engine.read(image_extension, quality)
-            media = Media.from_result(result)
+            media = Media(result, mimetype=content_type)
```
masom: The `content_type` value needs to be re-evaluated from the `image_extension` or the `result` value, otherwise the engine could return an object that does not match the previously specified value.

thumbor/handlers/__init__.py
```diff
@@ -258,23 +269,22 @@ def _load_media(self, context):
...
         except Exception as e:
-            logger.error(e)
+            # logger.error(e)
+            logger.exception('BaseHandler:_load_media failed:')
```
masom: The actual exception should probably be passed in there somewhere.

thumbor/handlers/__init__.py
```diff
@@ -481,53 +495,57 @@ def _fetch(self, url):
         try:
             storage_result = yield gen.maybe_future(storage.get(url))
 
-            fetch_result.media = Media.from_result(storage_result)
+            if storage_result is not None:
+                self.release_url_lock(url)
```
masom: The lock release is already handled lower in the `finally` block

thumbor/loaders/file_loader.py
```diff
@@ -31,8 +41,10 @@ def load(context, path, callback):
 
             media.buffer = f.read()
             media.metadata.update({
-                'LastModified': datetime.utcfromtimestamp(stats.st_mtime)
+                'LastModified': datetime.utcfromtimestamp(stats.st_mtime),
+                'ContentLength': stats.st_size
```
masom: This value is actually not needed because the `buffer` itself holds the true value of the length.

thumbor/media.py
```diff
@@ -1,48 +1,85 @@
...
+        # If this is already a Media instance, pass it on.
         if isinstance(result, Media):
-            media = result
+            return result
 
-        elif isinstance(result, LoaderResult) or isinstance(result, ResultStorageResult):
-            media = Media(result.buffer)
+        elif isinstance(result, LoaderResult):
+            # try to get the mimetype from the Result
+            mime = result.get('mime')
```
masom: `LoaderResult` is an object, `get` will not work on it.

thumbor/media.py
```diff
@@ -51,35 +88,53 @@ def file_extension(self):
 
     @property
     def mime(self):
-        mime = None
+        if self._mimetype:
```
masom: `mime` is not called often and any changes to the `buffer` property has a chance of altering the value. It's probably best to not cache into `_mimetype`.

thumbor/handlers/__init__.py
sbaechler: I agree, the engine should not be stored in the Media object. But do you need to store the engine at all in the result object? It is already in the context.
What exactly is `normalized`?

thumbor/handlers/__init__.py
sbaechler: No, it belongs to the if.

thumbor/handlers/__init__.py
sbaechler: There is a lot of re-loading and re-evaluation going on and I didn't really get the oversight. Could you draw a diagram that shows the flow of the request and the image loading? This would really help any contributor.

thumbor/handlers/__init__.py
sbaechler: The exception is in the context. With logger.exception the stacktrace is logged as well. Although we should probably put it back to error once everything is working.

thumbor/loaders/file_loader.py
sbaechler: I agree. You could use len() to get the content length.

thumbor/handlers/__init__.py
masom: I'll try but I've mostly dealt with concurrency, loaders, storages, and handlers. The filters, transformers, and engines are still black box for me.

sbaechler:
@masom While I think ultimately the mime property of the Media object should be calculated only, currently there has to be a way to override it. The `meta` endpoint is using the same Media object to create its json response. I haven't really looked into that part of the framework yet so I don't know what is going on exactly.

Currently it looks like the JSON is created in the `get_image()` method of the BaseHandler. Maybe it would make more sense to do this in the `finish_request()` method instead (or anywhere at the end of the request-response cycle). So that only the response has the content type of 'application/json' and not the Media object.

masom:
@sbaechler welcome to the world of side effects and dynamic shared data :(

The `json` produced by thumbor for the `/meta` endpoint comes from pillow. It would be possible to generate that at the very end of the request cycle and return the data at that point, without touching the `Media` object.

heynemann:
Sorry guys for that ;( Probably my fault.

heynemann:
@masom anything we can do to get this in master?

masom:
@heynemann There's a lot of edge cases with the Media object and most of the issues revolve around the shared state of the engine.

@sbaechler has done a great job at improving the initial PR I proposed for the Media object, we'd need to figure out a few edge cases before this can get merged.

masom:
Let's merge this into the `media-object-again` branch and keep working on improving this.

sbaechler:
I agree. This branch is not ready for master yet. All components are expecting a shared state (the context). A media object would replace the context and abstract the engine object.

PR #636, opened 2016-01-13
Remove usage of shell=True

gi11es:
This is a security escalation vector (someone with edit rights for the
thumbor config could get arbitrary shell execution):

https://docs.python.org/3/library/subprocess.html#security-considerations

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -229,7 +229,14 @@ class Engine(BaseEngine):
         f.write(results)
         f.close()
 
-        popen = Popen("gifsicle --colors 256 %s" % tmp_file_path, shell=True, stdout=PIPE)
+        command = [
+            'gifsicle',
+            '--colors',
+            '256',
+            tmp_file_path
+        ]
+
+        popen = Popen(command, stdout=PIPE)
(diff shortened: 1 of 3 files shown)


masom:
Just an FYI but `(someone with edit rights for the
thumbor config could get arbitrary shell execution):` is true regardless of `shell=True`.

If someone can write to thumbor's configuration, it is possible to load a python file that can execute anything (ex: a filter that does `subprocess.Popen`)

masom:
@gi11es I got a case of morning, didn't see the PR failed CI.

gi11es:
I've just force-updated the commit on my fork, but I guess that doesn't work automatically when already merged. Do you prefer it this way, where you can run CI manually and force-update it here as well? Or as a follow-up commit fixing the issue (a missing comma)?

gi11es:
As for the far-fetched scenario, in what you describe, wouldn't that require being able to write python files to thumbor's python environment? All those scenarios are pretty unlikely anyway, I've just fixed this because it was easy to make the code follow best practices this way.

gi11es:
Ah, I see that you've already reverted, in that case I shall send a new PR.

PR #639, opened 2016-01-13
Remove usage of shell=True

gi11es:
This is a security escalation vector (someone with edit rights for the
thumbor config could get arbitrary shell execution):

https://docs.python.org/3/library/subprocess.html#security-considerations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,9 @@ addons:
     - libcurl4-openssl-dev
     - python-numpy
     - python-opencv
+    - libimage-exiftool-perl
     - libopencv-dev
+    - libjpeg-progs
     - libjpeg-dev
     - libpng-dev
     - libx264-dev
--- a/tests/handlers/test_base_handler.py
+++ b/tests/handlers/test_base_handler.py
@@ -15,6 +15,7 @@ from os.path import abspath, join, dirname
 import os
 from datetime import datetime, timedelta
 import pytz
+import subprocess
 
 import tornado.web
 from preggy import expect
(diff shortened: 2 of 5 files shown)


masom:
As soon CI passes this will be merged.

gi11es:
Hah, yeah, by splitting untested code into more lines I've "decreased" coverage :)

gi11es:
I'll try to add jpegtran coverage, just out of curiosity on how that can be done

masom:
@gi11es  all checks have passed. Are you still looking at adding coverage?

gi11es:
@masom I already did, the PR has been updated, jpegtran.py coverage is now 100%

guilhermef:
This PR looks really good, but like @masom said, if someone has access to your thumbor.conf, they still can run arbitrary python code.

gi11es:
Yeah, probably overkill, but it's probably better to have the code safer escaping-wise, as we don't know how people might extend it in the wild. This was flagged as "low risk" in our security team's review of Thumbor and we don't even use gifv not jpegtran at the moment. I just figured it wouldn't hurt to change it.

gi11es:
And the good news for Thumbor is that's really all they could find in terms of "security issues" in their thorough review, the rest was simply configuration advice, which for the most part match the defaults.

guilhermef:
We should probably add a warning somewhere in the docs about this thumbor.conf issue.

gi11es:
Can you just throw python code in the .conf and it would run at startup?

guilhermef:
Yes, that's how we load the thumbor.conf https://github.com/globocom/derpconf/blob/master/derpconf/config.py#L84

This way you can get the path or environment variables when it loads.

gi11es:
Good to know. Is that common practice for python services/tornado apps?

guilhermef:
A lot of python frameworks do this.
That's not really an issue, if someone gains root access to your production machine, messing with confs would be the last thing they would do.

guilhermef:
@masom this PR looks really good, what do you think ?

masom:
@guilhermef merge!

guilhermef:
Thanks @gi11es

PR #640, opened 2016-01-13
Add option to throttle requests per IP address

gi11es:
The use case is a given user with legitimate requests resulting
in high resource consumption. With these new options, that user's
requests are put on hold, favouring processing of other user's requests,
when the amount of concurrent requests for a given IP address reaches
a limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,9 @@ addons:
     - libcurl4-openssl-dev
     - python-numpy
     - python-opencv
+    - libimage-exiftool-perl
     - libopencv-dev
+    - libjpeg-progs
     - libjpeg-dev
     - libpng-dev
     - libx264-dev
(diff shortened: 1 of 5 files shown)


gi11es:
Meh, I'll look into the failure tomorrow.

masom:
@gi11es It might be better to handle this at the NGINX/Apache/Proxy level than within thumbor.

Many setups run several thumbor instances and the rate limit will not be respected between instances. a Load balancer can do a much better job.

gi11es:
Fair point, I'll try to reproduce the behavior I'm looking for with HAProxy or similar

gi11es:
HAProxy seems to do the job and solves other problems at the same time. I'm more than happy to "keep Thumbor dumb" that way :) Thanks for the reminder, it's a bad habit to try to solve everything in the piece of the stack one is currently working on.

masom:
@gi11es no worries. For example the PR you submitted that implements a lock to prevent thumbor from processing the same resource is great.

PR #643, opened 2016-01-24
refactor BaseEngine.get_orientation() so it's free from side effects

sbaechler:
The method `BaseEngine.get_orienation()` would reset the Exif orientation to 1 when called.
This should only be done in the method `BaseEngine.reorientate()`.

Added unit tests for `reorientate()` as well.


guilhermef:
@masom @heynemann, It looks good, what do you guys think ?

masom:
:+1:

gabn88:
Thanks for this fix! :+1:

Issue #549, opened 2015-09-01
Add support Client Hints (DPR, Width, Resource-Width)


igrigorik:
Chrome 46 is shipping with [support for Client Hints](https://www.chromestatus.com/feature/5504430086553600). Specifically, we'll send them on an opt-in basis and can advertise device DPR, viewport width, and resource display width (when it's known) to the server to enable automated delivery and optimization of responsive images.

For additional context on motivation and developer use cases, see: 
https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints

It would be great to have CH support in thumbor! :)

masom:
:+1: 

Have you considered how CDN would implement content caching for client hints? Right now most of them use the url as cache key.

Regardless of CDN suppoprt, this seems like a great feature to implement.

igrigorik:
At a minimum the server (thumbor) should return a `Vary` header indicating which hints it used, which tells the upstream cache to use those as part of its cache key. For more details, see: 
http://tools.ietf.org/html/draft-grigorik-http-client-hints-03#section-2.2.2

Depending on the actual server or provider you use, you may also be able to customize how the cache key is computed to reduce fragmentation, etc.

heynemann:
@igrigorik have any of you guys at google (that you know of) discussed the use of this type of Vary with any of the major CDNs (Akamai, Cloudfront, etc)?

I wonder when Amazon will support varying by this kind of thing. Right now this is what's in their docs:

    Content Negotiation
    The only acceptable value for the Vary header is Accept-Encoding. CloudFront ignores other values.

from http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#ResponseCustomContentNegotiation

I was wondering whether we would be able to help thumbor users with automatic detection (the same way we do with auto-webp).

heynemann:
Apparently now you can vary by different headers in Cloudfront. Nice.

igrigorik:
As you noted, CloudFront supports Vary: http://friism.com/webp-content-negotiation-cloudfront

AFAIK, Akamai does support Vary as well, and even allows you to customize the generated cache key via additional configuration. That said, I'll defer to @yoavweiss on this one :)

colinbendell:
You are correct. On Akamai you can use the Vary header to specify a different forward origin server, change the path or add the value to the cache key.

heynemann:
[...]

PR #645, opened 2016-01-27
Add responsive capabilities

sbaechler:
This code would add two features discussed in issue https://github.com/thumbor/thumbor/issues/549

Maybe master is not the best base branch. Do you guys have a development branch?

This code would allow Thumbor to output responsive images. Now you can request:

    http://localhost:8888/unsafe/200x200/filters:responsive(2)/Test-1.jpg 

and you get a 400x400 pixel image back.

The capability is added as a filter which just sets a global `dpr` value. That value is 1 by default so it doesn't change current behavior.
If there are no arguments, then the filter checks the request headers for the `Dpr` value (from HTTP client hints) and uses this value instead. If the value is used, then the `Vary` header in the response is updated accordingly.

The dpr value is stored in context.request.dpr as request parameter. If someone has a better idea then please tell me. It needs to be somewhere that is available during the full lifecycle of the request.

There is also a new property in the context called `vary_headers` which is a Python set. Any filter or plugin that updates something so it requires its own cache key can add it to the set. In the end it gets serialized into the HTTP response header.

I added about 10 test cases, but this feature would need a lot more testing. I had to make some adjustments to the Transformer class.

I had to adjust the FilterTestCase base class as well so the filter are executed at the right time. There is still an issue with the test class because the filter has access to the Transformer class while it shouldn't have.

### DIFF
--- a/docs/filters.rst
+++ b/docs/filters.rst
@@ -18,6 +18,7 @@ Filters
    noise
    quality
    red_eye
+   responsive
    rgb
    rotate
    round_corners
--- a/docs/responsive.rst
+++ b/docs/responsive.rst
@@ -1,0 +1,33 @@
+Responsive
+==========
+
+Description
+-----------
+
+This filter adds the possibility to create responsive images.
+If this filter is active and a user requests a 320x240 image with a dpr of 2,
+then a 640x480 image could be delivered.
+
+The output size depends on other
+things as well such as the source image resolution and the network quality.
+
+
+Arguments
+---------
+
+display resolution value (dpr) (optional, number)
+This value is the relationship between a CSS pixel and a physical pixel.
+For desktop browsers this value is 1. For retina displays it is 2.
+
+Valid values are from 0.5 to 4.
+
+
+Example
+-------
+
+    ``/filters:dpr(2)/`` or ``/filters:dpr()/``
+
+
+If no argument is given, then a dpr of 1 is assumed.
+HTTP Client Hints can override the default value. If a value is passed in
+as an argument, that value is used.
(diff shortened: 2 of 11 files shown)


guilhermef:
We still have a problem with the result_storage, 
http://localhost:8888/unsafe/200x200/filters:responsive()/Test-1.jpg with dpr header 2 and http://localhost:8888/unsafe/200x200/filters:responsive()/Test-1.jpg with dpr header 1 will have the same path on the file system.

sbaechler:
True, that's a problem. It will also happen with the accept header for webp and other image formats.

One solution would be to incorporate the cache response headers into the file name generation for the storage. The cache headers would be a dictionary instead of a set. For the response headers, only the keys are used, but for the file names the values are used as well. Most likely as a hash.

guilhermef:
RIght now we're appending a webp extension: https://github.com/thumbor/thumbor/blob/master/thumbor/result_storages/file_storage.py#L79

sbaechler:
I'm closing the request until a solution with the result storage is found.

sbaechler:
I have tried using the Vary cache headers to make unique storage paths but gave up on this idea because it started to get ugly. There are to many edge cases. The main problem is that the path needs to be known at the beginning of the request while the full cache headers are only known at the end of the request cycle. Currently, the only change to the path is 'webp'. This gets added if webp is in the accepted request headers. Even if in the end no webp is delivered, e.g. if an animated gif is requested. 

I think therefore, the storage path should only be dependent on information that is available in the request: Accept header, client hints and the path. What gets delivered to the client might not be consistent with the storage path, but the storage path remains consistent from request to request.

Issue #625, opened 2015-12-04
Make image metadata available to the processors.


sbaechler:
Metadata, not just Exif but also IPTC and XMP data can contain information that could be used by Thumbor. 

As far as I can see, PIL can only read Exif data. This would mean a CLI tool such as `exiftool` would have to be used. Exif and IPTC are key-value based. XMP has a tree structure.

Possible use cases would be:

1. The focus point of an image could be set by the photographer and stored in the Exif data and would not have to be defined by whomever uploads an image into the CMS.
2. Some images should not be cropped because of copyright issues. If a metadata entry existed that prevents an image from being cropped, then it would just have padding added instead.

You could assign the issue to me. I'll try to implement a solution within the next few weeks.

masom:
There was a disucssion on creating a `Media` object to hold the image buffer + metadata at https://github.com/thumbor/thumbor/issues/577

Adding IPTC and XMP metadata would make a lot of sense.

heynemann:
+1

sbaechler:
This turned out to be more difficult than I thought.

There is a [Python XMP Toolkit](http://python-xmp-toolkit.readthedocs.org/en/latest/index.html) that can read and write XMP metadata to and from an image file. Unfortunately only to image files, not image buffers. The limitation seems to be set by the underlying C-Library Exempi.

The only way to extract XMP metadata from an image buffer for now is to do it by hand. I have created a prototype implementation for jpeg images and the PIL engine here: http://python-xmp-toolkit.readthedocs.org/en/latest/index.html#

The XMP data can then be further processed by the XMP toolkit.

My ultimate goal is to be able to write XMP data as well. If we used the Media object and the FileStorage, it would be possible to pass the file path along to a filter that would modify the meta data.

gi11es:
If that framework only deals with files, why not dump the buffer into a temp file?

sbaechler:
@gi11es This would add a dependency on the file system. I found another Python XMP Framework [Pyexiv](https://github.com/escaped/pyexiv2) which is based on a different C library. I'll give that one a try first.

masom:
@sbaechler the current filters and optimizers receive a tmp file buffer, already on the filesystem.

sbaechler:
@masom Maybe that temp file won't be available once the Media object is used.
[...]

PR #661, opened 2016-02-11
Add Metadata extraction and make this and the target size available for filters.

sbaechler:
I created a [filter for Thumbor](https://github.com/sbaechler/thumbor-universalimages) that adds the possibility for responsive art direction using the universal images XMP metadata standard.  An example of the filter output can be seen on the [demo page](http://sbaechler.github.io/universalimages-demo/).

For the filter to work I had to fork Thumbor and make some modifications to the BaseEngine and the Transformer classes. If the pyexiv2 library is available, then Thumbor extracts all metadata (not just Exif) from the image and stores it in the metadata property. Unfortunately, the Engine.__init__ method is the only place that has access to the original image metadata. Pillow strips it from the image data.

The filter needed the target size as well. I therefore added a getter method to the Transformer class and instantiated the class before the AFTER_LOAD filters are run. Those filter have access to the transformer instance and the target size.

This pull request solves issue #625.

There are two versions of the pyexiv2 library. The original one which is available for Python 2 (via homebrew or apt-get) and [py3exiv2](http://www.py3exiv2.tuxfamily.org/) for Python 3.


guilhermef:
@sbaechler Travis is not running the tests with: [pyexiv2](https://travis-ci.org/thumbor/thumbor/jobs/108654298#L1569)

sbaechler:
python-pyexiv2 and python py3exiv2 are not whitelisted in TravisCI yet. I opened a request for whitelisting them. This might take a while (usually about a week). I'll check back in a few days. Sorry for spamming the Slack channel, but the notification code is hardcoded in travis.yml.

sbaechler:
TravisCI whitelisted python-pyexiv2 and the build is passing with the new library.
https://travis-ci.org/sbaechler/thumbor/jobs/109013961

Py3exiv2 is going to take a while longer because they have to whitelist the package repository first.

jimas14:
@sbaechler How can pyexiv2 be available if py3 was introduced in thumbor v7?

sbaechler:
@jimas14 This pull request was merged 5 years ago. At that time it worked. I haven't used thumbor recently.

jimas14:
@sbaechler I understand, thanks for your response. I'm just confused how pyexiv2 was available at the time you merged this. When I try running the server with thumbor==6.7.0 and python==3.7 (so pyexiv2 can be installed) there are lots of version/syntax errors. And when I try running with thumbor==6.7.0 and python==2.7, pyexiv can't be installed due to version.

I'm trying to customize the engine init to allow IPTC to pass through, or at least copy some relevant IPTC fields into exif before returning. It's related to EU laws about copyright requirements. (https://www.legifrance.gouv.fr/codes/article_lc/LEGIARTI000038835832/2019-10-24%20%5blegifrance.gouv.fr%5d).

If you or @kkopachev have 10 mins to chat with me about this, I would greatly greatly greatly greatly appreciate it.

sbaechler:
@jimas14 When I wrote the plugin, Thumbor was still at 6.0.0 and worked with Python 2.

I cannot remember it, but here is the readme for my plugin: https://github.com/sbaechler/thumbor-universalimages. It references Python 2 and Thumbor 6.0.0.

jimas14:
@sbaechler I've seen that but didn't realize until today that `apt-get install python-pyexiv2` is how pyexiv2 is available with python 2.7

sbaechler:
@jimas14 You are welcome.

I think I also got it to work with Python 3 since I asked Travis CI for whitelisting the Py3exiv2 package. I could however not find the code.

I did find my heroku project that thad the thumbor server configuration: https://github.com/sbaechler/thumbor-rmd-demo

PR #668, opened 2016-02-17
Issues template

okor:
Upon creating a new issue, use a template to scaffold out the issue. Example: https://github.com/okor/thumbor/issues/new

See related discussion: https://github.com/dear-github/dear-github/issues/125

We can change what the templates contents are, this is just a first pass to get the ball rolling. But I think it would be really helpful.


heynemann:
:+1:

PR #679, opened 2016-02-29
Update __init__.py

gabn88:
The rw flag is needed for changing the exif data (that is done in line 261 of this file). Otherwise this function fails with AttributeError.

### DIFF
--- a/thumbor/engines/__init__.py
+++ b/thumbor/engines/__init__.py
@@ -200,7 +200,7 @@ class BaseEngine(object):
 
     def _get_exif_segment(self):
         try:
-            segment = ExifSegment(None, None, self.exif, 'ro')
+            segment = ExifSegment(None, None, self.exif, 'rw')
         except Exception:
             logger.exception('Ignored error handling exif for reorientation')
         else:


guilhermef:
@gabn88 can you create a test to validate this behavior ?

gabn88:
I'd love to, but I'm not sure about my fix. It does seem to work (as in the AttributeError is gone), but I'm not so sure about the consequences...

guilhermef:
I think that's the correct behavior, next time the changed image is read, It will have the right orientation.

gabn88:
I tried to create a test, but I don't know how to start. Maybe just integrate this as it probably was just a type to do 'ro' instead of 'rw'?

guilhermef:
@gabn88 can you post the steps to replicate the error ?

gabn88:
While using a result storage, I set 'RESPECT_ORIENTATION' to True, then uploaded a picture with some exif orientation. 

Then the error came around.

ptrln:
👍 thumbor has been broken for us due to this bug as well. we're getting a large number of thumbor crashes due to this error.

r0bdiabl0:
Any update on this?

guilhermef:
@gabn88 You can write a test based on https://github.com/thumbor/thumbor/blob/master/tests/handlers/test_upload_api.py#L65 but with RESPECT_ORIENTATION set to True and an image with exif

fmunteanu:
@guilhermef, the correct behavior is ro or rw?

guilhermef:
@fmunteanu It should be 'ro', since we're not changing the exif of the current image.
Thumbor creates a new Image.

marcelometal:
Hi! I'm closing this issue because 'pexif' was replaced by 'piexif'.

Ref: 1d7f67c732a1a3c2088a2af3537c5fd4bbdb75c9

PR #683, opened 2016-03-09
Add in .NET library to list of libraries.

boyter:
It was mentioned on the previous wiki https://github.com/thumbor/thumbor/wiki/Libraries


guilhermef:
Thanks @boyter

PR #690, opened 2016-03-16
Fixed bug on loading images which contain wierd characters like space

bosim:
If you have /unsafe/1000x1000/http://x.y.z/Test Test/test.jpg thumbor does not encode that properly (i.e. whitespace replaced with %20). This pull request fixes that bug.

### DIFF
--- a/thumbor/loaders/http_loader.py
+++ b/thumbor/loaders/http_loader.py
@@ -11,7 +11,7 @@
 import re
 from functools import partial
 from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


masom:
The source url to an image should be encoded at the client, not by thumbor.

masom:
Basically:

```
thumbor_url = "http://server/unsafe/200x200/http%3A%2F%2Fx.y.z%2FTest%20Test%2Ftest.jpg"
```

PR #710, opened 2016-04-14
New image loader with http loader fallback

werfux:
This is a much cleaner (and working) version of pull request: https://github.com/thumbor/thumbor/pull/584

Additionally I added some tests to check if the dispatching is working.


hhankj2u:
I forgot to push my change of code after read the suggestion of @masom. 
I currently use below code for my current production server and no issue occur. Any suggestion for that?
```
#!/usr/bin/python
# -*- coding: utf-8 -*-

# thumbor imaging service
# https://github.com/thumbor/thumbor/wiki

# Licensed under the MIT license:
# http://www.opensource.org/licenses/mit-license
# Copyright (c) 2011 globo.com timehome@corp.globo.com

import tornado.httpclient
from tornado.concurrent import return_future
import http_loader
import file_loader

@return_future
def load(context, path, callback, normalize_url_func=http_loader._normalize_url):
    if path.startswith('http'):
        http_loader.load_sync(context, path, callback, normalize_url_func)
    else:
        file_loader.load(context, path, callback)
```

werfux:
@hhankj2u

You can use the code you posted. But in my opinion it's better to use the same args to call load() in both cases.

That means use the http_loader.load() instead of load_snyc(). You don't need specify "http_loader._normalize_url" this will bee handled in the loader class itself.

You method should look like this:

@return_future
    if path.startswith('http'):
        http_loader.load(context, path, callback)
    else:
        file_loader.load(context, path, callback)

It's simpler and more readable.

werfux:
I made a litte refacoring and now im definitely done. I think the source is really reable.

heynemann:
Anyone want to revise this further? (@masom @okor @guilhermef)

The only thing I'd like to see done before merging this, @werfux, is docs written on this loader. You need to update the docs page on configuration and the loaders. Don't worry about the wiki, we are getting rid of it soon. As long as the in-project docs are up-to-date we are good.

werfux:
@heynemann I added a short documentation image_loader.rst. Is this what you want or should I provide more information?

heynemann:
This looks good for me. As soon as the other committers say what they think we'll integrate.

masom:
I've mostly setup thumbor clusters to be horizontally scalable and stateless but this fallback loader has been requested in the past.

Some deployment scenarios could use it, ex: a website front-end images pre-loaded on thumbor.

There's a certain overhead if thumbor will be serving files that are mostly source from the http loader vs the disk but it's negligible compared to the HTTP call.

okor:
Seems fine to me. Kinda cool actually.

PR #716, opened 2016-05-03
Let engines do their own cleanup on application shutdown

gi11es:
I have a custom engine that relies on exiftool running as a daemon, getting commands from a named pipe. When Thumbor shuts down, I need to have this engine unlink the named pipe gracefully.

Suggestions on how to do that differently are welcome, I couldn't find any other way for my engine to be made aware of application shutdown.

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -52,6 +52,9 @@ class Context:
         return self
 
     def __exit__(self, type, value, traceback):
+        if self.modules:
+            self.modules.cleanup()
+
         self.thread_pool.cleanup()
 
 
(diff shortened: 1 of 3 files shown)


guilhermef:
That's a nice solution 👍

guilhermef:
@heynemann @rfloriano What do you guys think ?

heynemann:
The only thing missing is testing... Would be nice to have a test that proves that upon exit, thumbor cleans up after itself.

gi11es:
I can add a test later today

heynemann:
Awesome!!!

gi11es:
Moved to #719 because of some mess on my own repo

PR #717, opened 2016-05-05
Remove unused simplejson dependency

gi11es:
It's not used anywhere and it gets in the way of packaging.


guilhermef:
Thanks @gi11es

PR #719, opened 2016-05-05
Let engines do their own cleanup on application shutdown

gi11es:
The test is ugly because I couldn't find a way to do it cleanly with mock, given that the engine is imported programmatically via the importer class.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -28,6 +28,7 @@ from thumbor.server import (
     run_server,
     main,
 )
+from thumbor.engines import BaseEngine
 
 from tests.fixtures.custom_error_handler import ErrorHandler as CustomErrorHandler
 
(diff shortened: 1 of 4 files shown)


tests/test_server.py
```diff
@@ -220,6 +221,9 @@ def test_can_run_main(self, ioloop_mock, get_server_parameters_mock, server_mock
         main()
         ioloop_instance_mock.start.assert_any_call()
 
+    def cleanup(self):
+        ServerTestCase.cleanup_called = True
+
     @mock.patch.object(thumbor.server, 'HTTPServer')
     @mock.patch.object(thumbor.server, 'get_server_parameters')
     @mock.patch('tornado.ioloop.IOLoop.instance', create=True)
```
guilhermef: You can use @mock.patch.object on BaseEngine, instead of setting your own method, and then use assert_called.

gi11es: Tried it, doesn't work. You're welcome to give it a try, I wasted 2-3 hours on this yesterday.

It might look like it works at some point, but try replacing assert_called with assert_not_called and both pass, which is nonsensical.

I think this is because the engine (and as a side-effect, BaseEngine) is imported programmatically by importer.py. Mock is made for imports that are declared. I had the impression that the workaround was to mock.patch  __import__ itself to catch the dynamic import, etc. but then imho that's even uglier and complicated than my solution.

guilhermef:
It looks fine 👍

Issue #721, opened 2016-05-10
Issue with smart_detector when using s3 storage


mpilar:
(Note: I initially reported this here: https://github.com/thumbor-community/aws/issues/57 )
### Thumbor request URL

http://bogus/unsafe//350x50/smart/gfr-end-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.jpg
### Expected behaviour

Should serve the image smart cropped at 350x50. **NOTE**: This only happens with /smart enabled when using s3_storage, if I use file storage it works with smart, if I use s3_storage without smart it works (see above linked issue for more details)
### Actual behaviour

Operation times out with error messages in the log.

```
ubuntu@ip-10-0-255-97:~$ thumbor -c /etc/thumbor-config.conf  -l debug
libdc1394 error: Failed to initialize libdc1394
2016-05-10 00:19:59 root:DEBUG thumbor running at 0.0.0.0:8888
2016-05-10 00:20:06 thumbor:DEBUG Cleaning key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.jpg
2016-05-10 00:20:06 thumbor:DEBUG Cleansed key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.jpg
2016-05-10 00:20:06 thumbor:DEBUG Cleaning key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.jpg
2016-05-10 00:20:06 thumbor:DEBUG Cleansed key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.jpg
2016-05-10 00:20:06 thumbor:DEBUG METRICS: inc: storage.hit:1
2016-05-10 00:20:06 thumbor:DEBUG Cleaning key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.detectors.txt
2016-05-10 00:20:06 thumbor:DEBUG Cleansed key: gfr/source-assets/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.detectors.txt
2016-05-10 00:31:41 botocore.utils:DEBUG URI updated to: https://<bucketpath>/img/2016/05/09/9524d1f3-027e-4335-97c1-b12bc462780f.detectors.txt
[...]

PR #722, opened 2016-05-10
Fix for #721

mpilar:
This is the fix I came up with for https://github.com/thumbor/thumbor/issues/721

It seems that self.do_smart_detection().result() fails if the detector takes more than an instant on asynchronous activities (such as fetching a detection file from s3 it seems?).

Please review this, I am not familiar at all with tornado so this might be a naive fix for a complex problem or might be hiding the issue.

### DIFF
--- a/thumbor/transformer.py
+++ b/thumbor/transformer.py
@@ -131,6 +131,7 @@ class Transformer(object):
     def smart_storage_key(self):
         return self.context.request.image_url
 
+    @gen.coroutine
     def smart_detect(self):
         is_gifsicle = (self.context.request.engine.extension == '.gif' and self.context.config.USE_GIFSICLE_ENGINE)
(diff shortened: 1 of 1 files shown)


ludofleury:
Any update on this?

guilhermef:
@mpilar, can you fix the tests ?

jamrizzi:
I'm waiting for this :)

mpilar:
Can someone help me with the tests? I actually do not understand well why they are failing, I think there's an expectation that the method I modified is not a coroutine and in the tests I do not see a simple way to put the method into a IOLoop that behaves correctly. I think I may have just broken more than I fixed with my changes and I do not honestly understand completely the ramifications of my simple change.

kevbaldwyn:
#864

PR #724, opened 2016-05-11
Relax version requirements

gi11es:
I don't see why it should be that restrictive. The current
pinning makes packaging Thumbor needlessly difficult.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ import logging
 
 
 tests_require = [
-    "redis==2.4.9",
+    "redis>=2.4.9",
     "gevent",
     "tornado-pyvows>=0.6.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,7 +17,7 @@
 
 
 tests_require = [
-    "redis==2.4.9",
+    "redis>=2.4.9",
```
guilhermef: I would be careful with a major release, like >=2.4.9,<3.0

gi11es: But 3.0 doesn't exist yet and such a precaution hasn't been taken on any of the other dependencies.

guilhermef:
thanks @gi11es

PR #727, opened 2016-05-19
Remove Url and Signer classes

gi11es:
They are now part of libthumbor.
Also removes the old url option and test,
as the related code was removed from libthumbor
some time ago.


guilhermef:
Looks good.
What do you think @heynemann ?

heynemann:
I don't think we can remove url signers. It will break compatibility with people who use different signers. I think @masom does, if I'm not mistaken.

masom:
I think the changes are fine given it's a module rename. We'll just have to document that bc break.

gi11es:
The CHANGELOG file looks pretty outdated, is that where you want the breaking change notice to go?

guilhermef:
@gi11es we've abandoned this changelog file.
Now we use the releases page: https://github.com/thumbor/thumbor/releases

gi11es:
Is there anything more I can do to help this PR get merged?

guilhermef:
I had to revert your commit on 6.x branch to release 6.0.2, and fix #731.

gi11es:
We could always import the classes in their old place for b/c

PR #730, opened 2016-05-26
Decouple subsampling and quality settings

### LINKED ISSUE #712: Disable jpeg subsampling and use quality filter
Hi,

I noticed when I use `PILLOW_JPEG_SUBSAMPLING = 0` to disable chroma subsampling of jpeg images, the quality filters are ignored in request URLs.

Would it be possible to separate this into two separate blocks such that quality is only ignored when setting `PILLOW_JPEG_QTABLES`?

https://github.com/thumbor/thumbor/blob/master/thumbor/engines/pil.py#L150.

vreon:
This allows independent control of `PILLOW_JPEG_SUBSAMPLING` and
`PILLOW_JPEG_QTABLES`. A practical application of this is to disable
chroma subsampling so that black-on-red text is less smudgy while still
allowing the `quality()` filter to be applied.

Fixes #712.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -145,19 +145,20 @@ class Engine(BaseEngine):
                 self.image = self.image.convert('RGB')
             else:
                 subsampling_config = self.context.config.PILLOW_JPEG_SUBSAMPLING
-                qtables_config = self.context.config.PILLOW_JPEG_QTABLES
-
-                if subsampling_config is not None or qtables_config is not None:
-                    options['quality'] = 0  # can't use 'keep' here as Pillow would try to extract qtables/subsampling and fail
+                if subsampling_config is not None:
                     orig_subsampling = self.subsampling
-                    orig_qtables = self.qtables
 
-                    if (subsampling_config == 'keep' or subsampling_config is None) and (orig_subsampling is not None):
+                    if subsampling_config == 'keep' and orig_subsampling is not None:
                         options['subsampling'] = orig_subsampling
                     else:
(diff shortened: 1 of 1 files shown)


guilhermef:
Can't you just set the qtables to None, to keep the original from the image ?

vreon:
Currently, if we set `PILLOW_JPEG_SUBSAMPLING = 0`, [any `quality()` filters in the URL get ignored](https://github.com/thumbor/thumbor/blob/be582f5620281efbee635e043537f0a7bd1a696c/thumbor/engines/pil.py#L151). I don't think setting qtables to None would help here, but let me know if I'm missing something!

guilhermef:
@heynemann, what do you think ?

heynemann:
I think we should have some form of test for this PR, @vreon. Maybe a test that compares the resulting image, so we know that it's not smudgy. If you need to look at an example, just check the filters tests. Most of them compare images using similarity analysis.

andreaugusto:
Heya, @vreon! Are you still interested in this PR? If so, as @heynemann [asked](https://github.com/thumbor/thumbor/pull/730#issuecomment-222979302), could you write a test for it to raise the coverage and also guarantee that we have the expected behaviour.

vreon:
@andreaugusto Definitely! I'll take another look at this soon.

andreaugusto:
Cool! Let us know should you need anything.

heynemann:
Closing for lack of tests. If you want this to move forward reopen and let us know! Thanks!

PR #732, opened 2016-05-29
Use pyssim instead of scikit-image for SSIM calculation

gi11es:
scikit-image is a huge library, with some equally massive dependencies like pandas. It's overkill to use that just to calculate SSIM values, when pyssim can do the same.

This change would simplify debian packaging greatly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,6 @@ install:
   - tar -C /tmp/ffmpeg-release --strip 1 -xvf /tmp/ffmpeg-release.tar.xz
   - export PATH=/tmp/ffmpeg-release:$PATH
   - ffmpeg -version
-  - "pip install scikit-image --no-binary :all:"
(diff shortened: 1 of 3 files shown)


gi11es:
Sentry error fixed here: https://github.com/thumbor/thumbor/pull/733
Details on url error here: https://github.com/thumbor/thumbor/pull/733#issuecomment-222348917

Neither of these errors/failures have anything to do with this change.

guilhermef:
nice @gi11es!!

guilhermef:
@gi11es, you can remove this: https://github.com/thumbor/thumbor/blob/master/.travis.yml#L51, from travis, to speed up the tests.

gi11es:
Ah, yes, good point! I'll update the PR.

guilhermef:
Thanks @gi11es

PR #736, opened 2016-05-31
Remove unused hiredis dependency

gi11es:
No usage of that library found anywhere in the code


guilhermef:
@gi11es you're awesome!

PR #743, opened 2016-06-02
Make dependency on preggy explicit

gi11es:
Preggy is used a bunch in the Thumbor tests. It probably works right now because some sub-dependencies of Thumbor depend on it too.
However, minimum version is important, since the test_load_with_utf8_url test in tests.loaders.test_http_loader.HttpLoaderTestCase relies on error_not_to_happen, which was introduced in Preggy 1.3.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ tests_require = [
     "remotecv",
     "pyssim",
     "cairosvg",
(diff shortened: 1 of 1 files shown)


guilhermef:
Thanks @gi11es

PR #744, opened 2016-06-03
Make YCCK sample JPEG smaller and add CMYK JPEG 

gi11es:
The existing YCCK image is 1.7MB and has a tendency to
make the integration tests time out on my VM.

I'm also renaming it, since it's not a CMYK JPEG. And
I'm adding a proper CMYK JPEG.


guilhermef:
@gi11es Can you rename this image to YCCK and add a real cmyk image ?

guilhermef:
Thanks @gi11es

PR #714, opened 2016-04-29
Add 16bit image/tiff file support

Cicero-Zhao:
for PIL cannot handle 16bit tiff file, so try to use opencv to convert 16bit tiff file to png file, and keep going on same process.

PR #746, opened 2016-06-06
PIL engine: convert tif to png before processing

kkopachev:
This is derived from https://github.com/thumbor/thumbor/pull/714
I decided to put conversion directly into PIL engine, since there might be some other engines (graphicsmagic), which does support tiffs natively.
+ added some tests


cerber717:
👍

guilhermef:
:+1:  @heynemann what do you think ?

kkopachev:
I amended commit (no changes) to force-restart travis.

guilhermef:
@kkopachev It will be a nice feature, to support tiff, we just need these tests to pass

kkopachev:
@guilhermef Yeah, trying to figure out what's going on there. Passes on my local, though. Looks like cv library version incompatibility. Trying on ubuntu precise now.

kkopachev:
@guilhermef made another commit with a fallback to cv. Looks like there is a bug in one of older python-opencv bindings http://stackoverflow.com/a/24030332

tests/engines/test_pil.py
```diff
@@ -24,6 +24,11 @@
 except ImportError:
     METADATA_AVAILABLE = False
 
+try:
+    from unittest import mock  # Python 3.3 +
```
guilhermef: @kkopachev We don't need python 3 support right now, I think it would just inflate our code base.

guilhermef:
@rfloriano @heynemann It looks good to me

kkopachev:
Any news here?

guilhermef:
@rfloriano @heynemann can I merge ?

heynemann:
thank you very much @kkopachev for your contribution!!!

rfloriano:
Thanks @kkopachev :)

PR #749, opened 2016-06-10
Make app_class definable in config

### LINKED ISSUE #747: Add ability to define the custom app via config
I'm currently trying to package thumbor-community/core for debian, separately from the thumbor package. The custom app option to hook community/core into thumbor is proving limiting for a clean systemd service definition, because it's only available as a command-line parameter.

I intend to make that option definable through the thumbor config as well (in addition to the existing command-line parameter). My first guess would be to make the command-line parameter override the new config file if both are present, and either work on its own.

If there are any objections to proposing such a change, let me know.

gi11es:
Fixes #747

The command line -a argument overrides the config
value if both are present.

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -61,6 +61,42 @@ class ContextTestCase(TestCase):
         expect(ctx.modules).not_to_be_null()
         expect(ctx.modules.importer).to_equal(importer)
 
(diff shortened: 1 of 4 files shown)


guilhermef:
Nice feature.
@heynemann @rfloriano any comments ?

PR #751, opened 2016-06-11
Add config var to specify desired png compression

kkopachev:
I noticed this option is set for jpegs, so why not for png? 
Or the idea here to let people use optimizers after main transformations?

### DIFF
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -49,6 +49,8 @@ Config.define('PILLOW_JPEG_QTABLES', None,
 
 Config.define('WEBP_QUALITY', None, 'Quality index used for generated WebP images. If not set (None) the same level of '
               'JPEG quality will be used.', 'Imaging')
+
+Config.define('PNG_COMPRESSION_LEVEL', 6, 'Compression level for generated PNG images.', 'Imaging')
 Config.define('AUTO_WEBP', False, 'Specifies whether WebP format should be used automatically if the request accepts it '
               '(via Accept header)', 'Imaging')
 Config.define('SVG_DPI', 150,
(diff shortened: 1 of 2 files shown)


guilhermef:
@kkopachev we should add ssim tests for this optimization.

kkopachev:
Not sure we need that. `optimize` seems to only tell zip_encoder in pillow to compress harder. Since png is lossless in general. Pillow [does not do anything](https://github.com/python-pillow/Pillow/blob/a0602fb68c62ccf3825b7cd074751153772cef70/PIL/PngImagePlugin.py#L684) with image.
It just passes optimize flag [down to ZipEncoder](https://github.com/python-pillow/Pillow/blob/a0602fb68c62ccf3825b7cd074751153772cef70/PIL/PngImagePlugin.py#L771-L772) which [sets max compression level](https://github.com/python-pillow/Pillow/blob/d3727b523ece4f75def72bfc960e1abd104b5dcb/libImaging/ZipEncode.c#L77) 


Apart from that, looking at the pillow source, I think there is optimize option available for gif as well :grey_question:

kkopachev:
New plan: instead of forcing max compression, expose config var. This is useful if people use more efficient optimizers after image operations. In case of PNG there is no reason to compress it, decompress and then compress it harder one more time.

kkopachev:
Any updates here?

kkopachev:
Guys, should I change something? Or explain a little more how that's useful?

masom:
Thanks @kkopachev . We were expecting a difference in output but as you stated this compression happens at the file itself and is not lossy.

PR #752, opened 2016-06-11
Use draft to speed up resize operations. Add config var for Pillow resampling filters

kkopachev:
ANTIALIAS seems to be deprecated and they use BICUBIC now in their thumbnail method. not sure if it sould be LANCZOS? http://pillow.readthedocs.io/en/stable/releasenotes/2.7.0.html#antialias-renamed-to-lanczos

I tested draft locally and there is 30% less time required to do just resize operation and save an image.

### DIFF
--- a/tests/engines/test_pil.py
+++ b/tests/engines/test_pil.py
@@ -14,9 +14,12 @@ from os.path import abspath, join, dirname
 from unittest import TestCase, skipUnless
 from preggy import expect
 
+from PIL import Image
+
 from thumbor.context import Context
 from thumbor.config import Config
 from thumbor.engines.pil import Engine
+from thumbor.engines.pil import get_resize_filter
 
 try:
     from pyexiv2 import ImageMetadata  # noqa
@@ -78,6 +81,25 @@ class PilEngineTestCase(TestCase):
             buffer = im.read()
         engine.convert_tif_to_png(buffer)
         expect(engine.extension).to_equal('.png')
+
+    def test_can_set_resampling_filter(self):
+        to_test = {
+            'LANCZOS': Image.LANCZOS,
+            'NEAREST': Image.NEAREST,
+            'BiLinear': Image.BILINEAR,
+            'bicubic': Image.BICUBIC,
+            'garbage': Image.LANCZOS,
+        }
+
+        if hasattr(Image, 'HAMMING'):
+            to_test['HAMMING'] = Image.HAMMING
+
+        for setting, expected in to_test.items():
+            cfg = Config(PILLOW_RESAMPLING_FILTER=setting)
+            expect(get_resize_filter(cfg)).to_equal(expected)
+
+        cfg = Config()
+        expect(get_resize_filter(cfg)).to_equal(Image.LANCZOS)
 
     @mock.patch('thumbor.engines.pil.cv', new=None)
     @mock.patch('thumbor.engines.logger.error')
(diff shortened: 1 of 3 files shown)


guilhermef:
I understood that draft would already resize the image.

kkopachev:
Does not seem so: https://github.com/python-pillow/Pillow/blob/ebd3c35de520c1dd99b4c0e1ef0b5d602146f670/PIL/Image.py#L1801-L1803

guilhermef:
You're right @kkopachev .
@heynemann @rfloriano what do you think ?

heynemann:
Fine by me as long as we add a test that verifies image quality (similar to filter tests).

rfloriano:
Looks good for me.
However we need to test this, image similarity tests are welcome.

kkopachev:
Struggling to find out where I should put such test. It's pil_engine specific modification and from what I can see, for jpegs it's already covered (indirectly?) by another test: https://github.com/thumbor/thumbor/blob/master/tests/handlers/test_base_handler.py#L199-L249
and for gifs - https://github.com/thumbor/thumbor/blob/master/tests/handlers/test_base_handler.py#L646-L648

homm:
> `ANTIALIAS` seems to be deprecated and they use BICUBIC now in their thumbnail method.

`ANTIALIAS` is not deprecated, it was just the wrong name of `LANCZOS` filter in Pillow prior `2.7.0`.

guilhermef:
@heynemann @rfloriano We already have tests covering this case.
Jpeg similarity through Pillow is already tested and it didn't break.

kkopachev:
@homm Oh, thanks, my bad
@guilhermef Not sure if switch to BICUBIC filter is good enought, but we've running that for a week and seems fine. 
filters

guilhermef:
@kkopachev, We can leave the filter configurable and use LANCZOS as default, what do you think ?

kkopachev:
@guilhermef Sounds good for me. Will try it next Wednesday.

kkopachev:
@gi11es I noticed you guys moved to IM with `jpeg:size` trick due to performance issues. However looking at @homm 's [benchmarks](https://github.com/uploadcare/pillow-simd#benchmarks) looks like the trick could be in `draft` call?

gi11es:
@kkopachev thanks for the pointer! I didn't know about Pillow SIMD nor about draft. The description does sound like the jpeg:size trick.

At this point our ImageMagick engine does a bit more than just that performance optimization, including some EXIF wrangling that we would have to do in a custom engine anyway. But I'll definitely give the PIL engine another spin with draft. If performance is on par or better, it might be possible to make our custom engine based on PIL again and save some code.

kkopachev:
@guilhermef updated with config variable. Let me know if I should change anything.

rfloriano:
@kkopachev, @guilhermef Looks good for me.
Can you @guilhermef take a last look on this?

homm:
@gi11es

> Also, regarding the SIMD benchmark, it would be interesting to know if IM was used with jpeg:size. If it wasn't and the Pillow part uses draft, it's not really a fair comparison.

Benchmarks are not about file storage format, they are about the raw resizing performance. So input images (which is simple 2D matrix in memory) for all libraries are equal and output images are equal. Moreover, results of work also almost pixel-by-pixel equal for tested libraries. For more accurate results and tests source code, please check [this page](https://python-pillow.org/pillow-perf/).

homm:
>> Why not make this part of the Engine class?
> I thought that's too specific to Pillow

Actually, no. All of these filters are common for convolution resampling and can be used for many libraries. Of course you'll need some mapping to native library's constants. For example: `"bicubic"` for `Image.BICUBIC` (Pillow) and `"catrom"` (wand).

PR #758, opened 2016-06-20
Allow loading of truncated images

kkopachev:


kkopachev:
Any news here? Should I expose it as config param as well?

masom:
👍 Looks good to me.

masom:
@thumbor/contributors Should this be a configurable value? I don't really see why someone would want to have thumbor completely fail the operation if the source data is corrupted.

rafaelcaricio:
Looks good to me. I would leave it as default until someone actually ask or need it to be configurable.

PR #761, opened 2016-06-22
Move integration tests out of the package

gi11es:
There's no reason to have integration tests installed when one installs thumbor.


guilhermef:
I agree, @heynemann and @rfloriano ?

heynemann:
I think they are included to make sure the tests pass. If the tests are passing without including them in the package, I'm fine with it.

gi11es:
They do run and pass when outside of the package. You can see that in the Travis run right here: https://travis-ci.org/thumbor/thumbor/jobs/142688570

guilhermef:
Thanks @gi11es

PR #762, opened 2016-06-22
Remove useless shebang in thumbor.conf

gi11es:
The file's permissions don't show it as executable anyway


guilhermef:
Thanks @gi11es

Issue #702, opened 2016-04-03
Exif orientation not respected when picture is taken with iPhone


gabn88:
If I use a image created with an iphone, the RESPECT_ORIENTATION = True flag does not put the image upright. The exif must still be there, since my photo view program does rotate the image correctly.

guilhermef:
@gabn88 can you post an image taken with an Iphone ?

gabn88:
[this](https://dl.dropboxusercontent.com/u/704385/IMG_6494.JPG) is the original.

r0bdiabl0:
I am having this issue as well.

![img_1798](https://cloud.githubusercontent.com/assets/432557/18025673/7516eab0-6bee-11e6-8667-f5cee4e355c3.JPG)

sixcycle-personal:
I encountered the same issue and looked into the thumbor logs just to find that a library called pexif is failing to parse EXIF tags from photos that are missing other tags such as the "maker" tag (which often happens with android phones given the different manufacturers and different implementations ). Thumbor sees the error and ignores the re-orientation.

One workaround is in this ticket on pexif: https://github.com/bennoleslie/pexif/issues/21

but it involves manually editing the pexif.py file to ignore these missing tags, which doesn't really hurt thumbor.

seffenberg-naspers:
is this still an issue? The latest version (6.2.1) seems to work for me on the image posted above. (pexif v0.15)

andreaugusto:
This issue is solved. For future reference, you need to set RESPECT_ORIENTATION to True. This is documented [here](http://thumbor.readthedocs.io/en/latest/configuration.html#respect-orientation).

PR #768, opened 2016-06-25
Patch 2 - Respect orientation with image that has exif data

gabn88:
This is the requested test for #702 (I think).


andreaugusto:
@gabn88, since the [issue is fixed](https://github.com/thumbor/thumbor/issues/702), the build is failing, and last comment was on 2016, I'm going to close it. Feel free to reopen it should you want to revisit it.

PR #776, opened 2016-07-05
Use generated or freely licensed test images

gi11es:
This replaces a lot of commercial images or images with unknown licensing.

"make sample_images" regenerates & redownloads the test images from their source (for the ones that can't be generated from scratch). Except for a handful in the filters tests, because it was too difficult to reproduce some filters' behavior with imagemagick.


guilhermef:
Nice @gi11es .
@rfloriano @heynemann what do you think ?

rfloriano:
@guilhermef @gi11es
Looks good for me :)

PR #777, opened 2016-07-06
measure the size of the images loaded via http

phoet:
we recently had an incident where the resource-consumption of our thumbor cluster skyrocketed and we assume that it was because of increased sizes of images that are fetched from S3. having the size of the response body would have been a great help to triage the issue.


rfloriano:
Metrics are always welcome.
Thanks @phoet

phoet:
thx!

PR #778, opened 2016-07-07
Use pipes for jpegtran optimizer to avoid temp files creation

kkopachev:
jpegtran from libjpeg-turbo progs is (almost?) useless, since it's already used in Pillow. However I found that one can install mozjpeg version of jpegtran to another location and run it to further optimize thumbs.

What was bothering me is 2 temp files creations in order to run command, pipes should be bit faster and avoid fs interaction.

### DIFF
--- a/tests/optimizers/test_jpegtran.py
+++ b/tests/optimizers/test_jpegtran.py
@@ -1,0 +1,62 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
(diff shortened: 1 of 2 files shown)


kkopachev:
Rebased against latest master and added more tests

guilhermef:
@kkopachev, you're right, that should improve the optimizer performance.
@heynemann, @rfloriano, what do you think ?

rfloriano:
Looks good for me.
Thanks @kkopachev

PR #780, opened 2016-07-14
why is this calculated twice?

phoet:
is there any reason why `mime` and `context.request.extension` are calculated twice within 10 lines of code with slight variations?

```python
         if mime is None:
             mime = BaseEngine.get_mimetype(fetch_result.buffer)
 
         self.context.request.extension = EXTENSION.get(mime, '.jpg')
 
         original_preserve = self.context.config.PRESERVE_EXIF_INFO
          self.context.config.PRESERVE_EXIF_INFO = True
  
          try:
            mime = BaseEngine.get_mimetype(fetch_result.buffer)
            self.context.request.extension = extension = EXTENSION.get(mime, None)
```

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -545,9 +545,6 @@ class BaseHandler(tornado.web.RequestHandler):
         self.context.config.PRESERVE_EXIF_INFO = True
 
         try:
-            mime = BaseEngine.get_mimetype(fetch_result.buffer)
-            self.context.request.extension = extension = EXTENSION.get(mime, None)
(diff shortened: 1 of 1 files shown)


rfloriano:
Probably not!
And the code is schizophrenic, each EXTENSION.get has a different default (None or '.jpg')

Can you fix the build?
'extension' variable need to be replaced by 'self.context.request.extension'.

phoet:
Yeah, saw that while debugging thorough the code today...

rfloriano:
Thanks @phoet

PR #787, opened 2016-07-25
Move Pillow resize filter decision method into Engine class

kkopachev:
addition for https://github.com/thumbor/thumbor/pull/752

/cc @gi11es


rfloriano:
Thanks @kkopachev!

PR #790, opened 2016-07-29
fix unquote path on file loader

oesteve:
FIle loader don't unquote urls with special characters, `http://localhost:8888/unsafe/700xorig/viva%20espa%C3%B1a.jpg` try to get `/tmp/viva%20espa%C3%B1a.jpg` in the file system and not `/tmp/viva\ españa.jpg`

### DIFF
--- a/thumbor/loaders/file_loader.py
+++ b/thumbor/loaders/file_loader.py
@@ -13,11 +13,15 @@ from datetime import datetime
 from os import fstat
 from os.path import join, exists, abspath
 from tornado.concurrent import return_future
+from urllib2 import unquote
+
 
 
 @return_future
 def load(context, path, callback):
-    file_path = join(context.config.FILE_LOADER_ROOT_PATH.rstrip('/'), path.lstrip('/'))
+
+    file_path = unquote(path).decode('utf-8')
+    file_path = join(context.config.FILE_LOADER_ROOT_PATH.rstrip('/'), file_path.lstrip('/'))
     file_path = abspath(file_path)
     inside_root_path = file_path.startswith(context.config.FILE_LOADER_ROOT_PATH)
 


rfloriano:
Hi @oesteve.
It's a really problem.
Can you add some tests for this PR?

oesteve:
@rfloriano Sorry, but I don't speak english and I'm not python developer, but I will try it:

The file loader will try to find a file with the same name of request content, including the escape characters like '%20' (space) or %C3%B1 (ñ, a common spanish character). I think that thumbor must be decode the url and find the original file name.

If I will server 'hello world.png' file, I must encode the request to 'hello%20world.png' and when I send the request 'http://localhost:8888/unsafe/origxorig/hello%20world.png', thumbor serve '/tmp/hello%20world.png' and not '/tmp/hello world.png' file from the file system.

I don't know if this is the expected behavior or this is a bug.

```bash
thumbor --version
Thumbor v6.0.1 (22-Mar-2016)

cat thumbor.conf 
LOADER = 'thumbor.loaders.file_loader'
FILE_LOADER_ROOT_PATH = '/home/oscar/workspace/thumbor_orig/images'

ls images/
hello world.png

curl -i 'http://localhost:8888/unsafe/origxorig/hello%20world.png'
HTTP/1.1 404 Not Found
Date: Sun, 31 Jul 2016 07:42:19 GMT
Content-Length: 0
Content-Type: text/html; charset=UTF-8
Server: TornadoServer/4.3

log
2016-07-31 09:42:19 tornado.access:WARNING 404 GET /unsafe/origxorig/hello%20world.png (127.0.0.1) 1.25ms

mv images/hello\ world.png images/hello%20world.png

2016-07-31 09:44:52 tornado.access:INFO 200 GET /unsafe/origxorig/hello%20world.png (127.0.0.1) 111.20ms
```

scorphus:
@oesteve I appreciate a lot your effort but we can do such a change. This will change the way the `file_loader` file loader works by default in thumbor. Which means we would break a fair amount of existing applications that already use this loader as it is. That's why those tests fail.

One option you have is to create a new file loader – say `thumbor_unquoting_file_loader` – and use it in you your setup:

```python
# ...
LOADER = 'thumbor_unquoting_file_loader'
```

oesteve:
Thanks for your approach about new file loader

PR #794, opened 2016-08-09
Fixing bug when sending a quoted char on a hash

andreaugusto:
So, I've got a problem. When someone sent a hash with a quoted char like %D1%80 the request would not finish and no log output would be printed.

The fix tries to call the quote function on a try except block and treats the error, if it happens, the thumbor way.


rfloriano:
Thanks @andreaugusto

PR #807, opened 2016-09-15
Allow loader failures to 500

gi11es:
If a loader error happens and it's not a 404, a 502 nor a 504, currently an AttributeError will happen on that condition, because engine_error doesn't exist for a FetchResult in the loader case.


scorphus:
@gi11es: Could you please provide steps to reproduce the `AttributeError`? Better yet, if possible, could you please write a test case for it? Thanks!

PR #821, opened 2016-10-26
Feature/distributed collage filter

dfleury:
This filter makes thumbor able to create a composited thumbnail with up to four images sent by filter's parameters.
At this point, the orientation of the collage and the alignment are both fixed to "horizontal" and "smart", respectively.
The future plans are:
- vertical orientation
- custom alignment
- give more power to filters by decoupling the logic of focus points and the processing of detectors from each request's context


rfloriano:
Thanks @dfleury!

PR #826, opened 2016-10-31
Make use of `MAX_SOURCE_SIZE` in the http_loader

hus787:


masom:
@hus787 we cannot accept this PR as-is because it would break current behaviour.

Would it be possible to add a guard before applying the value and keep it with a default of 0?

Basically the argument to AsyncHttpClient would be optional if the value is > 0.

andreaugusto:
Hi, @hus787! This is a needed PR since we are not using MAX_SOURCE_SIZE anywhere. Thanks! We'd like to merge it, but as @masom [asked](https://github.com/thumbor/thumbor/pull/826#issuecomment-279403602), could you:

1- Set the default value to 0 so that it does not break current behaviour. Right now it is unlimited.
2- Validate that ```MAX_SOURCE_SIZE < 0``` before using it.
3- Write a test for it.

scorphus:
One needs to be very cautious while changing the default `http_loader` (as ell as any other default loader, storage, etc) as it may potentially introduce a breaking change.

heynemann:
Any news here @hus787 ? Tests, defaults, documentation? Thanks!

heynemann:
Closing this as this appears not to be moving forward.

Issue #779, opened 2016-07-08
Can avoid Image.open for image from result storage?


kkopachev:
Hey guys!

I noticed that each time thumbor retrieves an image from result storage[ it is opening it with engine](https://github.com/thumbor/thumbor/blob/e6a5a90ac635af6f152f0938179739d344c13b09/thumbor/handlers/__init__.py#L86) which might be not so cheap. 
From what I can seem this is used [only for deciding](https://github.com/thumbor/thumbor/blob/e6a5a90ac635af6f152f0938179739d344c13b09/thumbor/handlers/__init__.py#L350-L354) if we should add `Vary` header for an image. But even that part was stripped a bit in https://github.com/thumbor/thumbor/commit/07c9dba13cc835c1a4472090218df65edc3c5184 to resolve https://github.com/thumbor/thumbor/issues/731 (code is not in master, I'm afraid it get's dropped in next release)
So currently (as of 6.0.2) it only checks if image height or width is less then limit (for already resized image :fearful:)

I am questioning if this is in any way useful currently and is there anything could be done to avoid loading already resized image just to serve it back to client?

PS What was the reason for WEBP pixel side limit?

cerber717:
👍

guilhermef:
@kkopachev we're still working on some changes to store metadata on the result storage, to avoid setting the Vary Header every time.

That's the reason behind WEBP side limit: https://developers.google.com/speed/webp/faq#what_is_the_maximum_size_a_webp_image_can_be

kkopachev:
closed via https://github.com/thumbor/thumbor/pull/839

PR #839, opened 2016-12-13
Fix gif output and optimize Vary header logic

kkopachev:
[this](https://github.com/thumbor/thumbor/issues/779) bug was really bothering me. Why do we want to unpack an image from result storage just to see if we should set `Vary: Accept` header. That's potentially CPU intensive operation.
So it was used to cover cases where we can not convert an image to webp:
- AUTO_WEBP is set to true in config
- browser sends Accept webp (was removed in release 6.x, still in remains master branch)
- it's animated gif
- we can convert that image to webp (checking max dimensions of an image)

Last 2 steps is where we want to load image to get it's metadata.
So, instead, my suggestion is that we check gif-file data and count number of frames in there. If there are >2 frames, then it's animated. That's much cheaper then loading the image either into Pillow or Gifsicle.
For max webp dimensions, well, worst case we'll add Vary header to a jpeg which could not be converted, but that's rather a rare edge case and I don't think there is much value sacrificing CPU load for that case.
Additionally I added a check that file we're trying to serve is actually an image file. I am not 100% sure, but I guess if we use gifv optimizer, we'll put mp4 file into result storage and then next time when we load it and try to open with engine - it'll fail anyway.
Plus one more check to see if format was strictly specified in the request, then there is no need to add Vary header.

Running tests locally, the one where it supposed to output animated gif, with Pillow 3.4 it outputs broken gif file, which then partially fixed by gifsicle and otput as not-animated. So that fixed as well.

Also fixed annoying error in logs when use gifsicle engine where after Image.verify it tries to close a file.

@masom you might be interested in this.

### DIFF
--- a/thumbor/engines/gif.py
+++ b/thumbor/engines/gif.py
@@ -118,8 +118,8 @@ class Engine(PILEngine):
 
         # Make sure gifsicle produced a valid gif.
         try:
-            with Image.open(BytesIO(self.buffer)) as image:
-                image.verify()
+            with BytesIO(self.buffer) as buff:
+                Image.open(buff).verify()
         except Exception:
             self.context.metrics.incr('gif_engine.no_output')
             logger.error("[GIF_ENGINE] invalid gif engine result for url `{url}`.".format(
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -11,6 +11,7 @@
 import sys
 import functools
 import datetime
+import re
 import pytz
 import traceback
 
@@ -106,16 +107,6 @@ class BaseHandler(tornado.web.RequestHandler):
             else:
                 self.context.metrics.incr('result_storage.hit')
                 self.context.metrics.incr('result_storage.bytes_read', len(result))
-
-            if result is not None:
-                buffer = result.buffer if isinstance(result, ResultStorageResult) else result
-                mime = BaseEngine.get_mimetype(buffer)
-                if mime == 'image/gif' and self.context.config.USE_GIFSICLE_ENGINE:
-                    self.context.request.engine = self.context.modules.gif_engine
-                else:
-                    self.context.request.engine = self.context.modules.engine
-                self.context.request.engine.load(buffer, EXTENSION.get(mime, '.jpg'))
-
                 logger.debug('[RESULT_STORAGE] IMAGE FOUND: %s' % req.url)
                 self.finish_request(self.context, result)
                 return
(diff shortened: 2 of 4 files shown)


kkopachev:
Just to back up my PR with some graphs: 
image
We're serving ~20 images/sec from thumbor. ~99% are coming from result storage. 4 instances running 16 thumbor processes each (no thread pool). After deploying this fix to prod CPU usage dropped in half - from 60% to 30%

thumbor/engines/gif.py
```diff
@@ -118,8 +118,7 @@ def read(self, extension=None, quality=None):
 
         # Make sure gifsicle produced a valid gif.
         try:
-            with Image.open(BytesIO(self.buffer)) as image:
-                image.verify()
+            Image.open(BytesIO(self.buffer)).verify()
```
guilhermef: Just make sure to call close() after verify.
The context manager was doing this: https://github.com/python-pillow/Pillow/blob/master/PIL/Image.py#L537

guilhermef:
@kkopachev that's an amazing improvement.
Just be careful with the Image.open without the context manager, It could leak.

thumbor/engines/gif.py
kkopachev: Ah, here it is. However, `verify` releases the pointer: https://github.com/python-pillow/Pillow/blob/ba5f9b88da66b80aa7311ade601a5caf3d05401b/PIL/ImageFile.py#L118
so when context manager was trying to close it - it throws an error. I'll figure out a better way.

guilhermef: @kkopachev You're right, fp was already None on close.
Your way is better.

kkopachev:
@guilhermef added back "with", but on different object.

masom:
@kkopachev Great job! Such a simple fix to a nasty situation.

thumbor/handlers/__init__.py
```diff
@@ -378,16 +404,29 @@ def _write_results_to_client(self, context, results, content_type):
...
+        should_vary = should_vary and not (
+            context.request.format  # format is supported by filter
+            and bool(re.search(r"format\([^)]+\)", context.request.filters))  # filter is in request
+        )
+        # our image is not animated gif
+        should_vary = should_vary and not self.is_animated_gif(buffer)
+
+        if should_vary:
+            self.set_header('Vary', 'Accept')
+            print ("Hello")
```
kkopachev: :bug:

guilhermef:
@rfloriano what do you think ?

rfloriano:
Thanks @kkopachev. Nice job! 👍

rfloriano:
@kkopachev Release 6.2.1 has you PR :)

Thanks again

PR #812, opened 2016-09-28
Fixed unhandled "'NoneType' object has no attribute 'mode'" detector exceptions.

mvdklip:
Unhandled exceptions inside the face and feature detectors caused HTTP connections to hang and 'Future exception was never retrieved' tornado error messages to be logged. 

Example problem image:
http://s3-eu-west-1.amazonaws.com/nrcapp/cbf1017e5d2f798d48f5cf07bc428f0b12835e9e

This fix handles the exception by going to the next detector.

This problem only seems to occur when USE_GIFSICLE_ENGINE = True and when using the Metadata Endpoint.

Issue #808, opened 2016-09-19
Requests hang on exceptions in engine.read


joetempleman:
If an exception is thrown in `engine.read`, the exception is not handled properly and the request hangs forever. It also takes a long time for the error to be logged and never logs the access log at all. To reproduce I just added `raise Exception()` in the beginning of the pil engine.read method (the original exception we encountered was from a custom engine). Any ideas on how to handle this error better? 
### Thumbor request URL

'localhost:8888/unsafe/smart/filters:quality(30)/https%3A%2F%2Fi.imgur.com/5Rz3K34.jpg'
### Expected behaviour

It should return a 500
### Actual behaviour

It hangs indefinitely
### Operating system

ubuntu
### Your thumbor.conf

```
FILTERS = [
    'thumbor.filters.quality'
]
ENGINE = 'thumbor.engines.pil'


STORAGE = 'thumbor.storages.no_storage'
STORAGE_EXPIRATION_SECONDS = 0
RESULT_STORAGE_EXPIRATION_SECONDS = 0
HTTP_LOADER_CURL_ASYNC_HTTP_CLIENT = True
LOADER = 'thumbor.loaders.http_loader'
```

Savar:
please see my pull request #871 for "why" this happens..

PR #841, opened 2016-12-14
Fixed unhandled "'NoneType' object has no attribute 'mode'" detector exceptions

mvdklip:
Unhandled exceptions inside the face and feature detectors cause HTTP connections to hang and 'Future exception was never retrieved' tornado error messages to be logged.

Example problem image:
http://s3-eu-west-1.amazonaws.com/nrcapp/cbf1017e5d2f798d48f5cf07bc428f0b12835e9e

This PR handles the exception by going to the next detector and replaces https://github.com/thumbor/thumbor/pull/812 because I messed that one up.

This problem only seems to occur when USE_GIFSICLE_ENGINE = True and when using the Metadata Endpoint.

### DIFF
--- a/thumbor/detectors/face_detector/__init__.py
+++ b/thumbor/detectors/face_detector/__init__.py
@@ -25,7 +25,10 @@ class Detector(CascadeLoaderDetector):
         return top
 
     def detect(self, callback):
-        features = self.get_features()
+        try:
+            features = self.get_features()
+        except Exception:
+            features = None
 
         if features:
             for (left, top, width, height), neighbors in features:
--- a/thumbor/detectors/face_detector/__init__.py
+++ b/thumbor/detectors/face_detector/__init__.py
@@ -10,6 +10,7 @@
 
 from thumbor.detectors.local_detector import CascadeLoaderDetector
 from thumbor.point import FocalPoint
+from thumbor.utils import logger
 
 HAIR_OFFSET = 0.12
 
@@ -28,7 +29,9 @@ class Detector(CascadeLoaderDetector):
         try:
             features = self.get_features()
         except Exception:
-            features = None
+            logger.warn('Error during face detection; skipping to next detector')
+            self.next(callback)
+            return
 
         if features:
             for (left, top, width, height), neighbors in features:
--- a/thumbor/detectors/feature_detector/__init__.py
+++ b/thumbor/detectors/feature_detector/__init__.py
@@ -19,12 +19,16 @@ class Detector(BaseDetector):
 
     def detect(self, callback):
         engine = self.context.modules.engine
-        img = np.array(
-            engine.convert_to_grayscale(
-                update_image=False,
-                with_alpha=False
+        try:
+            img = np.array(
+                engine.convert_to_grayscale(
+                    update_image=False,
+                    with_alpha=False
+                )
             )
-        )
+        except Exception:
+            self.next(callback)
+            return
 
         points = cv2.goodFeaturesToTrack(
             img,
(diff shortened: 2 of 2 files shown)


thumbor/detectors/face_detector/__init__.py
```diff
@@ -25,7 +25,10 @@ def __add_hair_offset(self, top, height):
         return top
 
     def detect(self, callback):
-        features = self.get_features()
+        try:
+            features = self.get_features()
+        except Exception:
+            features = None
```
masom: Could you add logging here? Silencing exceptions isn't great. Also added a metric would be awesome so things can be tracked.

thumbor/detectors/feature_detector/__init__.py
```diff
@@ -19,12 +19,16 @@ class Detector(BaseDetector):
...
+        try:
+            img = np.array(
+                engine.convert_to_grayscale(
+                    update_image=False,
+                    with_alpha=False
+                )
             )
-        )
+        except Exception:
+            self.next(callback)
```
masom: Same here.

thumbor/detectors/face_detector/__init__.py
mvdklip: Sure I could. I'm just not very sure what should be logged/tracked here. The detector is trying to get features within a meta request. To be honest I'm not sure what exactly makes this error. My best guess is that something is not being initialized as a "metadata only" performance optimalization. Any additional thoughts?

masom: At least the exception message should be logged otherwise this is silently ignoring potential problems with no way of knowing something went wrong.

masom:
@mvdklip is this ready to be merged?

mvdklip:
@masom Yes as far as I'm concerned. I moved on to a different client, but would be great to have this merged to the base Thumbor.

Savar:
please see for the "hanging" problem also #808

scorphus:
Hey, @mvdklip, any hints on how to reproduce the `'NoneType' object has no attribute 'mode'` exception? Thanks.

mvdklip:
@scorphus As mentioned in the initial comment this exception can be triggered by requesting metadata for a GIF image with USE_GIFSICLE_ENGINE enabled. A link to an example problematic image is provided, but I think it can actually be triggered by any GIF image.

scorphus:
@mvdklip Nice. Would you be willing to write such a test?

mvdklip:
@scorphus Willing yes, but to be honest I will probably never do it. I moved on to a different client which means I'm not actively working on Thumbor anymore. Sorry.

Savar:
I wonder how this exception can happen in general. Do you have a config file + real request where this error happened?

Because if `USE_GIFSICLE_ENGINE = True` then the `smart_detect()` is not really executed

```python
    def smart_detect(self):
        is_gifsicle = (self.context.request.engine.extension == '.gif' and self.context.config.USE_GIFSICLE_ENGINE)
        if (not (self.context.modules.detectors and self.context.request.smart)) or is_gifsicle:
            self.do_image_operations()
            return
```

the only other place where I find the face_detector to be used is in the filter `distributed_collage()`.. did you use that one?

mvdklip:
This URL exhibited the behaviour on both Thumbor 5.x and 6.x:

/unsafe/meta/1000x1000/smart/s3-eu-west-1.amazonaws.com/nrcapp/cbf1017e5d2f798d48f5cf07bc428f0b12835e9e

Also attaching config file:

[thumbor.conf.zip](https://github.com/thumbor/thumbor/files/834018/thumbor.conf.zip)

Savar:
@mvdklip thanks! helped me to find the root cause of your exception

PR #846, opened 2016-12-27
Fixed MAX_PIXELS error handling when the image is already in file storage

mvdklip:
When an admin decides to lower the MAX_PIXELS limit this may leave some files in file storage which were below the mentioned limit before but now above. This case was not properly handled by existing code. Without this fix Thumbor throws an unhandled exception and causes a hanging HTTP connection. A positive side effect of this fix is that it removes a lot of duplicate code.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -485,6 +485,7 @@ class BaseHandler(tornado.web.RequestHandler):
         :rtype:
         """
         fetch_result = FetchResult()
+        loader_result = None
 
         storage = self.context.modules.storage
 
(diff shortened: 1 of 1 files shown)


mvdklip:
I'd like some feedback for this. It seems like a proper fix which improves code quality at the same time.

thumbor/handlers/__init__.py
```diff
@@ -485,57 +485,45 @@ def _fetch(self, url):
...
         storage = self.context.modules.storage
 
         yield self.acquire_url_lock(url)
 
         try:
             fetch_result.buffer = yield gen.maybe_future(storage.get(url))
-            mime = None
 
             if fetch_result.buffer is not None:
                 self.release_url_lock(url)
```
Savar: is this for speed to release the url  a little bit earlier than inside the `finally:` block? Or why is this still here? (not sure if the `metrics.incr` is so slow that this duplication of code is still worth it)

thumbor/handlers/__init__.py
```diff
@@ -485,57 +485,45 @@ def _fetch(self, url):
...
-
                 self.context.metrics.incr('storage.hit')
-                mime = BaseEngine.get_mimetype(fetch_result.buffer)
-                self.context.request.extension = EXTENSION.get(mime, '.jpg')
-                if mime == 'image/gif' and self.context.config.USE_GIFSICLE_ENGINE:
-                    self.context.request.engine = self.context.modules.gif_engine
-                else:
-                    self.context.request.engine = self.context.modules.engine
-
-                raise gen.Return(fetch_result)
```
Savar: this block is gone completely and more down below a similar part is doing almost the same but also `engine.load()` the image and has additional checks. Is this the fix hear (regarding MAX_PIXELS)? If not, then this would at least change the behavior quiet a lot (which doesn't mean that it is the right thing to do, actually I have some problems overseeing all the affected parts)

thumbor/handlers/__init__.py
mvdklip: Leaving it there was not intentional. I think it can be removed.

thumbor/handlers/__init__.py
mvdklip: The fix is to always engine.load the image to make sure that the MAX_PIXELS check is actually executed and handled correctly. Otherwise Thumbor would crash later trying to engine.load the image in a different part of the application. In the process I removed duplicate code.

Savar:
so with tests I would add a LGTM but I'm not in charge :wink:

mvdklip:
I'm not going to write a test as I've moved on to a different client by now and am not working on Thumbor anymore. Sorry.

heynemann:
Wondering if this is still a problem. Closing this since it has been long since last activity. If this is still something you'd like to work on, please reopen.

PR #849, opened 2017-01-18
fixing "'Engine' object has no attribute 'exif'" exception logging

seffenberg-naspers:
the get_orientation() method is testing for the exif attribute/data
but the reorientation() method is later on doing a test which isn't
correct so it still tried to call the _get_exif_segment() method.

* fixes wrong 'if' statement `if orientation != 1 and override_exif:`
  which was also `true` in case there is no exif information at all
  by completly avoiding the reorientation logic if `orientation is None`
  (might improve performance a little bit)
* moved test for the `self.exif` attribute to the place it is actually
  used to avoid side effects


masom:
Thanks!

Issue #698, opened 2016-03-30
Set Background Color on Transparency and convert to JPG


nek4life:
I'm exploring using Thumbor for a site where we usually store product images as tiff files (or I'm thinking with Thumbor of storing them as pngs instead) then through an image service we request the image as jpg instead and set the BG color to match the site background which may vary from page to page. Is this possible with Thumbor? The fill filter only seems to fill in the edges when you resize the image.

heynemann:
You can create your own filter and modify the background easily.

nek4life:
Is there any documentation on extending the filter set? Is there an api for that or would I need to create my filter and submit a patch to thumbor to get it included? I just looked through the docs and did not find anything, but perhaps I'm missing something. Thanks!

guilhermef:
@nek4life You should take a look on the already implemented filters

vvh-empora:
so there is currently no way to set a background color for transparent area on converting from PNG to JPG ?

scorphus:
We should update `fill`'s docs and include examples with the `fill_transparent` optional argument set to `True`.

ajaypande88:
I am trying to fill transparent color inside the watermark but it doesn't work

PR #850, opened 2017-01-19
Set background color for image - useful when handling PNG->JPEG

hermanmaritz:
Solves https://github.com/thumbor/thumbor/issues/698
Added background_color filter. This sets the background layer set to specified color.
This is specifically useful when converting transparent images (PNG) to JPEG
Added unit test
Added documentation

The `fill` filter has a Boolean to "fill_transparent" however I found that this is adding additional padding. Rather than changing the `fill` filter, which might be used in this way by some. I added a new `background_color` filter to fill only the background. The "fill_transparent" parameter is not currently documented in ReadTheDocs.

See difference between `fill(f00,1)`:
`http://localhost:8888/unsafe/fit-in/500x200/filters:quality(60):format(jpeg):fill(f00,1)/http://pngimg.com/upload/water_PNG3290.png`
test_with_fill_red

And the new `background_color(f00)`:
`http://localhost:8888/unsafe/fit-in/500x200/filters:quality(60):format(jpeg):background_color(f00)/http://pngimg.com/upload/water_PNG3290.png`
test_with_background_color_red


seffenberg-naspers:
Any chance to get this included in the next upcoming release?

PR #854, opened 2017-01-26
Send HTTP 400 if a watermark filter entity cannot be fetched, instead…

noamsml:
… of throwing an error

Otherwise an error will be thrown and a 500 will be sent back to the client.

### DIFF
--- a/tests/handlers/test_base_handler.py
+++ b/tests/handlers/test_base_handler.py
@@ -258,6 +258,10 @@ class ImagingOperationsTestCase(BaseImagingTestCase):
         response = self.fetch('/unsafe/image_invalid.jpg')
         expect(response.code).to_equal(400)
 
+    def test_getting_invalid_watermark_returns_bad_request(self):
+        response = self.fetch('/unsafe/filters:watermark(boom.jpg,0,0,0)/image.jpg')
+        expect(response.code).to_equal(400)
+
     def test_can_read_monochromatic_jpeg(self):
         response = self.fetch('/unsafe/grayscale.jpg')
         expect(response.code).to_equal(200)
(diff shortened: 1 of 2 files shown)


nerdrew:
Bump. Does someone from the thumbor project want to take a look at this?

nerdrew:
@scorphus or @masom Can one of you take a look?

masom:
@nerdrew Would it be possible to add a log message? https://github.com/thumbor/thumbor/pull/854#discussion_r101041666 was added 14 days ago.

nerdrew:
@masom What kind of logs do you want?

Maybe:
```python
            logger.warn(
                    'bad watermark result error=%s metadata=%s' %
                    (result.error, result.metadata))
```

=>
```
thumbor: WARNING: bad watermark result error=not_found metadata={'updated_at': datetime.datetime(2016, 11, 28, 18, 6, 57), 'size': 1}
```

scorphus:
This is absolutely better than failing with 500 👍  But I'm still wondering whether it's better to fail with 400 or just display the image with no watermark at all.

scorphus:
As failing with 400 is, for most cases, the desired outcome, I'm merging this one. Thanks @noamsml and @nerdrew!

kkopachev:
Unfortunately, that PR broke compatibility with aws loader. 
```
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/watermark.py", line 100, in on_fetch_done
    if not result.successful:
AttributeError: 'str' object has no attribute 'successful'
```
Since AWS loader for some reason just outputs bufferr, not wrapped in LoaderResult object:
https://github.com/thumbor-community/aws/blob/master/tc_aws/loaders/s3_loader.py#L112

gingerlime:
@kkopachev we bumped into the same issue... (see #996). Shall we try to submit a PR? we're not too familiar with thumbor internals, but guess we can just make the check a bit more robust? Would appreciate any tips but this is currently breaking stuff for us on production :(

PR #858, opened 2017-02-12
Remove 'draft' option in resizing operations (quality loss issue)

okor:
Pillow's `draft` option seems to cause some pretty severe quality issues for some images. I only focused on one PNG for my tests but likely applies to other images. This fixes the problem by partially reverting the change which introduced the option: https://github.com/thumbor/thumbor/pull/752 The resampling config changes are good though.

##### Example request:
http://localhost:8888/unsafe/1220x0/cdn0.vox-cdn.com/uploads/chorus_asset/file/7953191/5_vox_cohesion_obamacare.png

##### Original Image
orig

##### Resized, using draft option
resized_with_draft

##### Resized, **without** draft option
resized_no_draft

The difference is pretty dramatic along the fine arced lines and the text. In my opinion, the quality loss is totally unacceptable.

Based on [this Pillow tutorial](https://github.com/python-pillow/Pillow/blob/master/docs/handbook/tutorial.rst#reading-in-draft-mode), the `draft` option `is only available for JPEG and MPO files`. It also appears to be [an optimization which sacrifices quality for speed](http://pillow.readthedocs.io/en/3.4.x/reference/Image.html#PIL.Image.Image.draft):
`This can often be used to speed up decoding when creating thumbnails (when speed is usually more important than quality)...`

I don't think it's a good idea to make that decision for all Thumbor users, even if the possibility for quality loss is tiny.

Additionally, I benchmarked this draft mode option in a simple script. I used the original image in both PNG and JPEG source formats. I kept I/O operations outside of the benchmarked code. For JPEGs it saved 20ms. For PNGs it saved 63ms ... but slaughtered the PNG just as it does using Thumbor. So the performance savings is pretty trivial for JPEGs and the docs indicate it may decrease quality. For PNGs the image gets messed up so I don't think it's even a valid option to consider. I recommend removing the draft option completely. 

This caused a pretty big problem for me when I recently upgraded Thumbor. I had to quickly revert to an older version due to the mentioned quality issues. I'm really surprised no one else has noticed this problem : (


okor:
So, that test is failing on a filter called `distributed_collage`. I tested this locally and the images look perfect. However, I bet that the creator of this filter used output from thumbor using his filter to base his tests on. And since the test resizes the images they probably now have a **higher** quality than they originally did when he made the filter. So while the output of thumbor compared to his test images have a higher ssim value - I bet you they are actually better quality then they were before. That is just a theory but since all of the other tests pass I bet that's it. I'll try to confirm this.

okor:
Ok, I tested all of the images with DSSIM, where <N> in <N>.png is the collage count test. The `b` variants are after my changes which remove the draft option. Url format was: `http://localhost:8888/unsafe/300x200/filters:format(png):distributed_collage(horizontal,smart,<images>/<fallback_image>` in the same order as the test specs. The DSSIM deltas are as follows.

 ~/Desktop $ dssim -o 1_diff.png 1.png 1b.png
0.00000819  1b.png
 ~/Desktop $ dssim -o 2_diff.png 2.png 2b.png
0.00000660  2b.png
 ~/Desktop $ dssim -o 3_diff.png 3.png 3b.png
0.00013622  3b.png
 ~/Desktop $ dssim -o 4_diff.png 4.png 4b.png
0.00008964  4b.png

That's quite good. I'm not sure what SSIM is finding is so different. Check out the before images (what was in the test fixtures folder of thumbor) and the after, which was thumbor output I downloaded from the thumbor responses.

##### before/after 4

4b

##### before/after 3

3b

##### before/after 2

2b


##### before/after 1

1b

I don't think there is anything wrong with the new output. I'm going to update the test images.

okor:
Another example source image that gets messed up with the `draft` option:

##### Source (https://upload.wikimedia.org/wikipedia/commons/8/86/Dulbahante.png)
dulbahante-2

##### Thumbor request url:
`http://localhost:8888/unsafe/1220x0/https://upload.wikimedia.org/wikipedia/commons/8/86/Dulbahante.png` with default config.

##### Using draft
dulbahante

##### Not using draft
dulbahante-1

Something about those fine lines. Would be interesting to here what the folks at @python-pillow Wonder if @homm or @wiredfool would be interested.

okor:
@kkopachev Maybe you should take a look at this too, since you made that collage filter.

wiredfool:
Don't use `draft` if you're worried about the quality. Especially if you're going to be processing it further. IIRC `draft` will give you a size close to that that you're requesting by throwing out the higher order terms from the JPEG. 

(Also note -- line drawings are generally a problem for JPEG. That map would be much better represented in PNG format.)

okor:
`Don't use draft if you're worried about the quality.` ^ That sounds pretty straight forward to me. There are any number of other processes the image may undergo before finally being served.

masom:
👍

masom:
@okor Just rebase when you are ready to get this merged :)

kkopachev:
LGTM. Sorry for my code being cause of the problem. Was not expecting it affect formats other then jpg.
Weird that it does something for png images at all. Could be an issue with pillow?

kkopachev:
Did some debugging and figured out that `draft` is not implemented for PNG images, so what happens is there is no conversion from 'P' to 'RGBA' mode happening at all for png. 
I'd propose 

```python
    def resize(self, width, height):
        self.image.draft(None, (int(width), int(height)))

        if self.image.mode == 'P':
            logger.debug('converting image from 8-bit palette to 32-bit RGBA for resize')
            self.image = self.image.convert('RGBA')

        resample = self.get_resize_filter()

        self.image = self.image.resize((int(width), int(height)), resample)
```
If my understanding is correct, `draft` will only be the case for jpegs, which don't have palettes anyway. For all other images it'll try to convert from palette to RGBA and `draft` won't do anything.

Thoughts?

okor:
@kkopachev Thanks for looking into that, great to know exactly what it was doing for PNGs 👍 

Other than that, my conclusion is the same - we shouldn't use this option anywhere in the thumbor codebase. The benefits in a best case scenario are very small. In the worst case scenario, even if we coded around non-JPEG images, it could still cause degraded quality. Additionally, the JPEG is very likely to go through other operations which will only exacerbate the effect. I believe the use case for the `draft` option is very limited. If you are reading, loading and resizing ... and the image is completely done at that point ... and quality loss is not an issue, then `draft` is for you. Otherwise I would not use it.

@masom Rebased on off master. Squashed my commits into a single pretty one.

homm:
> Pillow's draft option seems to cause some pretty severe quality issues for some images. I only focused on one PNG for my tests but likely applies to other images.

First of all, draft **doesn't affect PNG files somehow**. If you look at sources, [png plugin](https://github.com/python-pillow/Pillow/blob/master/PIL/PngImagePlugin.py#L512) doesn't even contain this word. And it's inherited from `ImageFile.ImageFile`, where the [draft method is empty](https://github.com/python-pillow/Pillow/blob/master/PIL/ImageFile.py#L108)

So the problem with images is **somewhere else**. Please, rename the issue to stop confusing.

As I can see from [source code](https://github.com/thumbor/thumbor/pull/858/files#diff-4766ae01206209cf72a83c6784669e00), the problem is not in the `draft` presence. The problem is that the code assumes that passing mode to the `draft` method will always convert the image do desired mode. This is false. Draft operates only on a not loaded images and only with some formats. It does noting in other cases.

okor:
@homm Absolutely, we all agree. That was early, naive analysis. I'll update the issue details now that we know more. Thank you : )

okor:
@homm Sorry for the miscategorization of the issue. Do you think the updated issue and notes now sufficiently exonerate Pillow from any wrong doings?

homm:
> The benefits in a best case scenario are very small.

Actually, It could be up to 5x faster for some usecases.

```python
In [6]: %timeit Image.open('photo.jpg').resize((200, 120), Image.LANCZOS)
1 loop, best of 3: 269 ms per loop

In [7]: %timeit Image.open('photo.jpg').thumbnail((200, 120), Image.LANCZOS)
10 loops, best of 3: 48.1 ms per loop
```

> Remove 'draft' option so PNGs get colorspace converted before resizing operations (quality loss issue)

But draft doesn't prevent colorspace conversion. You can convert image with `self.image.convert('RGBA')` and leave draft. This is two separate things.

PR #859, opened 2017-02-14
Convert 1bit PNGs before resizing

kevin1024:
When PIL resizes 1-bit images, it always falls back to the "nearest
neighbor" algorithm, which makes some pretty ugly thumbnails.

This will force thumbor to convert 1-bit images, the same way it
converts 8-bit images, before resizing.

Example of the difference:

resize

See also: #376 which does the same thing for 8-bit PNGs.


masom:
Thanks for this patch!
🐑

PR #864, opened 2017-02-21
Fix for #721 #722 (fixed unit tests)

decebal:
Following [Another pull request](https://github.com/thumbor/thumbor/pull/722) which fixes an issue we're interested in as well.

The problem lies in the unit test expecting an Exception, while decorating the function as a coroutine means it won't raise any Exceptions, instead they will be passed on `.Future` , from the docs: 
```
.. warning::

       When exceptions occur inside a coroutine, the exception
       information will be stored in the `.Future` object. You must
       examine the result of the `.Future` object, or the exception
       may go unnoticed by your code. This means yielding the function
       if called from another coroutine, using something like
       `.IOLoop.run_sync` for top-level calls, or passing the `.Future`
       to `.IOLoop.add_future`.
```

### DIFF
--- a/tests/test_transformer.py
+++ b/tests/test_transformer.py
@@ -116,8 +116,8 @@ class TransformerTestCase(TestCase):
         )
         context = test_data.to_context(detectors=[MockErrorSyncDetector], ignore_detector_error=False)
         trans = Transformer(context)
-        with expect.error_to_happen(Exception, message='x'):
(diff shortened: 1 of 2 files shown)


mpilar:
Thank you so much for this @decebal I missed that part of the documentation when I first looked at it.

masom:
Thanks @mpilar  @decebal

PR #866, opened 2017-02-22
adding another metrics class 'prometheus'

Savar:
this is an initial implementation for the prometheus exporter and will
be accessible on the configurable port PROMETHEUS_SCRAPE_PORT (default
8000)

the class tries to determine the labels (see
https://prometheus.io/docs/practices/naming/) while using a hardcoded
mapping and will prefix everything with "thumbor." to fit more the
prometheus standard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,7 @@ install:
   - make setup
   - pip install coveralls
   - pip install cairosvg
+  - pip install prometheus_client
 before_script:
   - make redis
 script:
--- a/tests/metrics/test_prometheus_metrics.py
+++ b/tests/metrics/test_prometheus_metrics.py
@@ -1,0 +1,47 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+from preggy import expect
+
+from thumbor.context import Context
+from thumbor.importer import Importer
+from thumbor.config import Config
+import thumbor.metrics
+
+from tests.base import TestCase
+
+
+class PrometheusMetricsTestCase(TestCase):
+    def get_context(self):
+        conf = Config()
+        conf.METRICS = 'thumbor.metrics.prometheus_metrics'
+        imp = Importer(conf)
+        imp.import_modules()
+        return Context(None, conf, imp)
+
+    def test_should_initialize_metrics(self):
+        expect(self.context.metrics).to_be_instance_of(thumbor.metrics.prometheus_metrics.Metrics)
+
+    def test_should_not_fail_on_use(self):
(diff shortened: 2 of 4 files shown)


masom:
Hi @Savar 

I think this is a great addition to the thumbor ecosystem.

We are trying to keep the thumbor installation rather small and use plugins to complement pluggable modules.

Here's two possibilities to promote your plugin:
- We can create a new repo at https://github.com/thumbor-community and give you admin access to it.
- You could host this plugin under your github username

Savar:
@masom I'm fine with adding it to the thumbor-community but there is a little bit a mixed feeling about that:
a) the CE is claiming to be in beta still (and it didn't change for more than 8 months)
b) it is not really part of the thumbor project but injects stuff from the outside and is duplicating functionality (context loading and stuff like that) which in my opinion is a weird setup if you manage the original source code and you could add this as a main functionality into thumbor
c) it is a rather small addition which actually might not be worth the effort to install the CE with the additional risk of bugs inside that stuff
d) community stuff is nice but I think there should be a way how "good stuff" (I'm not saying mine is good in any way!) can go into the main project if it is a contribution which seems to fit well for all thumbor users..

next to that I'm still thinking about the metrics itself because it is still not optimal to use the naming from statsd if you use thumbor so I was thinking to change that part in a more generic way so that metrics aren't having the final naming at occurrence time but rather inside the metric class and to support natively the label functionality.. but unfortunately it isn't as easy as I thought

but it is not my project so I'm fine to try out the CE

Savar:
additionally: the documentation is still only having the shortener inside and not the other modules :(

Savar:
and if there would be changes in the metric classes in general this would mean, that an outside module is responsible for an internal change ;) .. just saying

masom:
@Savar Changes are fine and we welcome them. For example, the metrics used to be statsd-only until someone needed to add another service. That part of the system was made "generic" and two/three adapters exists.

URL signing received the same treatment, it is now a configurable module instead of being hardcoded.


The CE general organization does not dictate any level of stability or that it must use `tc_core` (that module is highly experimental but works for many use cases, monkey-patching thumbor).

A great module is the AWS storage/loader combo: https://github.com/thumbor-community/aws

masom:
https://github.com/thumbor-community/thumbor-memcached
https://github.com/thumbor-community/mongodb

Were also two supported loaders/storages in core but got moved out to reduce the footprint we have to maintain in `thumbor/thumbor`

Savar:
@masom then please create the repository and I will use the `librato` one as a reference how to create it

masom:
@Savar https://github.com/thumbor-community/prometheus is now live.

If you would like to add/change how metrics can be collected please open an issue to discuss the approach.

There is also the possiblity of adding a new type of module to do better tracing than what statsd/librato supports.

Savar:
@masom do I have already admin rights? I cannot push to the repository.

masom:
@Savar I think you'll need to accept the invite to be an admin.

image

Savar:
ah sorry, only visible on the org level but not on the repo level.. missed that :( thanks!

Issue #726, opened 2016-05-18
Upscale and keeping ratio with the fill filter


COil:
### Thumbor request URL

[http://localhost:8888/unsafe/80x40/filters:fill(white,true)/http%3A//assets.kelformation.com/logos/organismes/3672/logo_centre_3672_1.gif](http://localhost:8888/unsafe/80x40/filters:fill%28white,true%29/http%3A//assets.kelformation.com/logos/organismes/3672/logo_centre_3672_1.gif)

Original image: [Original image](http://assets.kelformation.com/logos/organismes/3672/logo_centre_3672_1.gif)
### Expected behaviour

Hi, I am trying to get a thumbnail of an exact size of 80x40, that's why I use the fill filter. I use the fit-in parameter if the original image if bigger in width or length than 80x40. 

In this case the original image is 50x37, (it can be smaller or bigger than the requested width or length)

It should generate a 80x40 image but not cropped. In fact want I want is: [this](http://assets.kelformation.com/img/m/1/organismes/3672/logo_centre_3672_1.gif)
### Actual behaviour

Image is cropped and bottom and top parts of the original image are loosed, but its 80x40 as requested. if I use 0x40 the ratio is Ok but the image is 55x40 not 80x40.

[Wrong generated logo](https://thbr.figarocms.net/external/t5QLV0owiGWOjLx049eMFf3uzag=/80x40/filters:fill%28white,true%29/http%3A//assets.kelformation.com/logos/organismes/3672/logo_centre_3672_1.gif)
### Operating system
- Ubuntu 14.04
- Thumbor 6.0.1

hhankj2u:
Have you try smart crop option yet?

kkopachev:
Looks like it's a bug when `USE_GIFSICLE_ENGINE=True`. It scales png images as you'd expect.

COil:
Thank you, I'll give a try tuesday.

COil:
I try with both values but it doesn't work. When changing this setting the difference is the quality of the generated image.

kkopachev:
Sorry, forgot to mention that this works when you set `fit-in`. 
http://localhost:8000/unsafe/fit-in/80x40/filters:fill(white,true)/http%3A//assets.kelformation.com/logos/organismes/3672/logo_centre_3672_1.gif

and this is the output: 
[...]

PR #870, opened 2017-02-25
changing description of fit-in

Savar:
@masom can you check if this is correct what I'm writing there? Also two things:

1) is adaptive-fit-in and full-fit-in deprecated or why isn't it in the docs anymore? (if it should be in again, then I can say that I don't get the explanation in the old wiki.. or at least the result is different than the explanation, like there is never a `cropping` but this is said in the wiki)
2) maybe I missed it, but is there any way to specify a box WxH where images are either shrinked (like fit-in is doing) or upscaled to this box so that it keeps the ratio but will always use the maximum space in the box available?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -83,7 +83,7 @@ This is very useful when you need to fit an image somewhere, but you
 have no idea about the original image dimensions.
 
 Keep in mind that it won't enlarge your image in case it is smaller than
-the specified "E" withd and "F" height! One way of getting the image in the
+the specified "E" width and "F" height! One way of getting the image in the
 size requested is to use the :doc:`filling` filter.
 
 Image Size
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -64,8 +64,8 @@ Fit in
 ~~~~~~
 
 The fit-in argument specifies that the image should not be auto-cropped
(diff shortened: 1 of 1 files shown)


Savar:
also see #726 regarding part 2.

docs/usage.rst
```diff
@@ -82,6 +82,10 @@ how thumbor would resize it:
 This is very useful when you need to fit an image somewhere, but you
 have no idea about the original image dimensions.
 
+Keep in mind that it won't enlarge your image in case it is smaller than
+the specified "E" withd and "F" height! One way of getting the image in the
```
scorphus: Please change: `s/withd/width/`

docs/usage.rst
```diff
@@ -64,8 +64,8 @@ Fit in
 ~~~~~~
 
 The fit-in argument specifies that the image should not be auto-cropped
-and auto-resized to be **EXACTLY** the specified size, and should be fit
-in an imaginary box of "E" width and "F" height, instead.
+but auto-resized (shrinked) to fit in an imaginary box of "E" width and
+"F" height, instead.
```
scorphus: Makes much more sense IMHO. Thanks!

Savar:
updated :)

scorphus:
Thanks, @Savar!

PR #871, opened 2017-02-26
fixes #808 - requests hangs in engine.read exception

### LINKED ISSUE #808: Requests hang on exceptions in engine.read

Savar:
in case an exception is thrown inside the engine.read() method the
exception was either not catched properly (in case the THREADPOOL size
is >0) or the result was not a tuple and the unpacking waits for more
data while calling .result() (in case the THREADPOOL size is 0)

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -380,7 +380,7 @@ class BaseHandler(tornado.web.RequestHandler):
             try:
                 future_result = future.result()
             except Exception as e:
-                logger.exception(future_result)
+                logger.exception(e)
                 self._error(500)
                 return
 
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -378,16 +378,8 @@ class BaseHandler(tornado.web.RequestHandler):
 
         def inner(future):
             try:
-                # depending of ENGINE_THREADPOOLSIZE >0 or ==0 the
-                # future.result() is calling the corresponding
-                # operation right here, so an exception can occur here
-                # or it was already been called and the result() is simply
-                # returning the already cached result
                 future_result = future.result()
             except Exception as e:
-                future_result = e
-
-            if isinstance(future_result, Exception):
                 logger.exception(future_result)
                 self._error(500)
                 return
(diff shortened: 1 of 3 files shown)


thumbor/handlers/__init__.py
```diff
@@ -377,7 +377,22 @@ def finish_request(self, context, result_from_storage=None):
...
+                # or it was already been called and the result() is simply
+                # returning the already cached result
+                future_result = future.result()
+            except Exception as e:
+                future_result = e
+
+            if isinstance(future_result, Exception):
+                logger.exception(future_result)
+                self._error(500)
+                return
```
scorphus: Could you move this block to the `except` block? Also, why not keep the original code with just an added try/except?

scorphus:
Thanks for the contribution, @Savar. Can you please add tests for this?

Savar:
@scorphus I'll try to add tests.

Savar:
@scorphus I fixed the checks but to be honest the part I would like to test (the function passing the `def inner` method) I cannot mock. So can you please tell me what kind of tests you would like to have?

scorphus:
Thanks, @Savar!

> […] can you please tell me what kind of tests you would like to have?

Keep an eye on upcoming tests 😉

PR #877, opened 2017-03-07
fixing issue with full-fit-in and requested width/height

Savar:
if full-fit-in is being used the resized image dimensions are normally
bigger then the requested WxH but the `fill()` filter will use the WxH
to transform the image which is then wrong

# Test Image:
fill_test

# Request:
http://localhost:8081/unsafe/full-fit-in/1000x1000/filters:fill(auto)/https://cloud.githubusercontent.com/assets/246341/23646258/58a1eb14-0310-11e7-80ec-18a279f50077.jpg

## without the patch:
fill_test_without_patch

## with the patch:
fill_test_with_patch


scorphus:
Thanks, @Savar! Can you please add tests that assert it works as expected? You're likely willing to change `tests/filters/test_fill.py` in this case. 😉  ☕️ 

Although this proposes changing the way `fit-in` works for all existing use cases, I believe the current behavior is unexpected. Also, for instance, this will render `background_color` completely unnecessary – I see no problem in keeping it.

I'll give this a LGTM as soon as we have tests – of those that compare images. Let's hear from others, though.

Savar:
I'll add tests, but why does it render `background_color` as completely unnecessary?

Savar:
@scorphus anything you need for this? Is the test enough?

Savar:
@scorphus did I miss anything? I would rebase with master but first I want to know if you would be willing to merge it then or if I have to change anything which I would do while re-basing. Otherwise it makes no sense for me to work on it :disappointed:

Savar:
@scorphus @masom any update on this one?

thumbor/transformer.py
```diff
@@ -340,6 +340,12 @@ def fit_in_resize(self):
             resize_height = self.target_height
             resize_width = round(source_width * self.target_height / source_height)
 
+        # ensure that filter should work on the real image size and not on the request
+        # size which might be smaller than the resized image in case `full-fit-in` is
+        # being used
+        self.context.request.width = int(max(self.context.request.width, resize_width))
```
kkopachev: This is throwing exception in case when width or height specified as 'orig'
`/unsafe/fit-in/200xorig/http://www.wallsave.com/wallpapers/2560x1600/cloud/66266/cloud-zeom-computing-g-free-66266.jpg`
that would return unresized image

PR #883, opened 2017-03-11
fix exception with smart detection + gif + meta endpoint

Savar:
the JSONEngine was not having the `extension` attribute stored from the
wrapped Engine which leads to the exception fixed (ignored) in #841

with the added extension part the transformer won't call smart detection
if gifsicle is enabled and it's a GIF image


Savar:
somehow the JSONEngine isn't tested at all.. so I'm not feeling that well with this engine in general but also I have no idea how to start "test" this in a proper way.

Savar:
@masom anything you need to get this in?

PR #884, opened 2017-03-11
never hide unexpected exceptions without logging it

Savar:
in #841 the exception were caugth and ignored and only a small warning
was logged, this will not allow to debug the underlying issue (because
it was actually a fixable issue, see #883).. so now the exception is
logged as well


masom:
🚀

PR #886, opened 2017-03-12
resync BaseEngine with existing implementations

Savar:
BaseEngine was using methods of "engines" without even implementing them
by itself (with `raise NotImplementedException`) and others with a
complete different set of parameters than the real implementations were
using.

Also the `convert_to_grayscale` is different in the implementations so I
tried to sync them as well as good as I was able to do


Savar:
@masom how about this one... ?

masom:
I'll merge once it's green

PR #887, opened 2017-03-13
fixing _composite.apply with bigger source images

Savar:
_composite.apply is pasting an image A into an image B at a virtual
position in space but not expanding the size of image B. If the image A
is bigger than image B and should be placed at x < 0 and/or y < 0 then
the image isn't using all the available space in image B in case image A
is bigger.

This is fixed now. The added test was failing before applying the
bugfix.

Example URL (can be tested with the image in this pull request, `test/fixtures/filters/`): http://localhost:8081/unsafe/filters:watermark(http://172.17.0.1:8888/watermarkBig.png,-10,-100,50)/http://172.17.0.1:8888/source.jpg

before the patch:
bad

after the patch:
good


Savar:
any reason, why this is not going through?

masom:
@Savar Would you mind rebasing this PR? I'll merge it after.

Savar:
sure.. let me do this

Savar:
@masom rebased.. it should go green as soon as the tests are done.. please merge :smile:

PR #890, opened 2017-03-14
resizing the watermark relative to image (fixes #874)

### LINKED ISSUE #874: Watermark Not Scaling Down
### Thumbor request URL
http://localhost/unsafe/400x400/filters:watermark(http://localhost/very_large_watermark.png,-1,-1,50)/http://localhost/photo.jpg

### Expected behaviour
When using the watermark filter, I would create a large watermark image that can be used for large and small images. The watermark image should scale down to fit the size of the output image.

### Actual behaviour
When using the watermark filter along with any resizing of an image, the watermark does not scale to the output image. If I have a small image at 100x100 and my watermark is 1000x10, the watermark is too big for the image. If I have large image at 1000x1000 and my watermark is small, then it is too small to see on the big image.

### Operating system
Ubuntu

### More detail

The watermark image I am using has a logo to the left and copyright information to the right. It is 1000x30 (300 dpi) in size. Images on our site very from 4096x4096 to 100x100. By using a large watermark, it ensures that it can be seen on any of these images. But it has to scale down to fit the image output for this to work.

I can resize the image on the fly using thumbor for the watermark itself but this ends up putting a huge load on the server because it has to recreate the watermark for every image. If I'm displaying 20 or more images, the server starts to lag and crash. So this is not a very efficient work-around.

It should be built into thumbor that the watermark auto scales with the image.

Am I doing something wrong?

Savar:
two new (optional) parameters added to the watermark filter to resize
the watermark relatively to the image it should be placed on

see the provided doc file to understand how it works


Savar:
should fix #874

r0bdiabl0:
This looks great.

Savar:
anything which needs to be done to get this into thumbor?

Savar:
@masom ? :smile:

Savar:
@masom how about that one?

LorisLunedei:
That's great! When will it be released?

PR #892, opened 2017-03-15
using the content detection to determine  extension

Savar:
The filters were relying on the image url to contain the extension.
If no extension is given they passed an empty string to the engine which
somehow worked fine within the PIL engine but using OpenCVs engine the
extension is being used and gives back an `KeyError`.

In general the extension should be determined on the content and not
by name to not handle accidentally PNGs as GIFs or the other way around.


Savar:
I'm anyhow unsure why the `engine.load()` allows to specify the extension.. is there a good reason for that?

PR #893, opened 2017-03-15
watermark() relative positioning

Savar:
now instead of giving the exact offset from top/left/bottom/right or
center or repeat, it is now also possible to say something like:

10percent offset relative to the image size (see docs for more details)


r0bdiabl0:
This is great as well.

Savar:
is there any chance to get this pending pull request merged into? It is (in my opinion) well tested, well explained in the docs and not changing any actual behavior if not the parameters have a `p` after the number which was not possible before (error) so it shouldn't be in use.

masom:
Thanks @Savar Sorry about the delay in merging this.

PR #894, opened 2017-03-16
Refactory storages to use coroutine

diegosperes:
;)

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -638,7 +638,7 @@ class BaseHandler(tornado.web.RequestHandler):
     def get_blacklist_contents(self):
         filename = 'blacklist.txt'
 
-        exists = yield gen.maybe_future(self.context.modules.storage.exists(filename))
+        exists = yield self.context.modules.storage.exists(filename)
         if exists:
             blacklist = yield gen.maybe_future(self.context.modules.storage.get(filename))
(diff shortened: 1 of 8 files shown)


thumbor/handlers/__init__.py
```diff
@@ -638,8 +638,8 @@ def _fetch(self, url):
     def get_blacklist_contents(self):
         filename = 'blacklist.txt'
 
-        exists = yield gen.maybe_future(self.context.modules.storage.exists(filename))
-        if exists:
+        exist = yield self.context.modules.storage.exists(filename)
+        if exist:
```
scorphus: Try not to rename variables and leave only changes related to the coroutine rewrite. Also, I believe `exists` to be a better suited name for that variable.

thumbor/handlers/image_resource.py
```diff
@@ -25,9 +25,9 @@ class ImageResourceHandler(ImageApiHandler):
     def check_resource(self, id):
         id = id[:self.context.config.MAX_ID_LENGTH]
         # Check if image exists
-        exists = yield gen.maybe_future(self.context.modules.storage.exists(id))
 
-        if exists:
+        exist = yield self.context.modules.storage.exists(id)
+        if exist:
```
scorphus: Same here and across all the way down...

heynemann:
Closing this in favor of other PR: #1000

PR #898, opened 2017-03-22
Improvement to gifsicle error handling

okor:
Sometimes gifsicle will produce a `1` return code when processing some gif. We handle this scenario by assuming the output gif is invalid. This is often not the case. Gifsicle is fairly strict and will complain a lot but **most of the time** ... it produces valid output even though it exits with a `1`.

So I've modified our error handling so that:
- If gifsicle produces a `1` return code, log the error. This is also MUCH better for debugging when you are crawling through logs.
- Completely separately we check if gifsicle actually produced any output. If the buffer is empty, then and only then we actually raise an error in our response.

The end result is that I should get a lot less pings in Slack for "why is this image not showing up?". And same for you ; )

Now we can see helpful debugging information in the error logs:
```
  2017-03-22 15:17:34 thumbor:ERROR gifsicle:<stdin>:#0: warning: 10 superfluous pixels of image data
  gifsicle:<stdin>:#1: warning: 3 superfluous pixels of image data
  gifsicle:<stdin>:#2: warning: 9 superfluous pixels of image data
  gifsicle:<stdin>:#3: warning: 7 superfluous pixels of image data
  gifsicle:<stdin>:#4: warning: 2 superfluous pixels of image data
  gifsicle:<stdin>:#5: warning: 7 superfluous pixels of image data
  gifsicle:<stdin>:#6: warning: 3 superfluous pixels of image data
  gifsicle:<stdin>:#7: warning: 10 superfluous pixels of image data
  gifsicle:<stdin>:#8: warning: 2 superfluous pixels of image data
  gifsicle:<stdin>:#9: warning: 8 superfluous pixels of image data
  gifsicle: (plus more errors; is this GIF corrupt?)
  gifsicle: warning: too many colors, using local colormaps
    (You may want to try ‘--colors 256’.)
  gifsicle: background color not in colormap
```


masom:
LGTM

okor:
I have a meeting, will be back to fix that test failure (feel free to log failure notes if you got em').

okor:
So ... I fixed the test. We basically want the opposite behavior as before ... as long as there is output from gifsicle, the test/request should pass. This is good. Soooo many headaches will disappear for people doing gif things with Thumbor. 

The other condition ... where gifsicle is not able to produce output ... I literally can not fuck up a gif bad enough that gifsicle won't produce output. It's maddeningly amazing how durable the library is. I was just smashing keys in a hex editor, saving, running though gifsicle ... just can't get that library down. If someone can create a gif that gifsicle can NOT produce output from AND the license on said gif is permissive, then I will happily add a test for that scenario.

okor:
Ok. Last note. With genuinely corrupt gif images, even if you are using gifsicle thumbor will throw an error when Pillow tries to identify the image higher up in the code path so it will never even make it to gifsicle. Pillow seems to be more strict about encoding, possibly less strict less important stuff like color count, etc. So yeah, needing throwing an error manually is probably going to be extremely rare since gifsicle seems to be able to produce valid gifs from random binary data : )

PR #899, opened 2017-03-23
Add MP4 engine to support producing webp/gif directly from mp4

winguse:
In the PR, I implemented a new engine for producing `webp` / `gif` from `mp4`. In my case, I want to serve my users animated images from videos. (`webp` will be returned only when `AUTO_WEBP=True`)

I have an original idea that transcode video into gifs, but the gifs are too large and bad in quality. `webp` is very good, but some devices do not support it.

I also considered saving the `webp` as the source file to be converted to `gif` when needed, but [`ffmpeg` do not support decoding the animated `webp` currently](https://trac.ffmpeg.org/ticket/4907). So in this PR, I am using a pretty tricky hack to save the source file: return the original `mp4` buffer when `read(extension='.mp4', quality=None)`.

I started learning Python only a few days ago and had never wrote so much Python, please help to correct me if I made any stupid mistakes.

Thank you!


andreaugusto:
Heya @winguse! Thanks for this PR!

First of all, gratz for you first Python code :) Way to go!

We've been talking about a better way of adding new Engines to Thumbor, but the actual code has some issues that needs to be addressed first. Right now, through configuration, we are only able to change an Engine for all images or change an Engine for gifs, separately. The only way to add a brand new Engine for another mimetype is, as this PR did, adding another *if* to the [`_fetch`](https://github.com/thumbor/thumbor/blob/master/thumbor/handlers/__init__.py#L536) method. IMO we should fix it and add the possibility to select an Engine based on the resource's mimetype. We'd have to change [this](https://github.com/thumbor/thumbor/blob/master/thumbor/handlers/__init__.py#L601) so that it is not needed anymore to add more *if*s to the code. We'd also have to change Thumbor's configuration to support different Engines based on a resource's mimetype.

e.g.:
```python
Config.define(
    'ENGINES', [
        ('video/mp4', 'thumbor.engines.mp4'),
        ('image/gif', 'thumbor.engines.gif'),
        ('default', 'thumbor.engines.pil'),
    ]
```

With that being said, I'd like to propose that:

1- This PR would be only about adding a new mp4 Engine.
2- Change some things in the actual implementation of the PR'ed Engine (long methods, code legibility and stuff). We'd help you with the actual changes. 
3- Open a new issue to add support to new Engines based on configuration.
4- Wait for 3 to be finished so that this Engine could use the correct way to add a new Engine to Thumbor.

I'll take care of 3 and will link it in here ASAP.

Then again, thank you for the PR and keep up the good work!

winguse:
Thanks @andreaugusto , good to know we can avoid *ifs*, that should be a good idea. 👍

fdintino:
Using this PR as a starting point, I created a thumbor engine that can handle reading and writing videos ([thumbor-video-engine](https://github.com/theatlantic/thumbor-video-engine)). It supports mp4 (h264 or hevc), webm, gif, and animated webp. It's currently light on documentation, and it does not yet support features like the `max_bytes` filter, but it is functional and [highly configurable](https://github.com/theatlantic/thumbor-video-engine/blob/0a3352fb487c77c9437216057719fb2654b9c2be/src/thumbor_video_engine/__init__.py)

heynemann:
We're in the process of reducing thumbor's codebase and we'll only keep one example implementation of each concept (loader, storage, engine, etc) with some exceptions. I highly encourage you to push this to a project that people can just install alongside thumbor and plug it in.

PR #901, opened 2017-04-03
Fix race condition for preserving exif data in de1d96f614f5798a3d6f3e

mb-m:
In de1d96f614f5798a3d6f3e20e36a9ec061f8e960, because of the API to engines, the
only way to preserve the EXIF data for storage is to change the global (effective
singleton) configuration. Unfortunately, if another request is doing an Engine.read
at the same time with the same configuration, it suddenly has its configuration
forced by this block. We change this instead to have quasi copy-on-write semantics.

There are no new tests as this is a timing-based race and easy to show, but difficult to write automated tests for.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -10,6 +10,7 @@
 
 import sys
 import functools
+import copy
 import datetime
 import re
 import pytz
@@ -592,7 +593,6 @@ class BaseHandler(tornado.web.RequestHandler):
         self.context.request.extension = extension = EXTENSION.get(mime, '.jpg')
 
         original_preserve = self.context.config.PRESERVE_EXIF_INFO
-        self.context.config.PRESERVE_EXIF_INFO = True
 
         try:
             if mime == 'image/gif' and self.context.config.USE_GIFSICLE_ENGINE:
@@ -620,6 +620,20 @@ class BaseHandler(tornado.web.RequestHandler):
             is_mixed_no_file_storage = is_mixed_storage and isinstance(storage.file_storage, NoStorage)
 
             if not (is_no_storage or is_mixed_no_file_storage):
+                # in the case that we have local storage, it's important that we don't try
+                # and strip the exif data - so that we can make the choice about stripping
+                # it on output, instead of input.
+                # unfortunately, the only interface to the engine is the global config
+                # referenced by our context, which has the current setting of preserving
+                # the exif info or not. this is also being read by concurrent
+                # coroutine-threads, so just setting this value outright causes a race.
+                # we solve this in order not to influence these other coroutine-threads
+                # by doing an effective copy-on-write for the config until the end of this
+                # context (ie. this request).
+                # at some point in the future - the API should change, and this horrid hack
+                # should go away.
+                self.context.config = copy.deepcopy(self.context.config)
+                self.context.config.PRESERVE_EXIF_INFO = True
                 fetch_result.buffer = self.context.request.engine.read(extension)
                 storage.put(url, fetch_result.buffer)
 


mb-m:
To test this, we ran a two tight loops with a PRESERVE_EXIF_INFO=False configuration running until exif returned 0 - this is reproducible within a second or two of starting both loops. As I say, however, its basis as a timing-based race condition makes the test difficult (if not impossible) to automate. What I can say is that the behaviour appears not to have changed because of the existing tests - apart from not having the race anymore.

Savar:
I'm not sure if everyone is aware of the fact, that this bug is a 'privacy' issue which might lead to legal issues in some countries. Think of customers uploading images and the system serves the image but doesn't strip the EXIF data out which might contain personal data.

masom:
> Think of customers uploading images and the system serves the image but doesn't strip the EXIF data out which might contain personal data.

EXIF data should be striped during upload if the system that is being dealt with represents a privacy issue.

Thumbor's behaviour of a race condition on exif data is a bug, exposing exif is not.

thumbor/handlers/__init__.py
```diff
@@ -620,6 +620,8 @@ def _fetch(self, url):
             is_mixed_no_file_storage = is_mixed_storage and isinstance(storage.file_storage, NoStorage)
 
             if not (is_no_storage or is_mixed_no_file_storage):
+                self.context.config = copy.deepcopy(self.context.config)
+                self.context.config.PRESERVE_EXIF_INFO = True
```
masom: It would probably be better to refactor code to not use a global variable.

Savar:
@masom this is nothing you can say generally. In our case we *want* to keep the original image completely unchanged because it can be accessed via an authenticated way and should be the original one without any change. At the same time we have a public endpoint to access these images in a special way which will have to strip the EXIF data. So we rely on thumbor here. And at least in my opinion it is a valid requirement.

thumbor/handlers/__init__.py
Savar: probably, but this would be a bigger change and I'm fine if someone is having the overview and/or time to do this. In the meanwhile I think it is absolutely valid to fix this bug in such a way (we actually thought about passing the 'strip exif' attribute to the engine but this would change a lot)

mb-m: @masom: nice idea, but the big problem with that is that that requires changing your API for plugins as they stand. Since these aren't even all in one repository, that, as @Savar says is a major piece of work, and one that doesn't make sense for something which, as you rightly point out, is a bug, introduced 4 years ago. Since "done is better than perfect", it makes sense to merge this change now, and think more carefully about long-term APIs for the transform plugins.

Savar:
@masom would it be okay to get this somehow into thumbor and then better create a new `issue` which is  of type `refactor`? The bug is fixed with this 2 lines without any real risk to introduce a new bug but changing the whole api would be probably something which needs much more love and might touch much more topics than only this one?!

mb-m:
This PR depresses me. A 4-line patch fixing an actual verifiable bug remains unmerged a month after submission, and the bug is still unfixed in upstream. What happens if I need to submit a 100-line one, or a 1000-line one?

Is there anything we can do to actually get this merged? Should I bother submitting future patches to thumbor?

thumbor/handlers/__init__.py
```diff
@@ -620,6 +620,8 @@ def _fetch(self, url):
             is_mixed_no_file_storage = is_mixed_storage and isinstance(storage.file_storage, NoStorage)
 
             if not (is_no_storage or is_mixed_no_file_storage):
+                self.context.config = copy.deepcopy(self.context.config)
```
andreaugusto: @mb-m The only thing that I'd change here is add a comment on to why we are copying a config locally. It would make clear to anyone that reads it later on.

andreaugusto:
Heya, @mb-m! First of all, thanks for this PR. Sorry for the time taken, but we ran into a stalemate with this PR because we were not reaching a conclusion as to how we should fix it. There's this PR that is a quick fix to the issue and there's the necessity to add to the interface parametrization that would override a given configuration race-condition-free. The latter is a breaking change and has to be done in a major release.

I'm proposing that we accept this PR and that it stays as is until we release a 7.0 with some other changes that we might want to add.

There is also [this comment that I made](https://github.com/thumbor/thumbor/pull/901#discussion_r115359639).

mb-m:
@andreaugusto perfect result, that's exactly what I was thinking (do this now, fix properly later). I don't honestly care about the fix, just as long as there is one ;-)

I can add the comment.

andreaugusto:
Thanks for the comment, @mb-m!

This PR LGTM. Anyone wants to add anything more?

As of the real fix, I've opened a [7.0 milestone](https://github.com/thumbor/thumbor/milestone/13) and an [discussion issue](https://github.com/thumbor/thumbor/issues/920) as well.

scorphus:
LGTM as well

Thanks, @mb-m!

p.s.: :+1: for the amend

mb-m:
No worries - I think you're right that adding the comment was the right thing to do - too often on internal work I rely on the commit history for the commenting, but in a public OS project, commenting in code is the better strategy! Now let's get the turnaround for these down from a couple of months ;-) Keep on with the good work, guys, and thanks!

Issue #903, opened 2017-04-07
Delay in last GIF frame is incorrect when using gifv


okor:
### Thumbor request URL
https://cdn0.vox-cdn.com/thumbor/BbnJ1AHcTSqqFdPKSVuyJgf_yr8=/0x0:2432x1556/1720x0/filters:focal(0x0:2432x1556):gifv():no_upscale()/cdn0.vox-cdn.com/uploads/chorus_asset/file/8251197/7GPW_LivingRoom.gif

### Expected behaviour
The resulting GIFv video should be 4 seconds in duration.

### Actual behaviour
The video is 2 seconds in duration.

### Operating system
OSx & Ubuntu

### Your thumbor.conf
```
...
OPTIMIZERS = [
    # 'thumbor.optimizers.jpegtran',
    'thumbor.optimizers.gifv',
]
...
```

### Source of the bug
I did some debugging and have discovered that this issue is being caused by the `scale` filter used in the ffmpeg command. The scale filter we are using ensures that the video has an even number of pixels for both width and height - because odd pixel numbers in mp4 at least is not valid and ffmpeg will puke.

I have submitted a bug with ffmpeg https://trac.ffmpeg.org/ticket/6294#ticket

However, in the mean time (which could be an hour or years in ffmpeg world) we have some options on our end as well. Mainly, we could drop the scale filter from the command and instead enforce even number of pixels ourselves before we perform the transformation to mp4/webm format. I have confirmed that removing the filter option fixes this problem.

So, I'd like to get some opinions on what we should do.

Savar:
This is kind of tricky, because it somehow will create a bad situation as soon as the bug is fixed.

Is there any chance to first scale with the gif engine the gif into an even width/height and then push it to ffmpeg?

Savar:
something like this could handle the issue as long as ffmpeg cannot scale correctly, but this is really not good tested.. I'm not sure if it is a problem to reload the image into the `context.request.engine` and also I'm not sure if this would crash if gifsicle is enabled and the opencv-engine is being used, because if `context.request.engine` is opencv-engine it cannot handle the image.

```diff
diff --git a/thumbor/optimizers/gifv.py b/thumbor/optimizers/gifv.py
index 21525a7..98e97f4 100644
--- a/thumbor/optimizers/gifv.py
+++ b/thumbor/optimizers/gifv.py
@@ -22,6 +22,29 @@ class Optimizer(BaseOptimizer):
[...]

PR #904, opened 2017-04-10
Remove scale option from ffmpeg - bugfix

okor:
This fixes a bug for mp4 videos (see: https://github.com/thumbor/thumbor/issues/903 and https://trac.ffmpeg.org/ticket/6294).

This PR removed the scale option from ffmpeg when converting a gif to a video. Instead, the request dimensions are munged in the transformer.py before the image is processed. It works with Pillow or Gifsicle engines.

Unfortunately, it does not fix the same issue which is present for webm videos. Removing the scale option has not affect and the last frame is still cut off. Similarly if all options are removed, making the simplest possible case `ffmpeg -i input.gif output.webm` the output still messes up the last frame duration. I can't find a solution but have submitted a bug to ffmpeg for that issue as well: https://trac.ffmpeg.org/ticket/6302 My suggestion would be to avoid using webm for now, as I have done.

If and when both bugs are fixed on the ffmpeg side, we can and should revert this change and push the responsibility of enforcing even dimensions back to ffmpeg since this restraint only applies to videos.

### DIFF
--- a/tests/handlers/test_base_handler.py
+++ b/tests/handlers/test_base_handler.py
@@ -665,6 +665,14 @@ class ImageOperationsWithGifVTestCase(BaseImagingTestCase):
         expect(response.code).to_equal(200)
         expect(response.headers['Content-Type']).to_equal('video/webm')
 
+    def test_should_convert_animated_gif_to_video_and_force_even_dimensions(self):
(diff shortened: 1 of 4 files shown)


okor:
Tests are good. No opposition. Mergin'.

PR #905, opened 2017-04-10
Add manual garbage collection to free memory

okor:
More recent versions of thumbor have had a pretty serious problem with memory leaks of some kind. See https://github.com/thumbor/thumbor/issues/753 for discussion. I spend a non-trivial amount of time troubleshooting/investigating and testing work arounds. 

My conclusion is that we are **not** being affect by any memory issues which might exist in Pillow - that issue only affects Python 3+. Thumbor is not compatible with Python 3. My tests confirmed that.

I also dug into our request and object allocation cycles, looking for leaked objects ... but I could not find any. I did quite a bit of looking! But each time I looked, performing many thousands of requests to a local thumbor, not a single request resulted in a true memory leak. If an object was created it was able to be GC'ed and thus freed memory. This analysis does not align completely with production as we still see a leak over time but this patch which we have been running in production for about a month **DOES** dramatically increase the amount of time it takes to run the server out of memory.

Without a better solution I strongly propose this patch. The patch was initially composed by @mvdklip and he should get all the credit if it goes well. I am more than happy to let someone smarter / has a better solve but until then I think this is the best solution and should be in core thumbor.

As for concerns about CPU usage we have only seen a very small increase. Mileage will vary I'm sure. Even with some non-trivial increase in CPU for some users I still think that's better than crashing servers and tons of failed requests.

### DIFF
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -131,6 +131,7 @@ Config.define('ALLOW_UNSAFE_URL', True, 'Indicates if the /unsafe URL should be 
 Config.define('ALLOW_OLD_URLS', True, 'Indicates if encrypted (old style) URLs should be allowed', 'Security')
 Config.define('ENABLE_ETAGS', True, 'Enables automatically generated etags', 'HTTP')
 Config.define('MAX_ID_LENGTH', 32, 'Set maximum id length for images when stored', 'Storage')
+Config.define('GC_INTERVAL', None, 'Set garbage collection interval in seconds', 'Performance')
 
 # METRICS OPTIONS
 Config.define('STATSD_HOST', None, 'Host to send statsd instrumentation to', 'Metrics')
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -617,6 +617,13 @@ Default filename for image uploaded.
 
 i.e.: ``UPLOAD_DEFAULT_FILENAME = 'image'``
 
+GC\_INTERVAL
+~~~~~~~~~~~~
+
+Set manual garbage collection interval in seconds. Defaults to None (no manual garbage collection). Try this if your Thumbor is running out of memory. May cause an increase in CPU load.
+
+i.e.: ``GC_INTERVAL=60``
+
 Example of Configuration File
 -----------------------------
 
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,7 @@ http://<thumbor-server>/300x200/smart/s.glbimg.com/et/bb/f/original/2011/03/24/V
             "futures",
             "argparse",
             "pytz",
+            "schedule",
         ],
 
         extras_require={
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -14,6 +14,7 @@ import datetime
 import re
 import pytz
 import traceback
+import schedule
 
 import tornado.web
 import tornado.gen as gen
@@ -394,6 +395,8 @@ class BaseHandler(tornado.web.RequestHandler):
 
             if should_store:
                 self._store_results(context, results)
+
+            schedule.run_pending()
 
         self.context.thread_pool.queue(
             operation=functools.partial(self._load_results, context),
--- a/thumbor/server.py
+++ b/thumbor/server.py
@@ -8,9 +8,11 @@
 # http://www.opensource.org/licenses/mit-license
 # Copyright (c) 2011 globo.com thumbor@googlegroups.com
 
+import gc
 import sys
 import logging
 import logging.config
+import schedule
 
 import os
 import socket
@@ -112,6 +114,12 @@ def run_server(application, context):
     server.start(1)
 
 
+def gc_collect():
+    collected = gc.collect()
+    if collected > 0:
+        logging.warn('Garbage collector: collected %d objects.' % collected)
+
+
 def main(arguments=None):
     '''Runs thumbor server with the specified arguments.'''
     if arguments is None:
@@ -128,6 +136,9 @@ def main(arguments=None):
     with get_context(server_parameters, config, importer) as context:
         application = get_application(context)
         run_server(application, context)
+
+        if (config.GC_INTERVAL and config.GC_INTERVAL > 0):
+            schedule.every(config.GC_INTERVAL).seconds.do(gc_collect)
 
         try:
             logging.debug('thumbor running at %s:%d' % (context.server.ip, context.server.port))


scorphus:
I'll review asap

thumbor/config.py
```diff
@@ -131,6 +131,7 @@
 Config.define('ALLOW_OLD_URLS', True, 'Indicates if encrypted (old style) URLs should be allowed', 'Security')
 Config.define('ENABLE_ETAGS', True, 'Enables automatically generated etags', 'HTTP')
 Config.define('MAX_ID_LENGTH', 32, 'Set maximum id length for images when stored', 'Storage')
+Config.define('GC_INTERVAL', 60, 'Set garbage collection interval in seconds', 'Performance')
```
scorphus: Default value needs to be `None`. This way we do not introduce any breaking changes.

thumbor/__init__.py
```diff
@@ -10,5 +10,5 @@
 
 '''This is the main module in thumbor'''
 
-__version__ = "6.3.2"
+__version__ = "6.4.0"
```
scorphus: Normally, versions are bumped in a separate, bump-only, tagged commit.

scorphus:
@andreaugusto, would you like to chime in and give us your thoughts?

thumbor/__init__.py
scorphus: I mean, they usually do not belong to a PR.

thumbor/config.py
okor: Is it a breaking change? It doesn't change any image functionality and should reduce the probability of a memory leak taking their server down.

I think this issue is more likely to affect people than not. Meaning that it should probably be on by default. Otherwise users are left to first experience the bug, then google for it, find a github issue/PR and try to adjust thumbor config.

thumbor/__init__.py
okor: Ok. So you'd prefer to have the version bumped on master as a separate commit?

scorphus: It's not a matter of preference. It's just that doing so would give us a change to include into 6.4.0 any other changes that may be ready to to go, not just this one.

okor: Roger. Reverted.

thumbor/config.py
okor: I went ahead and made the change. Personally I think it should default to on but if you think it's best to default to off then that's what we'll do.

scorphus: Sorry for not being clear enough. I see it as a _potential_ breaking change in a way that it may lead to a CPU usage rise in some scenarios we don't know about. We can't guarantee it won't. As we all know, forcing Python to do what it eventually is going to (collect collectable items) is not the real solution to the problem.

With that said, I believe it's okay supporting scheduled GC for those who can use and benefit from it, as long as it's an **_opt-in_**.

Having it on by default would lead us (Globo.com), for example, to either disable it or run massive tests to assert it doesn't affect our setup – remember we suffer from CPU usage and response time increase rather than OOM.

okor: Interesting. I'm totally fine with this then. Thanks for explaining!

okor: I already updated the default here https://github.com/thumbor/thumbor/pull/905/files#diff-cbd5a27df49cf19d11956cefe1692c6bR134 but for some reason it won't release your "review". I'm going to dismiss it but of course I made the change and appreciate the review : )

okor:
@scorphus Sure thing. Is it sufficient to do a "rebase and merge" in github?

Also, do we want to wait for more feedback before merging or are we good with this?

scorphus:
_Squash and merge_ is your friend 😉 

I'd really like to read @andreaugusto's thoughts on this, he has some good points. Specially regarding docs.

okor:
Cool, we can wait then.

andreaugusto:
Hi all, sorry for the long delay...

It seems like this discussion is already old (see #753) and whatever I say here has already been discussed. I'll not repeat the same things, but I do agree with @damaestro when [he says](https://github.com/thumbor/thumbor/issues/753#issuecomment-246533992) that this is a hack and I add, for now, it should be treated as such.

It is a bit strange that this GC calling every so often is able to hold the leakage because GC is enabled by default on the language. Maybe, as @okor said on #753, it is related to a finer tuning of GC to a 6.x Thumbor. If this is the case, this should definitely be enable by default on Thumbor core, but, since no one is really sure about it, I think we should dig a bit deeper and try to find a possible real issue to fix.

With that being said, I would recommend that we add some documentation on this conf saying that this is a temporary fix and it should not be relied on on later versions. If later we as a community reach to the conclusion that it really is a necessary tuning of GC, then we would remove this disclaimer, leave a documentation about it and move on.

As it is on this patch now, I would also not set it as default because, even though this bug impacted some people, some other people might not be having the same problem and we could be adding some CPU overhead to them unnecessarily.

As of me, I hope to be more engaged with the community from now on. I'll pair up with @scorphus and profile Thumbor looking for a possible leakage.

okor:
After running this patch in production for over a month, the leak is still present it's just slowed down dramatically. And I totally agree, it's a hack. But recycling servers or restarting Thumbor at arbitrary intervals doesn't seem a lot better.

When I did my testing to look for memory leaks locally I believe I was using all valid requests which would succeed. So you may want to dig into what happens when requests fail. For instance, it might be possible that when a request fails it holds onto a reference and for some reason isn't able to be GC'ed. If you restart Thumbor memory usage returns to flat, so it's definitely something within the Thumbor context and not a subprocess/shell command etc.

Another thing I would look at is how requests are queued up. Does a queued request initialize a Image objects? Is something some process getting zombied?

Last note. I noticed that if I made a small number of requests where Thumbor did not need to queue, memory stayed flat. I needed to get Thumbor to start queueing requests before I started to see memory climb. But if I stopped the test and kept thumbor running, once the queue was empty memory consumption returned to flat. So GC was working and releasing memory as expected.

In production however, where the requests never stop coming in, over time the memory leak will be visible and eventually lead to a OOM error on the host. And as soon as Thumbor starts swapping most request will fail due to timeout.

With or without the hacky patch, if I take one Thumbor out of the load balancer without restarting Thumbor the memory should drop to flat again ... if GC is working and there isn't a real leak. I'll test that now.

masom:
Could it just be files not being closed? We had a few `something = open('...')` that weren't closed.

okor:
Seems possible/probable.

andreaugusto:
We should probably move this conversation about the leakage to issue #753 since it is where everybody is. I'll add some more to this discussion there.

About this PR, could you add some documentation about this conf?

After that, if @scorphus is ok with it, I think we could merge it.

okor:
@andreaugusto This sufficient? https://github.com/voxmedia/thumbor/blob/7d9c78c0527ffd661eb563dd5cdb808a65cedd17/docs/configuration.rst#gc_interval

damaestro:
I like the suggestion to default to `None` and adding the disclosure. Good improvement.

I had also suspected file leaks in some shape or form. I didn't pursue that too far as it didn't make complete sense that a forced `gc.collect()` would somehow find something the default GC would not.

Is it possible that defunct or dysfunctional threads from tornado are holding open the objects? Maybe the normal CG still sees a reference and we are actually forcing the collection of those threads... subsequently freeing the referenced objects? I'm not as familiar with the tornado method of threads and async but I recall seeing this when I had a bug in how I was using a twisted reactor.

andreaugusto:
LGTM, @okor :)

@scorphus wanna merge it now?

scorphus:
> LGTM, @okor :)

Thanks for the contribution, @okor! Also thanks for squashing the commits, you make it all too easy :+1: 

> @scorphus wanna merge it now?

Let's do it!

okor:
🎉

PR #878, opened 2017-03-07
adding new filter `upscale()` (fixes #726 and #856)

Savar:
upscale() works together with (adaptive-)fit-in like no_upscale()
together with either /WxH/ or (adaptive-)fit-in + filters:fill()

PR #907, opened 2017-04-16
Add background_color option to gifv 

okor:
This PR:
- Adds a `background_color` option which only works with gifv
- Moves the "scale to even pixel dimensions" work back to ffmpeg
- Preserves proper gif delay durations

The reason we need the background color option is that some gifs contain transparency. These transparent regions will be converted to `white` by default. But white isn't always the correct color for a websites design. ffmpeg has no simple option to change the background as far as I could tell so we use this complex filter. Since ffmpeg will fill transparency with white every time and this new command is just as fast as the old one (benchmarked locally) we just use the same command every time even if no background option is specified - we just use white like default ffmpeg would. If a background_color is specified we use that color. Web applications will need to keep track and use the appropriate color for their design. ffmpeg is a bit pickier about colors - it doesn't allow hex short code for instance. So I added `webcolors` and a clumsy but robust logic for extracting colors while maintaining color format compatibility with the color options used in the `background_color` and `fill` filters that work with pngs. **There may be a better way to do this, let me know if that's the case.**

We also reintroduce the scale option to ffmpeg as well with the complex filter. Both the scale and canvas (background color) filters are performed at the same time to avoid extra passes / processing time. This simplifies maintenance of thumbor because we only have a single command.

* This also works with transparent gifs with frames with transparency that are smaller than the full size of the gif. Using tools like imagemagic/convert handle that scenario poorly.
* This works if a gif only has only two frames. The duration specified in the gif version will be preserved in the mp4/webm version correctly. Many many other methods of converting gifs to videos do not preserve the corrected durations - including fixed frame rate methods and/or other ffmpeg scaling methods all of which will break the duration of the last frame of the gif making it shorter than it should be (a single tick).
* This does **not** add a `background_color` option to normal gifs. If present the option will simply be ignored.
* I used `background_color` instead of `fill` because the latter implies that all transparency should be filled but as mentioned this will mess up a gif with overlaying frame where a later frame doesn't take up 100% of the size of the image canvas or should appear to float over previous frames. This is a true background color option.

This PR has been tested against ffmpeg version `3.2.4` **There are known issues with using these filters with later releases such as 3.3** but earlier version may work as well. See open ffmpeg issue: https://trac.ffmpeg.org/ticket/6294 Since the gifv filter is still categorized as experimental I think it's ok to exclude some versions of ffmpeg. Hopefully ffmpeg will fix their bugs in the next release.

**Example requests with identical output:**
`http://localhost:8888/unsafe/500x0/filters:gifv():background_color(ff00ff)/http://localhost/top_pin_wide.gif`
`http://localhost:8888/unsafe/500x0/filters:gifv():background_color(f0f)/http://localhost/top_pin_wide.gif`
`http://localhost:8888/unsafe/500x0/filters:gifv():background_color(magenta)/http://localhost/top_pin_wide.gif`
`http://localhost:8888/unsafe/500x0/filters:gifv():background_color(Magenta)/http://localhost/top_pin_wide.gif`
`http://localhost:8888/unsafe/500x0/filters:gifv():background_color(MaGeNTa)/http://localhost/top_pin_wide.gif`

**Example input with transparency:**
top_pin_wide-1

**Example output (mp4 converted back to gif manually, no mp4s supported in GH markdown)**
top_pin_wide_ff00ff

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,20 +40,24 @@ addons:
     - libopenblas-dev
     - liblapack-dev
     - python-pyexiv2
+    - yasm
 install:
   - pip install -U pip
-  - wget http://johnvansickle.com/ffmpeg/releases/ffmpeg-release-64bit-static.tar.xz
-    -O /tmp/ffmpeg-release.tar.xz
-  - mkdir /tmp/ffmpeg-release
-  - tar -C /tmp/ffmpeg-release --strip 1 -xvf /tmp/ffmpeg-release.tar.xz
-  - export PATH=/tmp/ffmpeg-release:$PATH
-  - ffmpeg -version
   - pip install --upgrade pip
-  - make setup
+  - cd $TRAVIS_BUILD_DIR && make setup
   - pip install coveralls
   - pip install cairosvg
+  - wget https://ffmpeg.org/releases/ffmpeg-3.2.4.tar.xz
+    -O /tmp/ffmpeg-3.2.4.tar.xz
+  - mkdir /tmp/ffmpeg-source
+  - tar -C /tmp/ffmpeg-source --strip 1 -xvf /tmp/ffmpeg-3.2.4.tar.xz
+  - cd /tmp/ffmpeg-source && ./configure --prefix=/tmp --enable-gpl --enable-libx264 --enable-libvpx
+  - cd /tmp/ffmpeg-source && make
+  - cd /tmp/ffmpeg-source && make install
+  - export PATH=/tmp/bin:$PATH
+  - ffmpeg -version
 before_script:
-  - make redis
+  - cd $TRAVIS_BUILD_DIR && make redis
 script:
   make compile_ext ci_test
 notifications:
--- a/docs/gifv.rst
+++ b/docs/gifv.rst
@@ -1,0 +1,53 @@
+gifv
+==========
+
+The gifv optimizer is able to convert gifs to mp4 or webm videos, often resulting in dramatically smaller sized files.
+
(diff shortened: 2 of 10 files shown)


masom:
It would be nice if the option was documented :P

okor:
@masom haha, yes, totally agree. I'll work on that now.

okor:
As noted in docs, latest release of ffmpeg breaks things. I'll see if I can pin the ffmpeg verion in the travis.yml

okor:
@masom I ended up having to build ffmpeg from source, those ppas won't work for 12.04 unfortunately. Also, 12.04 get EOL'ed in 10 days ... so we might want to consider testing on 14.04 or better.

okor:
@scorphus @andreaugusto Any thoughts on this?

scorphus:
@okor, I forgot to mention: good work with the gifv optimizer also!

okor:
@scorphus Hey, you can merge this now if you want but I'm going to add some more robust tests for gifv either way.

okor:
@scorphus I occurs to me that it might be faster to pipe buffers into and out of ffmpeg, instead of using files. Are `input_file` and `output_file` on the disk? If they are it seems like it would be better to use buffers. Let me know what you think.

okor:
Actually looks like `tempfile` is in memory and probably pretty fast. Never mind ; )

andreaugusto:
Just in time: This PR looks great! Nothing to add to the discussion. :)

okor:
Ok, we all seem to be good with this. Gonna' merge.

@scorphus What version do you think this should be bumped to?

scorphus:
Right on! Superb job, @okor! 👍  Thanks once again!

> @scorphus What version do you think this should be bumped to?

I believe we're almost good to go with 6.4.0, as soon as we're done with some PR such as #878.

PR #908, opened 2017-04-17
added new timeout setting if using curl http client

Savar:
instead of an absolute timeout the curl http client also supports a
download-speed timeout (if for X seconds the download speed is below Y
bytes/s then abort)
this is now being configurable but disabled by default (and ignored if
the curl client isn't used)


masom:
👍  🚢

Savar:
@masom I cannot merge.. but feel free to merge it ;)

Issue #740, opened 2016-06-02
Navy blue color changes to black after cropping


ppatelcodal:
### Thumbor request URL

http://fakedomain.com/unsafe/500x500/fakefolder/2/fake_image_name.jpg
### Expected behaviour

Color should not change after cropping.

![actual_image](https://cloud.githubusercontent.com/assets/10356311/15742387/2dcbabe6-28dc-11e6-848f-491fdf58f710.jpg)
### Actual behaviour

navy blue color changed to black.

![after_cropping](https://cloud.githubusercontent.com/assets/10356311/15742386/2dcb28c4-28dc-11e6-9a08-33fdc1e20281.jpg)
### Operating system

Ubuntu
### Your thumbor.conf

RESPECT_ORIENTATION = True
LOADER = 'tc_aws.loaders.s3_loader'
STORAGE = 'tc_aws.storages.s3_storage'
UPLOAD_PHOTO_STORAGE = 'tc_aws.storages.s3_storage'
RESULT_STORAGE = 'tc_aws.result_storages.s3_storage'
SECURITY_KEY = 'blablablablablablalb'
ALLOW_UNSAFE_URL = True
ALLOW_OLD_URLS = True
OPTIMIZERS = [
    'thumbor.optimizers.jpegtran',
    'thumbor.optimizers.gifsicle',
    'thumbor.optimizers.pngcrush',
]
JPEGTRAN_PATH = '/usr/bin/jpegtran'
GIFSICLE_PATH = '/usr/local/bin/gifsicle'
PNGCRUSH_PATH = '/usr/bin/pngcrush'
AWS_ACCESS_KEY   = '<this is secret >'
AWS_SECRET_KEY   = '<this is secret >'
AWS_ACCESS_KEY_ID = '<this is secret >'
AWS_SECRET_ACCESS_KEY = '<this is secret >'
S3_LOADER_BUCKET = '<this is secret >'
STORAGE_BUCKET = '<this is secret >'
RESULT_STORAGE_BUCKET = '<this is secret >'
RESULT_STORAGE_STORES_UNSAFE = True
TC_AWS_REGION='<this is secret >'
TC_AWS_STORAGE_BUCKET='<this is secret >'
TC_AWS_STORAGE_ROOT_PATH='thumbor/storage'
TC_AWS_LOADER_BUCKET='<this is secret >'
TC_AWS_LOADER_ROOT_PATH=''
[...]

Issue #586, opened 2015-10-24
convert CMYK to sRGB filter


Perni1984:
I have several original .jpgs from our photographer saved with ICC profile Coated Fogra27 (ISO 12647-2:2004), which should be CMYK if I understood well.

Is there a filter or another way in thumbor that I can use to deliver "web-safe" sRGB pictures out of those CMYK pictures. Unfortunately I didn't find anything on the filters wiki page or here in the issues.

masom:
Hi @Perni1984 

This is an interesting problem we faced on crated.com

Artists would upload their originals in CMYK + embedded ICC profile. Browser support for CMYK is kinda clunky and as you discovered, the file displays fine on a computer but not on the web.

We discovered in some case a simple convert from CMYK to RGB worked. One issue we faced with the conversion is the loss of colours / parity on some art pieces.

You could use ImageMagick `identify` to determine if you have a CMYK image:
```
identify -verbose something.jpg
...
Colorspace: CMYK
...
```

You can convert a CMYK to RGB using the `convert` command using the original profile:
```
convert original.jpg -negate -profile <original_profile>.icc -profile <adobe rgb profile>.icc output.jpg
```


In thumbor it would be possible to write a filter that will apply the Adobe RGB profile.

From Pillow docs:
```
icc-profile
If present, the image is stored with the provided ICC profile. If this parameter is not provided, the image will be saved with no profile attached. To preserve the existing profile:

im.save(filename, 'jpeg', icc_profile=im.info.get('icc_profile'))
```

Welcome to image correction and colour space hell :)

Let me know if you need help writing the filter althoug if the amount of picture is small it might be faster / better if you let the artist know. Converting from CMYK to RGB usually changes the image.

Perni1984:
Hi @masom, thanks for pointing out the options.

As I am not really python savvy (unfortnately) I will probably use Imagemagick or Graphicsmagick to modify the file after the upload and before serving it o thumbor. 

I was just wondering if I have overseen a filter or something to directly throw a CMYK jpeg onto thumbor and have it being automatically converted to a web safe .jpg without ICC profile.

Thanks. Best.

masom:
@Perni1984 if you really need it "on-the-fly" within thumbor I could probably contribute the filter although there's several small issues.

ICC profiles usually cannot be re-distributed and to do a proper conversion the original profile needs to be available.

Perni1984:
@masom really thanks for your help. 

[...]

PR #927, opened 2017-05-23
RFC: sRGB filter converts images to sRGB using embedded profiles

kkopachev:
Hi! 
Please don't merge, as this is mostly proof of concept to get your comments. I'll add config option and tests, if main code is good enough.

New srgb filter converts images with embedded profiles to sRGB, embedding new tiny sRGB profile (could be stripped later with strip_icc filter)
That should solve issues with CMYK images as well as allow to replace embedded image profiles with tiny sRGB one. 

related issues: https://github.com/thumbor/thumbor/issues/586 https://github.com/thumbor/thumbor/issues/740

Noot sure if it's a good way implementing this, though. Currently filter relies on internal structure of Pil engine, because there is no another way to get image from the engine. `get_image_data` method does not provide access to embedded profile and `image_data_as_rgb` converts image to RGB without proper color management.

Maybe that should all be implemented inside engine and exposed to filter similar as `strip_icc`, but I'm afraid that could create BC break.

Thoughts?

### DIFF
--- a/thumbor/filters/srgb.py
+++ b/thumbor/filters/srgb.py
@@ -1,0 +1,77 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+
+import thumbor.filters
+from thumbor.filters import BaseFilter, filter_method
+from thumbor.utils import logger
+from thumbor.engines.pil import Engine as PILEngine
+
+from io import BytesIO
+
+try:
+    from PIL import ImageCms
+except:
+    ImageCms = None
+
+# Facebook version.
+# See https://www.facebook.com/notes/facebook-engineering/under-the-hood-improving-facebook-photos/10150630639853920
+tiny_srgb = bytearray([0, 0, 2, 12, 108, 99, 109, 115, 2, 16, 0, 0, 109, 110, 116, 114, 82, 71, 66, 32, 88, 89, 90, 32,
+                       7, 220, 0, 1, 0, 25, 0, 3, 0, 41, 0, 57, 97, 99, 115, 112, 65, 80, 80, 76, 0, 0, 0, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 246, 214, 0, 1, 0, 0, 0, 0, 211, 45,
+                       108, 99, 109, 115, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 100, 101, 115, 99, 0, 0, 0,
+                       252, 0, 0, 0, 94, 99, 112, 114, 116, 0, 0, 1, 92, 0, 0, 0, 11, 119, 116, 112, 116, 0, 0, 1, 104,
+                       0, 0, 0, 20, 98, 107, 112, 116, 0, 0, 1, 124, 0, 0, 0, 20, 114, 88, 89, 90, 0, 0, 1, 144, 0, 0,
+                       0, 20, 103, 88, 89, 90, 0, 0, 1, 164, 0, 0, 0, 20, 98, 88, 89, 90, 0, 0, 1, 184, 0, 0, 0, 20,
+                       114, 84, 82, 67, 0, 0, 1, 204, 0, 0, 0, 64, 103, 84, 82, 67, 0, 0, 1, 204, 0, 0, 0, 64, 98, 84,
+                       82, 67, 0, 0, 1, 204, 0, 0, 0, 64, 100, 101, 115, 99, 0, 0, 0, 0, 0, 0, 0, 3, 99, 50, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 116, 101, 120, 116, 0, 0, 0, 0, 70, 66, 0, 0, 88, 89,
+                       90, 32, 0, 0, 0, 0, 0, 0, 246, 214, 0, 1, 0, 0, 0, 0, 211, 45, 88, 89, 90, 32, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


vasekboch:
Exactly. I was trying to write same filter. But stumbled on this line.  There is a implicit conversion here https://github.com/thumbor/thumbor/blob/master/thumbor/engines/pil.py#L186. I So inside filter  self.engine.image.mode returns RGB. So its impossible to apply the embedded profile. Its possible to convert it back to CMYK (I can detect it from ICC profile) but still some data are lost. This needs to be fixed in ```thumbor/engines/pil.py```. And I don't know Thumbor that well, to do so.
Basically the conversion without applying ICC profile resolved to those issues  #586 #740

thumbor/filters/srgb.py
```diff
@@ -0,0 +1,77 @@
...
+                       3, 22, 0, 0, 3, 51, 0, 0, 2, 164, 88, 89, 90, 32, 0, 0, 0, 0, 0, 0, 111, 162, 0, 0, 56, 245, 0,
+                       0, 3, 144, 88, 89, 90, 32, 0, 0, 0, 0, 0, 0, 98, 153, 0, 0, 183, 133, 0, 0, 24, 218, 88, 89, 90,
+                       32, 0, 0, 0, 0, 0, 0, 36, 160, 0, 0, 15, 132, 0, 0, 182, 207, 99, 117, 114, 118, 0, 0, 0, 0, 0,
+                       0, 0, 26, 0, 0, 0, 203, 1, 201, 3, 99, 5, 146, 8, 107, 11, 246, 16, 63, 21, 81, 27, 52, 33, 241,
+                       41, 144, 50, 24, 59, 146, 70, 5, 81, 119, 93, 237, 107, 112, 122, 5, 137, 177, 154, 124, 172,
+                       105, 191, 125, 211, 195, 233, 48, 255, 255])
+
+
+class Filter(BaseFilter):
+    phase = thumbor.filters.PHASE_AFTER_LOAD
```
kkopachev: Filter is applied right after load, before engine does it's work.

kkopachev:
@vasekboch Not sure I understand why would you like to convert image back to CMYK?
That line you pointed out is not only place where forced RGB conversion happens. Filters that call `image_data_as_rgb` will force RGB conversion too.

Apart from hackiness, proposed filter is working.

vasekboch:
Well the problem is, that when I get the image inside the plugin its already in RGB with implicit conversion that does not care about ICCs. For example this image does not work.

plane

I used this filter, installed it and tried my problematic image. But the result is this error.

```
2017-06-01 19:54:43 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/tornado/gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "/usr/local/lib/python2.7/site-packages/thumbor/handlers/__init__.py", line 207, in get_image
    self.filters_runner.apply_filters(thumbor.filters.PHASE_AFTER_LOAD, transform)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 91, in apply_filters
    exec_one_filter()
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 90, in exec_one_filter
    f.run(exec_one_filter)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 198, in run
    results.append(self.runnable_method(*self.params))
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 23, in wrapper
    return fn(self, *args2)
  File "/usr/local/lib/python2.7/site-packages/thumbor_srgb_filter/srgb.py", line 74, in srgb
    outputMode=output_mode)
  File "/usr/local/lib/python2.7/site-packages/PIL/ImageCms.py", line 355, in profileToProfile
    raise PyCMSError(v)
PyCMSError: cannot build transform
```

I was digging around and found out that, ```self.engine.image.mode``` returns RGB. That is the reason why the profileToProfile fails. Due to the implicit conversion.

I changed the lines to this:
```
transform = ImageCms.buildTransform(embedded_profile, srgb_profile, "CMYK", output_mode)
image = ImageCms.applyTransform(image, transform)
```
This returns obvious nonsense. But applyTransformation works withnout error. Tweaking code to 
```
image = image.convert("CMYK")
transform = ImageCms.buildTransform(embedded_profile, srgb_profile, "CMYK", output_mode)
image = ImageCms.applyTransform(image, transform)
```
Works for this example, but the colors are a little bit wrong due to CMYK -> RBG -> CMYK -> RBG conversion.

Running it like this works correctly as expected.
```
from PIL import Image
from PIL import ImageCms
from io import BytesIO

image = Image.open('plane.jpg')

print(image.mode)

embedded_profile = image.info.get('icc_profile')
embedded_profile = BytesIO(embedded_profile)

srgb_profile = ImageCms.createProfile("sRGB")
transform = ImageCms.buildTransform(embedded_profile, srgb_profile, image.mode, "RGB")
image = ImageCms.applyTransform(image, transform)

print(image.mode) # Output: RGB

image.save('plane_new.jpg')
```
 You see why am I asking about the implicit conversion to sRGB. For this image it does not work correctly due to the ICC and CMYK combo. :)

kkopachev:
That's weird, because it works just fine on my setup. No errors and output in srgb. I specifically tested filter for CMYK image with embedded ICC profile. 
http://localhost:8001/unsafe/fit-in/1024x1024/filters:srgb()/https://cloud.githubusercontent.com/assets/1482932/26698213/5fa2705c-4715-11e7-9570-072611fad259.jpg

vasekboch:
Ahhh OK, well that is really weird. 
Before I was running 6.3.0. I tried to upgrade to 6.3.2. And even your link throws the same error.
What could be wrong, that would cause this behaviour? 
I've just installed fresh thumbor through pip and created simple setup.py for the srgb.py file and installed it as package. All in default configuration with added FILTER.

vasekboch:
@kkopachev I was digging around and located my issue to ```STORAGE="thumbor.storages.file_storage"``` If I use ```STORAGE="thumbor.storages.no_storage"``` option everything works as expected. But when I set storage to default eq. ```thumbor.storages.file_storage``` it fails on error that I was receiving before. With no_storage works as a charm. I was suing everything on default, so that's why it was failing. 

I would love to help, but my thumbor and python knowledge is very limited. But If I found anything new, I will let you know.

kkopachev:
@vasekboch found it: https://github.com/thumbor/thumbor/pull/930

vasekboch:
@kkopachev Cool, thx man. We will test it if has some side effects. :)

masom:
@kkopachev I think this should probably be embedded in the pil engine and enabled with a config variable.

Knowing if the image is CMYK requires prior knownledge at upload time. It would be great if the pil engine could just detect the CMYK mode and automatically transform to sRGB when enabled.

Having it as a filter is also a good option but it could just call in the pil engine to apply the profile conversion?

JOGUI22:
@masom yeah, i think just the same. Could be awesome if thumbor automatically change CMYK profiles to sRGB.

I think is quite an important issue, because when an user uploads an image like this one in CMYK:

cystart

Thumbor outputs this version, with changed colors:

cyend

znerol:
I have thumbor deployed as part of an automated print-first workflow. I resolved exact the same issue with a custom filter as well: https://github.com/znerol/thumbor-icc

I also think that `image_data_as_rgb` should apply embedded color profiles (if PIL is compiled with color management support).

kkopachev:
@znerol could you explain a little more what do you mean by 
> I also think that image_data_as_rgb should apply embedded color profiles

znerol:
https://xkcd.com/1882/

xkcd

heynemann:


```diff

Complexity increasing per file
==============================
- thumbor/filters/srgb.py  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=4967917&bid=16414100)

karptonite:
This definitely looks like something we could use--we are running into issues with a small number of user provided images with gigantic (700k) ICC profiles. I don't want to strip them willy-nilly, of course, but I also don't want a few giant images cause issues for our users. Is this on track to be added to Thumbor, or should I instead look at @znerol's custom filter?

lastthyme (approved):

heynemann:
Closing this PR since we had no activity in two years in the interest of keeping our PR list leaner. If this is something we still want to pursue, please reopen and I'll look into it. Thanks!

PR #928, opened 2017-05-30
Allow custom LoaderResult errors

gi11es:
Being constrained to the hardcoded list is too limiting.
Using the actual codes is also less prone to confusion.

### DIFF
--- a/tests/handlers/test_base_handler.py
+++ b/tests/handlers/test_base_handler.py
@@ -19,6 +19,7 @@ import subprocess
 from json import loads
 
 import tornado.web
+from tornado.concurrent import return_future
 from preggy import expect
 from mock import Mock, patch
 import unittest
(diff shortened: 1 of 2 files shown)


thumbor/loaders/__init__.py
```diff
@@ -10,11 +10,6 @@
 
 
 class LoaderResult(object):
-
-    ERROR_NOT_FOUND = 'not_found'
```
masom: @gi11es Removing those const would be backward incompatible. For thumbor 6 it should be possible to support both constants and response codes.

```
                elif result.loader_error == LoaderResult.ERROR_TIMEOUT:
                     # Return a Gateway Timeout status if upstream timed out (i.e. 599)
                     self._error(504)
                 elif isinstance(result.loader_error, int):
                     self._error(result.loader_error)
                      return
```

gi11es:
Fair enough, I'll update the PR to that effect

gi11es:
@masom your concerns should be addressed now.

gi11es:
Merge pls? :)

masom:
Thanks @gi11es for rebasing this. 🚀

PR #930, opened 2017-06-06
Do not re-read image data coming from loader to store in storage

kkopachev:
I'm not completely sure on this change, but tests pass and that would fix some weird hidden issues it could be causing.

What's the value of calling engine.read method on buffer, containing image, which we just retrieved from loader? `engine.read` re-saves image taking current quality settings and other configuration into account, so stored copy will be different from original.
This behavior forces CMYK image with embedded CMYK ICC profile to be saved as RGB image with CMYK ICC profile. https://github.com/thumbor/thumbor/pull/927 

Also, if we save whatever bytes we fetched from loader, then we could eliminate issues like this: https://github.com/thumbor/thumbor/pull/901

Originally introduced [here](https://github.com/kkopachev/thumbor/commit/492659c5e9913f19aa653a63ceca978b35bb4c3e#diff-bfc9541389c9ec7130c9a9b3a7219ed3R54)

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -10,7 +10,6 @@
 
 import sys
 import functools
-import copy
 import datetime
 import re
 import pytz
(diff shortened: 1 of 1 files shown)


kkopachev:
any news here?

kkopachev:
Bump. Do you guys have any comments/concerns or should I change something?

masom:
@kkopachev I'm not sure why re-reading is necessary. Let's merge this and under an beta/rc candidate we'll be able to collect feedback of any weird behaviours.

PR #889, opened 2017-03-14
ensuring no exception if filter is async

Savar:
while trying something out I realized that we get exceptions on the `watermark` filter in the background because this filter is an  `async_filter` and requires a callback. All others will ignore the callback so I think it is safe to always pass this dummy callback.

How to see the error, apply this patch:
```diff
simon@ironman:~/repos/thumbor$ git diff
diff --git a/tests/filters/test_watermark.py b/tests/filters/test_watermark.py
index 93b4022..decc1b2 100644
--- a/tests/filters/test_watermark.py
+++ b/tests/filters/test_watermark.py
@@ -54,4 +54,4 @@ class WatermarkFilterTestCase(FilterTestCase):
         image = self.get_filtered('source.jpg', 'thumbor.filters.watermark', 'watermark(watermark.png,30,-50,60)')
         expected = self.get_fixture('watermarkSimple.jpg')
         ssim = self.get_ssim(image, expected)
-        expect(ssim).to_be_greater_than(0.98)
+        expect(ssim).to_be_greater_than(0.9999)
diff --git a/thumbor/filters/watermark.py b/thumbor/filters/watermark.py
index 5d7a73b..ab2d853 100644
--- a/thumbor/filters/watermark.py
+++ b/thumbor/filters/watermark.py
@@ -131,8 +131,12 @@ class Filter(BaseFilter):
         self.watermark_engine = self.context.modules.engine.__class__(self.context)
         self.storage = self.context.modules.storage

-        buffer = yield tornado.gen.maybe_future(self.storage.get(self.url))
-        if buffer is not None:
-            self.on_image_ready(buffer)
-        else:
-            self.context.modules.loader.load(self.context, self.url, self.on_fetch_done)
+        try:
+            buffer = yield tornado.gen.maybe_future(self.storage.get(self.url))
+            if buffer is not None:
+                self.on_image_ready(buffer)
+            else:
+                self.context.modules.loader.load(self.context, self.url, self.on_fetch_done)
+        except Exception as e:
+            logger.exception(e)
+            raise e
```

and run the tests and you will see:
```sh
test_watermark_filter_centered (tests.filters.test_watermark.WatermarkFilterTestCase) ... ok
test_watermark_filter_centered_x (tests.filters.test_watermark.WatermarkFilterTestCase) ... ok
test_watermark_filter_centered_y (tests.filters.test_watermark.WatermarkFilterTestCase) ... ok
[...]

PR #932, opened 2017-06-06
This test should fail, but it does not.

kkopachev:
I tried to dig around and fix that thing, but can't find out how to do so. 
Essentially I want to test tat exception is thrown inside async filter. 

Test tries to apply non existing watermark file to an image. Filter throws exception, but it never surfaced. 
Somewhat related to https://github.com/thumbor/thumbor/pull/889

### DIFF
--- a/tests/filters/test_watermark.py
+++ b/tests/filters/test_watermark.py
@@ -55,3 +55,10 @@ class WatermarkFilterTestCase(FilterTestCase):
         expected = self.get_fixture('watermarkSimple.jpg')
         ssim = self.get_ssim(image, expected)
         expect(ssim).to_be_greater_than(0.98)
+
+    def test_watermark_filter_should_fail(self):
+        image = self.get_filtered('source.jpg', 'thumbor.filters.watermark',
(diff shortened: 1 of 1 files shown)


Savar:
I think the issue is, that all the `future` stuff is not throwing it in the moment you expect so the data is not changed and the filter is somehow "half working"... but you should check

```
    def test_getting_invalid_watermark_returns_bad_request(self):
        response = self.fetch('/unsafe/filters:watermark(boom.jpg,0,0,0)/image.jpg')
        expect(response.code).to_equal(400)
```
in `tests/handlers/test_base_handler.py` and maybe use better this one to see if you get the correct response?!

kkopachev:
hrm.. Issue, as I see it, is that async filters are relying on callback to be actually called eventually. However, if exception occurred, it's buried in Future, which is never retrieved, hence request just hangs until Future-timeout is reached.

heynemann:
Is this still an issue?

kkopachev:
Yes. Solvable by #1000, if we want to re-evaluate it.

kkopachev:
Fails as expected in #1257

PR #940, opened 2017-06-28
Add new strip_exif filter in order to remove exif on demand

franek:
This filter removes any Exif information in the resulting image.
This is useful if you have set the configuration PRESERVE_EXIF_INFO = True but still wish to overwrite this behaviour in some cases (e.g. for image icons)


PascalThuet:
We badly need this feature. Would it be possible to draft a new release for this ? Think this would help a lot of people optimizing their web performance and make the www a better place :)

PR #511, opened 2015-07-18
CHANGELOG replaces stale History.md

pda:
`History.md` is woefully out of date, with a latest entry of `3.7.1 / 2013-02-06`.

The [Release-notes] page on the wiki is marked as deprecated and links to the [GitHub releases page][releases], which is fine because both are online on GitHub.

Rather than just delete `History.md` I thought it might be useful to maintain an in-repo / offline copy of the release history. The [hub] CLI tool makes this pretty easy[*][caveat]:

```sh
hub release | tr -d '\r' > CHANGELOG
```

---

The file could be `History` or `History.txt` rather than `CHANGELOG`, but keeping the `.md` extension would necessitate some reformatting of the `hub release` output, which didn't seem worthwhile.

Alternatively the `History.md` file could just be deleted, or have a link to [releases] added to the top.

[Release-notes]: https://github.com/thumbor/thumbor/wiki/Release-notes
[releases]: https://github.com/thumbor/thumbor/releases
[hub]: https://github.com/github/hub
[caveat]: https://github.com/github/hub/pull/961

PR #958, opened 2017-08-23
CHANGELOG: add 5.0.5 to 6.3.2 from GitHub releases.

pda:
Uses https://github.com/github/hub to import releases 5.0.5 to 6.3.2 from GitHub Releases into CHANGELOG:

    hub release | sed -e 's/\s*$//' -e 's/\r//' > CHANGELOG

Note `hub release` doesn't handle pagination (https://github.com/github/hub/pull/961) so care must be taken not to commit the truncation of the file. I use `git add -p` for that.

The CHANGELOG was introduced in https://github.com/thumbor/thumbor/pull/511 to replace an out of date manually maintained `History.md` file. This approach of pulling from GitHub, while not hand-written, is also likely to be neglected. But it's nice to have a CHANGELOG in the repo. :man_shrugging:


masom:
thanks @pda!

Issue #959, opened 2017-08-23
We are testing against Python 2.7.3, not 2.7.13


cclauss:
https://github.com/thumbor/thumbor/blob/master/.travis.yml#L3

I do not understand well enough to fix it but we should not test against such an old version.

masom:
Some of it has to do with supporting Ubuntu precise on TravisCI. We could probably look at refreshing the CI pipeline or just also support 2.7.13

cclauss:
We are [now on trusty](https://travis-ci.org/thumbor/thumbor/jobs/271430245#L4), not precise.

hugovk:
This can be closed, Python 2.7.13 is now being tested.

https://travis-ci.org/thumbor/thumbor/jobs/301103433#L560

cclauss:
Just in time for Python 2.7.14 ;-)

PR #968, opened 2017-09-03
Add Python 2.7.13 to the testing

cclauss:
We are currently testing 2.7.3 which is a bit behind the times.  See #959


kkopachev:
I have something in progress for python 2.7.13. https://github.com/kkopachev/thumbor/pull/1 I'll make a PR when finish upgrading redeye filter to use cv2 instead of cv.

kkopachev:
this could be closed

PR #969, opened 2017-09-04
Fixed flake8 issues

marcelometal:
Please, use rebase instead of merge.


scorphus:
Good one, @marcelometal. Thanks!

PR #974, opened 2017-09-06
Updates for Travis: new python

kkopachev:
* Updated red eye filter to work with cv2 and add test
* Fix noise filter test
* Updated install scripts to use opencv-python from pip
* Fix skipped base_imaging handler test
* Added more caching for travis

FYI @cclauss


.travis.yml
```diff
@@ -1,62 +1,54 @@
...
+  - 2.7
+  - 3.5
+  - 3.6
 cache:
+  pip: true
+  apt: true
   directories:
-    - $HOME/.cache/pip
+    - /tmp/ffmpeg/bin
 sudo: false
```
cclauss: My take is you don't need the sudo and dist lines because these are not Travis defaults and it would be a bummer to lock in trusty when Travis updates distro but this is minor stuff.

.travis.yml
```diff
@@ -1,62 +1,54 @@
 language: python
```
cclauss: Really cool stuff!!  Just some minor comments...

.travis.yml
```diff
@@ -1,62 +1,54 @@
...
-    - python-numpy
-    - python-scipy
-    - libcairo2-dev
-    - gfortran
-    - libopenblas-dev
-    - liblapack-dev
-    - python-pyexiv2
     - yasm
 install:
   - pip install -U --upgrade pip
```
cclauss: Could this be `--user` instead of `-U`?  Explicit is better than...  I forgot to do this on an earlier PR.

kkopachev:
should be all good now

marcelometal:
Hi @kkopachev,

Who is the copyright holder of tests/fixtures/filters/redeye.png, could you please add this info and the license?

kkopachev:
@marcelometal [It's public domain](https://commons.wikimedia.org/wiki/File:BoldRedEye.JPG)

marcelometal:
@kkopachev, maybe here [tests/fixtures/filters/README.mkd](https://github.com/thumbor/thumbor/blob/master/tests/fixtures/filters/README.mkd) ?

kkopachev:
@marcelometal done! Thanks.

.travis.yml
```diff
@@ -1,62 +1,52 @@
...
+  - 2.7
+  - 3.5
+  - 3.6
 cache:
+  pip: true
+  apt: true
   directories:
-    - $HOME/.cache/pip
-sudo: false
+    - /tmp/ffmpeg/bin
```
masom: 🎉

scorphus:
Hey, @kkopachev. Can you please:

- Squash commits that are related to each other (like fix commits and/or commits that undo things done by an earlier commit, such as the miniconda in and out)
- Rebase the commits on top of current master (a.k.a. resolve conflicts)

Thank you very much!

kkopachev:
@scorphus do you prefer a way of pulling ffmpeg off of zesty repo - requires sudo on travis, might be slow to start the test. Or preserve ffmpeg compile stage and use travis cache?

scorphus:
Hey @kkopachev. Glad you're still with us! 🤗 

@marcelometal's PR – #975 – that installs ffmpeg from zesty was already merged, so I guess we're good on that.

kkopachev:
updated

setup.py
```diff
@@ -25,13 +25,13 @@
     "nose-focus",
     "colorama",
     "numpy",
-    "scipy>=0.16.1,<1.0.0",
     "flake8",
     "yanc",
     "remotecv",
-    "pyssim",
+    "pyssim>=0.4.0",
```
scorphus: I'm wondering why we're setting a lower threshold for `pyssim` here. There's nothing different in tests nor in docs on this commit. I looked briefly but could not see any relation between pyssim >= 0.4 and latest Python versions. Excuse me if I missed something.

setup.py
```diff
@@ -25,13 +25,13 @@
...
     "cairosvg>=1.0.0,<2.0.0,!=1.0.21",
     "preggy>=1.3.0",
+    "opencv-python",
```
scorphus: Here we're making this a mandatory dependency, which is not the case for Thumbor, I believe. Only if the user decides to use [such engine](/thumbor/opencv-engine) and [configure Thumbor to do so](http://thumbor.readthedocs.io/en/latest/configuration.html#engine) and in such cases he/she should install the desired engine. Perhaps docs regarding this should be improved.

setup.py
kkopachev: so, this is only a list of test dependencies. 
previously, scipy was manually included, but never used directly anywhere in the code. Only usage was through pyssim. 
Pyssim, in it's turn, does not seem to install its dependencies automatically, up to latest release (0.4). https://github.com/jterrace/pyssim/issues/24
Now that pyssim can automatically install it's own dependencies, I decided to remove explicit require of `scipy`, as it was not directly used and let pip's pyssim do the magic.

setup.py
kkopachev: This is a list of test dependencies. opencv could be used in PIL engine to convert 16bit TIFFs to png.

kkopachev: ah, it's also used in redeye filter and detectors. And in order to test that it's all required.

.travis.yml
```diff
@@ -1,23 +1,23 @@
 language: python
 python:
-  - 2.7_with_system_site_packages
```
kkopachev: so, `with_system_site_packages` thing forces travis to use whatever python is shipped with distribution. IIRC it's 2.7.6 at the moment. Then there were some problems installing opencv... It was easier to go with virtualenv and install packages from pip instead (reason to upgrade from cv/cv2 to cv3)

kkopachev:
Any updates here?

.travis.yml
masom: This is a nice improvement.

setup.py
scorphus: Oh, my bad, didn't notice that diff regarded changes made to `test_require`.

.travis.yml
```diff
@@ -28,38 +28,22 @@ env:
...
+    - libboost-python-dev
+    - libexiv2-dev
 before_install:
   - echo "deb http://archive.ubuntu.com/ubuntu zesty main universe" | sudo tee --append /etc/apt/sources.list
   - sudo apt-get update
   - sudo apt-get install ffmpeg -y -f
 install:
-  - pip install -U --upgrade pip
+  - pip install --upgrade pip
+  - pip install -I https://github.com/escaped/pyexiv2/archive/master.zip
```
marcelometal: Hi @kkopachev, `master` branch may not be completely stable at a given moment and could break Thumbor build. What do you think about installing this dependency via .deb package?

kkopachev: not sure I can use deb to make it automatically available in virtual env. I can reference commit hash instead.

kkopachev: pyexiv2 is dead anyway, not likely there will be any releases.

marcelometal: Currently, pyexiv2 is installed via [.deb package](https://github.com/thumbor/thumbor/blob/master/.travis.yml#L52) (Travis) and the [tests are running]( https://travis-ci.org/thumbor/thumbor/jobs/274855100#L2641).

kkopachev: current travis setup uses python with system provided packages, that's why it's working.
However, not working if we don't use system python and use virtualenv: https://travis-ci.org/thumbor/thumbor/jobs/280505386#L1689

kkopachev: Also, if you guys care that much about something breaking the build, there are other issues as well https://github.com/thumbor/thumbor/pull/975/files#r141397089

kkopachev:
any more comments?

marcelometal:
Hi @kkopachev, Good work! 

Thank you for contributing to Thumbor!

scorphus:
Merged in 2f1529604a0f5b2d6d87132b5616841842313215 🎉  You are a rockstar, @kkopachev! 🚀  Thank you for your contributions to Thumbor!

PR #975, opened 2017-09-07
Travis: Improve build time

marcelometal:
Install binary ffmpeg (3.3.3) package instead of build it

https://launchpad.net/~mc3man/+archive/ubuntu/trusty-media

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 cache:
   directories:
     - $HOME/.cache/pip
-sudo: false
+sudo: required
 matrix:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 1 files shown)


kkopachev:
there seems to be known issues with ffmpeg 3.3: https://github.com/thumbor/thumbor/pull/907

marcelometal:
Hi @kkopachev, thanks. I will change to 3.2.4.

marcelometal:
Done! ;)

scorphus:
We gained a lot, and TravisCI did as well! Thanks, @marcelometal!

.travis.yml
```diff
@@ -43,20 +43,15 @@ addons:
     - liblapack-dev
     - python-pyexiv2
     - yasm
+before_install:
+  - echo "deb http://archive.ubuntu.com/ubuntu zesty main universe" | sudo tee --append /etc/apt/sources.list
```
kkopachev: Zesty going to EOL in January 2018. Soon after that repos will become unavailable and it will be impossible to install ffmpeg, which will break the build.

marcelometal: Take a look at the [ubuntu archive](http://archive.ubuntu.com/ubuntu/dists/), [vivid EOL](https://wiki.ubuntu.com/Releases) was in February 4, 2016 and it was not removed.

All ubuntu repositories are available here: https://packages.ubuntu.com
For old or non-supported versions here: http://old-releases.ubuntu.com/ubuntu/dists/

If one day the build breaks, we can fix it changing the `archive.ubuntu.com` to `old-releases.ubuntu.com`

But we only depend on ffmpeg 3.2.4, because [this bug](https://trac.ffmpeg.org/ticket/6294) . As soon as it is fixed, we can move on.

PR #977, opened 2017-09-09
Remove debian folder

marcelometal:
This folder is only for Debian, so I've moved it to Debian VCS:
https://anonscm.debian.org/cgit/python-modules/packages/thumbor.git


scorphus:
Thanks, @marcelometal 👍

PR #981, opened 2017-09-13
Docs: Fixed getting started

marcelometal:
waterfalls.hamilton.ca domain does not exist


SergioJorge:
Thanks, @marcelometal 👍

PR #982, opened 2017-09-19
Replaced 'pexif' dependency with 'piexif'

marcelometal:
Replacing because piexif supports Python 2.7, 3.3, 3.4, 3.5, pypy, and pypy3; pexif only supports Python 2.7.


SergioJorge:
Thanks, @marcelometal 👍

PR #983, opened 2017-09-19
Fixed urllib Python 2 and 3 compatibility

marcelometal:
Added six package as dependency. six is a package that helps in writing code
that is compatible with both Python 2 and Python 3


SergioJorge:
Thanks, @marcelometal 👍

PR #989, opened 2017-09-25
Optimize resizing of some jpegs using pillow draft

kkopachev:
http://pillow.readthedocs.io/en/3.4.x/reference/Image.html#PIL.Image.Image.draft
In some cases might be faster https://github.com/thumbor/thumbor/pull/858#issuecomment-279519233
Works only for jpegs, does nothing with others.


heynemann:
Any side effects here @kkopachev ?

kkopachev:
No. That's essentially the [same as calling thumbnail](http://pillow.readthedocs.io/en/stable/reference/Image.html?highlight=thumbnail#PIL.Image.Image.thumbnail)

PR #995, opened 2017-10-11
Fix transformer tests and fix exception thrown when fit-in + requested size has "orig"

kkopachev:
/cc @Savar

### DIFF
--- a/tests/fixtures/transformer_test_data.py
+++ b/tests/fixtures/transformer_test_data.py
@@ -945,7 +945,7 @@ FIT_IN_CROP_DATA = [
         focal_points=[],
         crop_left=None, crop_top=None, crop_right=None, crop_bottom=None,
         fit_in=True
-    ), (200, 100, 1)),
+    ), (200, 100, True)),
 
     (TestData(
         source_width=1000, source_height=250,
@@ -954,7 +954,7 @@ FIT_IN_CROP_DATA = [
         focal_points=[],
         crop_left=None, crop_top=None, crop_right=None, crop_bottom=None,
         fit_in=True
-    ), (500, 125, 1)),
+    ), (500, 125, True)),
 
     (TestData(
         source_width=200, source_height=250,
@@ -963,7 +963,7 @@ FIT_IN_CROP_DATA = [
         focal_points=[],
         crop_left=None, crop_top=None, crop_right=None, crop_bottom=None,
         fit_in=True
-    ), (200, 250, 0)),
+    ), (200, 250, False)),
 
     (TestData(
         source_width=800, source_height=400,
@@ -972,7 +972,7 @@ FIT_IN_CROP_DATA = [
         focal_points=[],
         crop_left=None, crop_top=None, crop_right=None, crop_bottom=None,
         fit_in=True, adaptive=True
-    ), (200, 100, 1)),
+    ), (200, 100, True)),
 
     (TestData(
         source_width=800, source_height=400,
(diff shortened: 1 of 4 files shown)


tests/test_transformer.py
```diff
@@ -117,37 +111,25 @@ def test_can_resize_images_with_detection_error_not_ignoring_it(self):
...
     def test_can_fit_in(self):
-        for item in FIT_IN_CROP_DATA:
-            context = item[0].to_context()
+        for (test_data, (width, height, should_resize)) in FIT_IN_CROP_DATA:
+            context = test_data.to_context()
             engine = context.modules.engine
 
             trans = Transformer(context)
-            trans.transform(self.validate_fit_in(item, context, engine, trans))
+            trans.transform(lambda: None)
```
kkopachev: Reason for this change is that `expect` failure called inside this callback is never surfaced, so test passes.
I made a commit where [make it always fail inside callback ](https://github.com/kkopachev/thumbor/commit/4811ce64ba0c8f33a20ca1dfdf47c4c6f3e2b3b9) and [it was still passing](https://travis-ci.org/kkopachev/thumbor/jobs/286708390)

thumbor/transformer.py
```diff
@@ -343,8 +343,10 @@ def fit_in_resize(self):
         # ensure that filter should work on the real image size and not on the request
         # size which might be smaller than the resized image in case `full-fit-in` is
         # being used
-        self.context.request.width = int(max(self.context.request.width, resize_width))
-        self.context.request.height = int(max(self.context.request.height, resize_height))
+        requested_width = source_width if self.context.request.width == 'orig' else self.context.request.width
```
kkopachev: fixes https://github.com/thumbor/thumbor/pull/877#discussion_r144110881

tests/fixtures/transformer_test_data.py
```diff
@@ -999,6 +999,33 @@ def has_cropped_properly(self):
...
+        target_width="orig", target_height=250,
+        halign="middle", valign="middle",
+        focal_points=[],
+        crop_left=None, crop_top=None, crop_right=None, crop_bottom=None,
+        fit_in=True
+    ), (250, 250, True)),
+
+    (TestData(
+        source_width=500, source_height=500,
+        target_width="orig", target_height="orig",
```
kkopachev: added cases to catch https://github.com/thumbor/thumbor/pull/877#discussion_r144110881

thumbor/transformer.py
Savar: good catch... completely forgot about that.. shall we maybe add a test next to https://github.com/thumbor/thumbor/pull/877/commits/28d9eab34d7bbd630897d4dae85f11058196de5e to check that this behaviour is correct?

kkopachev: done

Savar:
LGTM

Am 16. Oktober 2017 19:05:14 MESZ schrieb Konstantin Kopachev <notifications@github.com>:
>kkopachev commented on this pull request.
>
>
>
>> @@ -343,8 +343,10 @@ def fit_in_resize(self):
># ensure that filter should work on the real image size and not on the
>request
># size which might be smaller than the resized image in case
>`full-fit-in` is
>         # being used
>-        self.context.request.width =
>int(max(self.context.request.width, resize_width))
>-        self.context.request.height =
>int(max(self.context.request.height, resize_height))
>+        requested_width = source_width if self.context.request.width
>== 'orig' else self.context.request.width
>
>done
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/thumbor/thumbor/pull/995#discussion_r144907415

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #1007, opened 2017-11-12
Fix lint problems

rht:


marcelometal:
Hi @rht, thanks for contributing to Thumbor!

Actually this bug was fixed here b55e24aea7b162f91ab626f6c936a973c9421139 , so I'm closing it.

Thanks again!

PR #1021, opened 2017-12-29
Pass through querystring params via new `HTTP_LOADER_FORWARD_QUERYSTRING*` settings

mrmachine:
This allows you to generate thumbnails for any number of private AWS S3 buckets with a single Thumbor server, as long as you pass absolute signed AWS S3 URLs to Thumbor.

This follows the pattern established with the `HTTP_LOADER_FORWARD_*HEADERS*` settings.

Case is ignored when comparing parameter names against the whitelist.

To prevent tampering, querystring parameters being passed through are not allowed to override or extend existing querystring parameters in the urlencoded original image URL.

NOTE:

Any querystring parameters that are not temporary/ephemeral, and which affect the original image to be downloaded, should be specified as part of the urlencoded original image URL, NOT as part of the Thumbor URL.

Querystring parameters in the Thumbor URL are (still) not considered to be part of the original image URL, which is used as a cache key, which would mean multiple requests for the same image with different signatures would be downloaded and processed multiple times.

A consequence of this is that you should NOT allow "unsafe" Thumbor URLs when passing through auth credentials as querystring parameters. Once Thumbor has successfully processed and cached the image for such a request, subsequent requests could access the cached image without providing any auth credentials.

Instead, you are expected to rely on the HMAC validation for "safe" Thumbor URLs, and ensure that clients are authorized before generating safe URLs for private images that require auth credentials.

The same is true for auth credentials passed through as request headers.

Likewise, a similar issue exists when allowing unsafe URLs with the AWS S3 loader, where Thumbor itself generates the AWS S3 signatures.

### DIFF
--- a/thumbor/loaders/http_loader.py
+++ b/thumbor/loaders/http_loader.py
@@ -13,7 +13,7 @@ import re
 from functools import partial
 
 import tornado.httpclient
-from six.moves.urllib.parse import quote, unquote, urlparse
+from six.moves.urllib.parse import quote, unquote, urlparse, urlunparse, parse_qs, urlencode
 
 from . import LoaderResult
 from thumbor.utils import logger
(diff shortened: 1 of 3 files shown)


thumbor/loaders/http_loader.py
```diff
@@ -130,6 +130,26 @@ def load_sync(context, url, callback, normalize_url_func):
         user_agent = context.config.HTTP_LOADER_DEFAULT_USER_AGENT
 
     url = normalize_url_func(url)
+
+    # Pass through querystring parameters from Thumbor URL when fetching
+    # original image URL.
+    query = parse_qs(context.request.query)
+    if not context.config.HTTP_LOADER_FORWARD_QUERYSTRING:
```
masom: I would recommend also short-circuiting when the whitelist is empty.

heynemann:
Sorry for the long time this has been here, but we don't feel like this should be in thumbor's codebase. Thumbor's nature of being extensible allows you to easily implement and use your own loader. If you require this functionality, just implement a custom loader and use it. Maybe the AWS Loader should support this, but then an issue there should be created. Sorry for the long time this has been sitting here unattended.

PR #1040, opened 2018-02-11
don't decode twice and encode if not required

Savar:
the watermark filter was actually decoding the watermark image 2x
and in case the NoStorage was being used it was also encoding it 1x
without a real need to do so.

Test:

http://localhost:8888/unsafe/filters:watermark(https://somewhere.com/image.jpg,-10,-10,20)/https://somewhereelse.com/another_image.jpg

Before the change the following engine calls were made:

engine.create_image() # main image decode
engine.create_image() # watermark decode 1st
engine.read()         # watermark encode => storage
engine.create_image() # watermark decode 2nd
engine.read()         # final image generation

After the change (without storage):

engine.create_image() # main image decode
engine.create_image() # watermark decode
engine.read()         # final image generation

### DIFF
--- a/thumbor/filters/watermark.py
+++ b/thumbor/filters/watermark.py
@@ -147,8 +147,7 @@ class Filter(BaseFilter):
         else:
             buffer = result
 
-        self.watermark_engine.load(buffer, None)
-        self.storage.put(self.url, self.watermark_engine.read())
+        self.storage.put(self.url, buffer)
         self.storage.put_crypto(self.url)
         self.on_image_ready(buffer)
 


Savar:
testing is kind of hard, because the watermark tests are not really loading anything..

I tested it mainly with the opencv engine without Storage.. but because it doesn't really change with a normal storage except the "not decode twice". The one thing I'm concerned is in pillow the:
```
            if self.image.mode != 'RGB':
                self.image = self.image.convert('RGB')
```
and
```
        try:
            if ext == '.webp':
                if self.image.mode not in ['RGB', 'RGBA']:
                    if self.image.mode == 'P':
                        mode = 'RGBA'
                    else:
                        mode = 'RGBA' if self.image.mode[-1] == 'A' else 'RGB'
                    self.image = self.image.convert(mode)

            if ext in ['.png', '.gif'] and self.image.mode == 'CMYK':
                self.image = self.image.convert('RGBA')
            self.image.format = FORMATS.get(ext, FORMATS[self.get_default_extension()])
```
so this might make it necessary for the pillow engine to load/decode the image again to reset everything but I'm not sure.. so in the worst case we need to remove the `loaded = True` thing and load it twice in case there is a storage backend.

@kkopachev @heynemann any idea if this is a problem?

kkopachev:
What was the reason to load (from buffer) and then save (`.read()`) the image before putting it into storage? Is it possible to simply put into storage whatever we get from loader? That would allow to simplify some code.

Savar:
Yeah I was wondering as well but I'm not sure if it is to avoid heavy decoding if the watermark is used often and the source format is expensive or whatsoever. I'm happy to just safe the buffer..

Am 14. Februar 2018 21:06:25 MEZ schrieb Konstantin Kopachev <notifications@github.com>:
>What was the reason to load (from buffer) and then save (`.read()`) the
>image before putting it into storage? Is it possible to simply put into
>storage whatever we get from loader? That would allow to simplify some
>code.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/thumbor/thumbor/pull/1040#issuecomment-365729038

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

kkopachev:
This seems similar to re-read issue for main flow https://github.com/thumbor/thumbor/pull/930#issuecomment-324403142
I'd vote for removing re-read part completely, which will unify code a bit.

Savar:
I think it still has two reasons why it does that:

a) in case the image isn't loadable by the engine the cache would have a useless image instead of checking for a fixed one.. this is questionable because if the image isn't working after some time instead of failing fast and cheap it would download and fail every time.. it might be okay if the cache expiration would be not too long but I guess this would actually require something bigger like having object specific cache time (could be done via `touch` into the future near to the expiry date probably but thats more of a hack)
b) streamlining and optimizing images so that the image is in a maybe easier format/structure than the original one. The same happened as you pointed out also in the normal flow. I see that your change was merged.. anyone having had an issue with that?

Savar:
@kkopachev followed your advice.. this makes it much simpler and in my opinion also clearer. The only concern is the point a)

@heynemann what do you think?

Savar:
PS: I don't agree with coveralls.. 2 lines changed and the coverage changed??

kkopachev:
@Savar 
b) I don't see any benefits of recompressing image. It'll add up to CPU time to decode+encode and it will introduce quality loss. In my opinion, it's always better to work with best available originals.
a) Not sure I understand the problem here. Could you elaborate? Cases where image used to work in the past, but stopped working seems like a regression which should be fixed. What if I'll have an image in future which will fail same way to read? I'd like the issue to be fixed.
Caching wasteful bytes of non-image or broken image might be an issue for `allow_unsafe = true` cases, but not sure that would be biggest concern at that point.

heynemann:
This is sensible. Sorry for the delay in approving this.

PR #1003, opened 2017-11-03
Fix Context object memory leak caused by circular references

kkopachev:
full explanation here: https://github.com/thumbor/thumbor/issues/753#issuecomment-341584293

PR #1041, opened 2018-02-14
Bring back 'headers' context property as it is used by result storages

kkopachev:
Should fix https://github.com/thumbor/thumbor/pull/1003#issuecomment-365550122
@heynemann @bjoernhaeuser

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -38,6 +38,8 @@ class ContextTestCase(TestCase):
         expect(ctx.filters_factory.filter_classes_map).to_be_empty()
         expect(ctx.request_handler).to_be_null()
         expect(ctx.thread_pool).to_be_instance_of(ThreadPool)
+        expect(ctx.headers).to_be_instance_of(dict)
+        expect(ctx.headers).to_be_empty()
 
     def test_can_create_context_with_importer(self):
         cfg = Config()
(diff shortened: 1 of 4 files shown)


bjoernhaeuser:
I just tested this and the error is gone! looks good to me!

heynemann:
Why did you remove the GC bit in the same PR?

kkopachev:
@heynemann that was a piece remain after removing scheduled calls to gc.collect in previous PR https://github.com/thumbor/thumbor/pull/1003/files#diff-96c1f5e9855ad12779e42ade3a0c99e1L152

heynemann:
Ok, LGTM

heynemann:
Please test with master guys! Thanks for the hard work!

bjoernhaeuser:
Tested master and looks good to me!

bjoernhaeuser:
We will do more testing tomorrow. :)

bjoernhaeuser:
Testing this no in production for a week and no new errors. Though I do not see the memory consumption to actually go done.

kkopachev:
@bjoernhaeuser do you have mem consumption graphs before and after patch? Do you use tc_aws for some kind of storage?

bjoernhaeuser:
Sorry, we do not have this data available. 

We are using tc_aws for retrieving source data from s3 and for s3 result storage.

kkopachev:
@bjoernhaeuser I think that botocore is leaving circularly-referenced objects behind. I submitted an issue to tc_aws package to try to reuse same botocore clients over requests https://github.com/thumbor-community/aws/issues/109

PR #1042, opened 2018-02-14
Optimize travis build: static ffmpeg and pyexiv2 from sources

kkopachev:
@marcelometal this fixes skipped pyexiv2 tests and drops sudo requirement.


caeugusmao:
@kkopachev This optimization is related to the time of build running?

PR #1044, opened 2018-02-18
obvious fix for bad fallback variable name

mcallistersean:
I just tripped over this on a fresh install where opencv wasn't available and I was working with tiff files.
Fix is trivial.

supersedes #1043 as that referenced the master, sorry for that


caeugusmao:
@heynemann Could you accept this PR?

PR #1048, opened 2018-02-27
Allow Pillow 5.1

### LINKED ISSUE #1038: Thinking about the future: Pillow 5
Recently, [Pillow 5.0 was released](https://github.com/python-pillow/Pillow/releases/tag/5.0.0) and testing it against Thumbor some tests failed. 

IMHO, in the near future Ubuntu and [Debian](https://packages.debian.org/sid/python-pil) will be released with Pillow 5, it would be great if it was fixed.

kkopachev:
Fixes #1038
@marcelometal @heynemann

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ http://<thumbor-server>/300x200/smart/s.glbimg.com/et/bb/f/original/2011/03/24/V
             "tornado>=4.1.0,<5.0.0",
             "pycryptodome >= 3.4.7",
             "pycurl>=7.19.0,<7.44.0",
-            "Pillow>=3.0.0,<5.0.0",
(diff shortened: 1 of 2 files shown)


caeugusmao:
Hi @kkopachev,

What is the difference that you noted with this change? If we keep the same behavior, we can accept.

tks

kkopachev:
@caeugusmao there are no functional difference as far as I can see.

caeugusmao:
@kkopachev The building is not passing!! Could you verify before we accept?

Many tks!!!

kkopachev:
@caeugusmao this should fix it: https://github.com/thumbor/thumbor/pull/1069

PR #1054, opened 2018-03-22
Graceful shutdown support

Globegitter:
We are planning to use thumbor at Ecosia in a kubernetes cluster. For that we would like to support graceful shutdown on `SIGTERM` to ensure we have no requests that will be left unhandled / cancelled in case of a upgrade, or scale event etc.

I tried to keep the changes minimal, with one change though that the `KeyboardInterrupt` will also be handled by the signal handler. I am happy to revert that to the old behaviour if wanted.

Further there are two new configs `MAX_WAIT_SECONDS_BEFORE_SERVER_SHUTDOWN` and `MAX_WAIT_SECONDS_BEFORE_IO_SHUTDOWN` to configure the graceful shutdown. Per default they are both set to 0, which would basically keep the current shutdown behaviour.

I just added unit tests, which does not full test the signal handling behaviour, so I would also be happy to add integration tests if wanted.

Let me know if that is something you would consider merging.

### DIFF
--- a/thumbor/signal_handler.py
+++ b/thumbor/signal_handler.py
@@ -1,0 +1,53 @@
+import signal
+import sys
+import time
+
+from thumbor.utils import logger
+import tornado.ioloop
+
+
+def signal_handler(server, config, sig, frame):
+    io_loop = tornado.ioloop.IOLoop.instance()
+
+    def stop_loop(now, deadline):
+        if now < deadline and (io_loop._callbacks or io_loop._timeouts):
+            logger.debug('Waiting for next tick')
+            now += 1
+            io_loop.add_timeout(now, stop_loop, now, deadline)
+        else:
+            io_loop.stop()
+            logger.debug('Shutdown finally')
+
+    def stop_server(now, deadline):
+        if now < deadline:
+            logger.debug('Waiting for next tick')
+            now += 1
+            io_loop.add_timeout(now, stop_server, now, deadline)
+        else:
+            server.stop()
+            logger.debug('Stopped http server.')
+            logger.debug('Will shutdown io in maximum %d seconds ...',
+                         config.MAX_WAIT_SECONDS_BEFORE_IO_SHUTDOWN)
+            now = time.time()
+            stop_loop(now, now + config.MAX_WAIT_SECONDS_BEFORE_IO_SHUTDOWN)
+
+    def shutdown():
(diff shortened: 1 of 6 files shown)


thumbor/signal_handler.py
```diff
@@ -0,0 +1,52 @@
...
+    
+    def stop_server(now, deadline):
+        if now < deadline:
+            logger.info('Waiting for next tick')
+            now += 1
+            io_loop.add_timeout(now, stop_server, now, deadline)
+        else:
+            server.stop()
+            logger.info('Stopped http server.')
+            logger.info('Will shutdown in maximum %s seconds ...',
```
heynemann: Shouldn't it be %d?

heynemann:
Your PR is failing the lint tests... You can make lint to ensure your code is compliant, or just check the warnings in Travis.

Globegitter:
@heynemann yeah sorry I had trouble installing all of the dependencies locally so did not run all tests and missed out on linting. Should be done now. In terms of the quality review, can the comments about the methods in the test class just be ignored?

Savar:
just for completeness .. we did a small trick how we ensured it without the change (which I still think is a good idea!!)

```
          lifecycle:
            # thumbor isn't doing a graceful shutdown so we
            # actually wait for 5 seconds to give it enough
            # time to process all the pending requests (it
            # is already out of the LB) before it gets the
            # TERM signal
            preStop:
              exec:
                command:
                - bash
                - -c
                - 'sleep 5'
```

Globegitter:
@Savar Thanks for sharing that here, after digging a bit deeper I found that as well. That does also require one to ship a shell and sleep :) (e.g. when shipping apps in [distroless](https://github.com/GoogleCloudPlatform/distroless))

@heynemann Also fixed the comment now.

heynemann:
One last thing before we merge this: do you think you can squash all the commits into one with a meaningful commit message? It's easier to track commits in thumbor's tree if each change is in one commit. Sorry for the delay. I'm trying to review all the PRs as to either accept or close them and yours seems like a sensible one to accept.

Globegitter:
Used for that to happen on merge but I think that should do it. Happy for this to be accepted - has been running for us in production now for a week :)

Globegitter:
@heynemann Any chance this could get merged now? :)

PR #1058, opened 2018-03-23
allowing separate storage for the watermark filter

savar:
Sometimes thanks to CDNs or other caching mechanisms it makes no sense
to have a storage for the files.

This doesn't necessarily apply to watermarks if there is a default
watermark which needs to be applied to all images it makes fully sense
to cache that one because a CDN in front of Thumbor doesn't help at all.

This change won't change the default behaviour. In case
WATERMARK_STORAGE isn't set it will take the one from STORAGE.


heynemann:
Since this hasn't had activity for a long time I'm closing it. If you still want to work on this please reopen it.

PR #1063, opened 2018-05-08
return a 404 if a request tries to fetch a dir

shabadoo75:
After a successful request, when a request is made to the same path without the filename filestore attempts to load a directory from it's cache and crashes with a 500 error.

e.g.
http://domain.com/unsafe/$2x0/http://origin.com/path/to/file.jpg  ---> 200 success
followed by
http://domain.com/unsafe/$2x0/http://origin.com/path/to/  ---> 500 error

### DIFF
--- a/thumbor/result_storages/file_storage.py
+++ b/thumbor/result_storages/file_storage.py
@@ -13,7 +13,7 @@ from uuid import uuid4
 from shutil import move
 import pytz
 
-from os.path import exists, dirname, join, getmtime, abspath
+from os.path import exists, dirname, join, getmtime, abspath, isdir
 
 from thumbor.engines import BaseEngine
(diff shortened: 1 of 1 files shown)


caeugusmao:
Hi @shabadoo75,

We don't recommend to save images which are requested through not signed urls. The unsafe mode should be used for tests. If you are using unsafe mode and saving the images, Indeed in this case Thumbor should return 404 instead of 500 if you are asking for a nonexistent image.

If you were using signed urls, this error would never happens. Thumbor would return 400, "Malformed url", because the hash would not be compatible with the rest of the URI.

Thanks for your PR. ;)

@heynemann Could you accept this PR?

shabadoo75:
thanks @caeugusmao !

PR #1067, opened 2018-05-13
Fix expiry check in file storage

eax64:
The ``__is_expired()`` function in storage and result storage is wrongly using ``.seconds`` instead of ``.total_seconds()``.
This leads to ``STORAGE_EXPIRATION_SECONDS`` and ``RESULT_STORAGE_EXPIRATION_SECONDS`` settings not to work with value greater than a day and other odd behaviour.

Here is a comparison between ``timedelta.seconds`` and ``timedelta.total_seconds()``:
```python
>>> timestamp = 1526000000
>>> subtract = lambda a, b: datetime.fromtimestamp(a) - datetime.fromtimestamp(b)
>>> values_to_subtract = [0, 60, 60*60, 60*60*24, 60*60*24+10]
>>> [subtract(timestamp, timestamp-x).seconds for x in values_to_subtract]
[0, 60, 3600, 0, 10]
>>> [subtract(timestamp, timestamp-x).total_seconds() for x in values_to_subtract]
[0.0, 60.0, 3600.0, 86400.0, 86410.0]
>>>
```
``timedelta.seconds`` maximum value is 86400, so files will never be considered expired when ``*_EXPIRATION_SECONDS`` settings is set tot be greater or equal to a day.

Using ``timedelta.seconds`` is similar to using ``timedelta.total_seconds() % 86400``, so with an expiry set to 2h, ``__is_expired()`` will return False if the file is 25h old .
```python
>>> 60*60*25 > 60*60*2
True
>>> 60*60*25 % 86400 > 60*60*2
False
```

I've also updated the expiration test from storage and added a similar test for result storage.

### DIFF
--- a/tests/result_storages/test_file_storage.py
+++ b/tests/result_storages/test_file_storage.py
@@ -92,3 +92,26 @@ class ResultStorageResultTestCase(BaseFileStorageTestCase):
         expect(len(result)).to_equal(5319)
         expect(len(result)).to_equal(result.metadata['ContentLength'])
         expect(result.last_modified).to_be_instance_of(datetime)
+
+
+class ExpiredFileStorageTestCase(BaseFileStorageTestCase):
+    def get_config(self):
+        return Config(
+            RESULT_STORAGE_FILE_STORAGE_ROOT_PATH=self.get_fixture_path(),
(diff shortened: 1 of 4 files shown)


heynemann:
Very well done! Thank you!

PR #1070, opened 2018-05-21
Docs: Add adaptive and full fit-in

cristiandean:
Improvements in documentation, added **full-fit-in** and **adaptative-fit-in**.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -7,7 +7,7 @@ retrieving the image and a very similar endpoint to retrieve metadata.
 Image Endpoint
 --------------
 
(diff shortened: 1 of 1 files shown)


Bogatinovski:
I don't understand what does it mean to be full fit in versus adaptive fit in. Currently I can only use the regular fit in in the url, for example: `https://example.com/fit-in/50x50/image.png` But if I try to use full or adaptive fit in, none of the following works:
`https://example.com/full-fit-in/50x50/image.png` 
`https://example.com/adaptive-fit-in/50x50/image.png` 
What am I missing ?

PR #1076, opened 2018-05-25
Allowing the use of variables in the config

caeugusmao:


kkopachev:
Turns out it's not much useful, since filters/optimizers could not be passed through environment. I suspect any non-string types would not work as expected in many cases. :disappointed:

kkopachev:
@heynemann @caeugusmao 
Oh, this breaks things. I use setup similar to [APSL's docker container](https://github.com/APSL/docker-thumbor/pull/94/files) and pass env vars at container start up time. Vars are then used to generate conf file with help of envtpl utility. Then thumbor runs with generated config file.
Changes in this PR are forcing config class to use and prefer environment variables over config. This makes it impossible to use with optimizers and filters array.

jtoy:
im very surprised that thumbor doesnt work with env variables, this makes the system not very secure for running in the cloud.

PR #1081, opened 2018-06-04
Allow Tornado 5.

kkopachev:
This PR fixes few tests - removes magic around HttpClient instantiation in one place (fyi @Savar) and simplifies ThreadPool implementation for async case and adds a test.

Mainly this PR allows to use thumbor with Tornado 5.


kkopachev:
Any comments here? Would be great to release it together with pillow version bump.

PR #1052, opened 2018-03-13
#1051 Adding MIN radius for avoiding Zero Division error.

jinesh90:
Currently, There is a bug , if user send filters:blur(0) , thumbor throws ZeroDivision Error, full trace is here 
================================
```2018-03-13 00:05:07 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/tornado/gen.py", line 307, in wrapper
    yielded = next(result)
  File "/usr/local/lib/python2.7/site-packages/thumbor/transformer.py", line 138, in smart_detect
    self.do_image_operations()
  File "/usr/local/lib/python2.7/site-packages/thumbor/transformer.py", line 231, in do_image_operations
    callback=inner
  File "/usr/local/lib/python2.7/site-packages/thumbor/context.py", line 287, in queue
    self._execute_in_foreground(operation, callback)
  File "/usr/local/lib/python2.7/site-packages/thumbor/context.py", line 275, in _execute_in_foreground
    callback(result)
  File "/usr/local/lib/python2.7/site-packages/thumbor/transformer.py", line 227, in inner
    self.done_callback()
  File "/usr/local/lib/python2.7/site-packages/thumbor/handlers/__init__.py", line 247, in after_transform
    self.filters_runner.apply_filters(thumbor.filters.PHASE_POST_TRANSFORM, finish_callback)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 91, in apply_filters
    exec_one_filter()
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 90, in exec_one_filter
    f.run(exec_one_filter)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 198, in run
    results.append(self.runnable_method(*self.params))
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/__init__.py", line 23, in wrapper
    return fn(self, *args2)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/blur.py", line 43, in blur
    matrix, matrix_size = self.generate_1d_matrix(sigma, radius)
  File "/usr/local/lib/python2.7/site-packages/thumbor/filters/blur.py", line 33, in generate_1d_matrix
    exp = math.e ** -(((adj_x * adj_x)) / two_sigma_squared)
ZeroDivisionError: float division by zero```
==========================================

to avoid this situation, Added MIN radius params.

PR #1082, opened 2018-06-05
Fix blur filter with zero value

cristiandean:
Fix issue when blur is 0 as in #1052, but set no blur when value is 0. This will fix de issue and will have a saner behaviour.

Thank you.


caeugusmao:
@cristiandean The CI build failed. It is lint, could you fix?

caeugusmao:
@cristiandean Could you squash your commits? yks

cristiandean:
Squashed @caeugusmao, thank you.

PR #1084, opened 2018-06-07
Use -trans_color option on gifv instead of mem hungry reverse filter

kkopachev:
'reverse' complex filter has to buffer entire video content in memory
before next filter can chop off last frame. In case of long big gif
inputs, amount of memory could exceed few gigabytes. If thumbor server
at the same time tries to optimize 2 gifv files, it's likely to run out
of memory and be killed with OOM killer. When ffmpeg process killed,
optimizer throws exception, but due to the fact thumbor misuses futures,
exception is never retrieved, so request waits until timeout happens.
There is a high chance that many requests will hang at the same time,
effectively exhausting open files limit and rendering service
interruption.
Reverse filter needed to overcome issue with ffmpeg dropping last frame
duration when used with overlay filter, which, in turn, was used to add
color-background for input gifs with transparency.
Turns out, that ffmpeg's gif decoder can be supplied with [`-trans_color`](https://github.com/FFmpeg/FFmpeg/blob/cdd3048134b8b032b69d3c076d73d24e329dc9ba/libavcodec/gifdec.c#L537)
parameter, which does exactly that, but on decoding stage. This allows
to remove memory-heavy reverse filter and overlay all together.
As a side effect, previously broken ffmpeg 3.3 should also work fine.

see https://github.com/thumbor/thumbor/pull/907

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
 cache:
   pip: true
   apt: true
-sudo: required
 matrix:
   fast_finish: true
   allow_failures:
@@ -27,6 +26,8 @@ env:
   - UNIT_TEST=1
 addons:
   apt:
+    sources:
+    - trusty-media
     packages:
     - libjpeg-progs
     - libimage-exiftool-perl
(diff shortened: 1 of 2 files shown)


kkopachev:
Unfortunately there was no tests checking content of result, but only tests checking command returns appropriate content-type.

PR #1088, opened 2018-06-14
Storage#get should always call the callback

nicolaspanel:


thumbor/result_storages/file_storage.py
```diff
@@ -52,7 +52,7 @@ def get(self, callback):
         file_abspath = self.normalize_path(path)
         if not self.validate_path(file_abspath):
```
nicolaspanel: NB: since `Storage#validate_path` always prepend `self.context.config.RESULT_STORAGE_FILE_STORAGE_ROOT_PATH` this situation cannot append

PR #1089, opened 2018-06-14
Enable environment variables config using a flag

### LINKED ISSUE #1086: BC break caused by enabling env vars as config acceptance
I use setup similar to [APSL's docker container](https://github.com/APSL/docker-thumbor/pull/94/files) and pass env vars at container start up time. Vars are then used to generate conf file with help of envtpl utility. Then thumbor runs with generated config file.
Changes introduced in 6.4.3 are forcing config class to use and *prefer* environment variables over config. This makes it impossible to use with optimizers and filters array as there is no way to pass arrays through env variables.

Was implemented here https://github.com/thumbor/thumbor/pull/1076

kkopachev:
resolves #1086
@caeugusmao @heynemann 
I can't run current release without unsetting env variables.

### DIFF
--- a/tests/fixtures/thumbor_config_server_test.conf
+++ b/tests/fixtures/thumbor_config_server_test.conf
@@ -9,3 +9,4 @@
 # Copyright (c) 2011 globo.com thumbor@googlegroups.com
 
 ALLOWED_SOURCES = ['mydomain.com']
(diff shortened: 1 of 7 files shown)


heynemann:
You are absolutely right. We introduced a breaking change by trying to be more conformant to 12-factor. Really sorry for it.

PR #1105, opened 2018-08-06
Limit highest quality to initial quality

### LINKED ISSUE #1104: Segfault with certain max_bytes sizes when using AutoWebP
### Thumbor request URL
unsafe/512x400/filters:quality(100):max_bytes(20400):fill(white)/https://upload.wikimedia.org/wikipedia/commons/5/5e/G_logo_of_Google.jpg

### Expected behaviour
Should return a WebP image with specified filters/size.

### Actual behaviour
Thumbor crashes with a segfault or memory corruption. Seems to happen when the `max_bytes` size is similar to the actual returned size. In the example above, a `max_bytes` size of 20000 returns a 8.4 KB image and a `max_bytes` size of 20500 returns a 20 KB image. The `max_bytes` size of 20400 crashes Thumbor. 

Above url will only crash Thumbor when `Accept: image/webp` is set.

On Alpine 3.8 Thumbor will segfault with exit code 139. No logs.

On Debian Thumbor will crash with:
`*** Error in `/usr/local/bin/python': double free or corruption (out): 0x00000000023f2b00 ***`

### Operating system
Tried on Alpine3.8 and Debian Jessie

### Your thumbor.conf
Running Thumbor 6.5.1

All default except:
AUTO_WEBP = True
STORAGE = 'thumbor.storages.no_storage'

kkopachev:
if we end up that far and trying to fit image into certain filesize,
that means initial save quality was already too much, so it should be
higher bound.

Fixes #1104 
@heynemann this is simple fix. Submitted just in case [binary search approach](https://github.com/thumbor/thumbor/pull/1106) is no good for some reason.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -507,8 +507,8 @@ class BaseHandler(tornado.web.RequestHandler):
             results = engine.read(extension, quality)
 
         prev_result = results
-        while len(results) <= max_bytes:
-            quality = int(quality * 1.1)
+        while len(results) <= max_bytes and quality < initial_quality:
+            quality = max(initial_quality, int(quality * 1.1))
             logger.debug('Trying to upsize image with quality of %d...' % quality)
             prev_result = results
             results = engine.read(extension, quality)


heynemann:
I think we should go with this since it's a smaller change. I'll give some thought about the binary search way anyway later on.

PR #1116, opened 2018-09-24
Add AutoJPG filter to override AUTO_PNG_TO_JPG config variable

kkopachev:
@heynemann this is follow up on https://github.com/thumbor/thumbor/pull/1083#issuecomment-396716490


caeugusmao:
@kkopachev Tks!! I really appreciated that, we could combine default behavior + opcional filter.

PR #1128, opened 2018-10-29
Allow all Pillow 5.x

kkopachev:
I don't see why not. Pillow is released once a quarter, bumping version constraint each time seems unnecessary. They are not following semver, but they do release breaking changes with new major version.


kkopachev:
Any objections merging this?

caeugusmao:
@kkopachev no objection.

heynemann:
Thanks!

PR #1130, opened 2018-10-30
Fixes #1129 make sure blur filter with PIL engine works for png images with P mode

### LINKED ISSUE #1129: blur doesn't work for png images with mode P using PIL engine
### Original image
image

### Thumbor request URL
http://localhost:8000/filters:blur(10)/image.png

### Expected behaviour
Should see a properly blurred image
image_filters-blur 10

### Actual behaviour
Tell us what happens instead
image_filters_blur 10 _incorrect

### Operating system
OSx

### Your thumbor.conf
```
TC_AWS_LOADER_BUCKET='mybucket'
LOADER='tc_aws.loaders.presigning_loader'
TC_AWS_REGION='us-east-1'
TC_AWS_ENABLE_HTTP_LOADER='True'
RESULT_STORAGE_STORES_UNSAFE='False'
RESULT_STORAGE="thumbor.result_storages.no_storage"
STORES_CRYPTO_KEY_FOR_EACH_IMAGE='False'
TC_AWS_STORE_METADATA='False'
OPTIMIZERS="['thumbor_plugins.optimizers.jp2']"
MAX_AGE='604800'
QUALITY='100'
```

I tried looking into why this is happening
```
>>> from PIL import Image
>>> im = Image.open('/Users/pbagrecha/Downloads/image.png')
>>> from PIL import ImageFilter
>>> im = im.filter(ImageFilter.GaussianBlur(10))
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PIL/Image.py", line 1084, in filter
    return self._new(filter.filter(self.im))
  File "/usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/PIL/ImageFilter.py", line 156, in filter
    return image.gaussian_blur(self.radius)
ValueError: image has wrong mode
>>> im.mode
'P'
```

I realized that thumbor blur filter is not using PIL gaussian filter
from https://github.com/thumbor/thumbor/blob/master/thumbor/filters/blur.py#L48
```
mode, data = self.engine.image_data_as_rgb()
imgdata = _convolution.apply(mode, data, self.engine.size[0], self.engine.size[1], matrix, matrix_size, True)
imgdata = _convolution.apply(mode, imgdata, self.engine.size[0], self.engine.size[1], matrix, 1, True)
```

Now self.engine.image_data_as_rgb() from https://github.com/thumbor/thumbor/blob/master/thumbor/engines/pil.py#L327 does 
```
def image_data_as_rgb(self, update_image=True):
    converted_image = self.image
    if converted_image.mode not in ['RGB', 'RGBA', 'P']:
        if 'A' in converted_image.mode:
            converted_image = converted_image.convert('RGBA')
        else:
            converted_image = converted_image.convert('RGB')
    if update_image:
        self.image = converted_image
    return converted_image.mode, converted_image.tobytes()
```

Notice it doesn't convert image with mode P to RGBA or RGB.

_convolution.apply at https://github.com/thumbor/thumbor/blob/master/thumbor/ext/filters/_convolution.c#L38 expects the image to be in RGB(A) mode.

We should be doing what is being done in resize at https://github.com/thumbor/thumbor/blob/master/thumbor/engines/pil.py#L122. No wonder when I resize the image before applying blur it works.

bagipriyank:


bagipriyank:
@caeugusmao could you please have a look?

PR #1135, opened 2018-11-11
[refactor branch] Blur filter

cristiandean:
@heynemann  what do you think about this PR for the refactor branch? I also separated some points.

Will Thumbor continue to support python 2.6?

The tests were not passing, I needed to remove /v2/ from the beginning of the url of the `__init__.py` files of the detectors and the filters, does it make sense to you? (I made a separate commit if it didn't)

I needed to modify `_convolution.c` so that it could support the mode parameter as bytes as well and not just as string.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -44,7 +44,7 @@ documentation](http://thumbor.readthedocs.org/en/latest/index.html "thumbor docs
 - [x] Full Fit-in
 - [x] Filters
    - [ ] background_color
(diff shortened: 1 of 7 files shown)


heynemann:
Looks great!!! Let’s keep going. 

Yep we are dropping 2.7 support. I think it’s time. What do you think?

cristiandean:
I agree, I also think it's time!

Issue #1039, opened 2018-02-07
Problem with svg


SergioJorge:
### Thumbor request URL
http://<thumbor_url>/unsafe/<path_to_image>/7b9f1630b40b315d.svg

### Expected behaviour
Return a valid svg.

### Actual behaviour
Thumbor return a 500.
`thumbor:ERROR [BaseHandler] get_image failed for urlhttp://<thumbor_url>/unsafe/<path_to_image>/7b9f1630b40b315d.svg. error: float division by zero
`

Cairosvg raise a error when a svg have a mask without width.

This is the file in txt because github does not accept svg files.
[7b9f1630b40b315d.svg.txt](https://github.com/thumbor/thumbor/files/1704261/7b9f1630b40b315d.svg.txt)

### Operating system
OSx, Ubuntu, etc

### Your thumbor.conf
Default.

HC-Team:
Hey have exact the same problem, svg is valid but thumbor hits the floor.

HC-Team:
By the way, installing cairosvg ``pip3 install cairosvg`` alongside ``cair`` && ``libffi`` did NOT help, also not installing thumbor after that.

kkopachev:
works for me with thumbor==6.4.1 and CairoSVG==1.0.22

jmif:
Having the same issue with thumbor==6.4.2 and CairoSVG==1.0.22, my SVG has a mask with without a width as well.

Was there any resolution or workaround found for this @SergioJorge?

Tenzer:
In my experience, in order to get SVG files to work, you need to make sure both the `CairoSVG` and `pycairo` packages are installed. However, if you are on Python 2 you want to makes sure you get `CairoSVG 1.x` (1.0.22 being the latest at the time of writing) as the 2.x versions don't have support for Python 2.

However, CairoSVG in version 1.0.22 has a bug which is fixed in the 2.x versions but hasn't been backported. It was reported [here](https://github.com/Kozea/CairoSVG/issues/86) and the fix is [here](https://github.com/Kozea/CairoSVG/commit/b73b3f1a98382038b573151985a77d40ef13c531).

I guess we need to get Python 3 support for Thumbor fixed before we can hope for this to be solved.

hrhitesh:
I have the same issues that I can not upload the SVG file using the curl command. 
Here is my CURL:
curl -v -i -XPOST https://myupload.example.com/image -F "media=@test.svg" and it gives me the error 
[...]

PR #1141, opened 2018-11-23
Python 3 fixes

Tenzer:
Note: This PR is still WIP, I'm just making use of the Travis CI to run tests across multiple Python versions since I have problems running them locally. I hope that's okay.

I'm intending to try and move the support for Python 3 forwards as there currently are [issues blocked](https://github.com/thumbor/thumbor/issues/1039#issuecomment-440760670) by the lack of Python 3 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,14 +33,15 @@ env:
 addons:
   apt:
     packages:
-    - libjpeg-progs
-    - libimage-exiftool-perl
+    - ffmpeg
     - gifsicle
-    - python-all-dev
-    - scons
     - libboost-python-dev
     - libexiv2-dev
-    - ffmpeg
+    - libgnutls-dev
+    - libimage-exiftool-perl
+    - libjpeg-progs
+    - python-all-dev
+    - scons
 install:
   - pip install --upgrade pip
   - pip install -I https://github.com/escaped/pyexiv2/archive/69dd6448f9831bd826137b7519f9d797b23ab4ec.zip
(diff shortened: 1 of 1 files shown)


Tenzer:
My first stumbling block is that on Python 3 the unit tests seg faults at the "tests.filters.test_background_color.BackgroundColorFilterTestCase" test. I have drilled this down to be caused by these lines of code:
https://github.com/thumbor/thumbor/blob/5b6e89c6a3154d409566466c2c0f11450726e549/thumbor/engines/pil.py#L375-L377

It calls code inside https://github.com/thumbor/thumbor/blob/master/thumbor/ext/filters/_composite.c, but here I come a bit short as my C coding skills are mostly non-existing, and especially with regards to Python bindings. If anybody with more knowledge in this area is able to look into it, that would be appreciated!

heynemann:
Take a look at the refactor branch. It is python 3.7 and we have fixed some of the filters there.

Tenzer:
Thanks for the pointer to that! Is there anywhere where the current status of that refactor can be tracked? Is the code in a state where it would be useful with some feedback from some production traffic hitting it?

heynemann:
We are still missing some features that would allow for a real world usage. But as soon as we hit that we'll definitely let the community know. I plan on having a RC version that we can try as soon as possible.

rht:
> We are still missing some features that would allow for a real world usage.

Is there a checklist of this? I can lend a hand for some of them.

cristiandean:
@rht Yes, you can see here:
https://github.com/thumbor/thumbor/blob/refactor/README.mkd

PR #1148, opened 2018-12-09
Fixed i386 build

marcelometal:
Tests failed on i386 architecture, this PR fix it.

`TypeError: longs are not supported for this option`


heynemann:
Thank you very much @marcelometal

PR #1149, opened 2018-12-13
Use SHA1-hashed paths in result storage

mpdude:
As outlined in #1146, source URLs passed to Thumbor can be very long. As the `file_storage` result storage handler does an almost 1:1 mapping of URLs to filesystem paths, this can lead to `IOError: [Errno 36] File name too long` down the road.

This PR suggests changing the file organization strategy for the result cache, using SHA1-based filenames. This also removes some quirks from the past:

* Now works similar to `storages/file_storage.py`
* Until now, `http://` was stripped from URLs before caching, but `https://` was not
* `v2` in cache paths was a quirk introduced in the past

~~Yes, this invalidates existing (warm) caches.~~

*Updated* 

Whenever the result cache does not find an entry at the "new" location, it will check the old path as well. If it finds a result there, it migrates the file to the new location. This should make a rolling migration possible so that users can keep their warm caches.

*Side note*

I am completely new to Python and did not manage to run tests locally. So, could someone please leave a short note here with the minimal steps necessary to setup a test environment locally? https://github.com/thumbor/thumbor/wiki/Hacking-on-thumbor was no help, as my `virtualenv` seems to be missing modules (`ImportError: No module named mock`).


heynemann:
Hey @mpdude thanks for your contrib. We'll try to look at your PR soon.

kkopachev:
2 things to consider:
* This should have more nesting levels. Having too many files in one dir could create some issues, like reading directory content could take way too long.
* Might be harder to find all saved cuts of specific image when take down request comes in. Maybe it's fine

and BC break.

mpdude:
@kkopachev What is that takedown request? By which method can you purge files from the result cache?

kkopachev:
Sometimes copyright owner for a photo comes in and politely asks to remove the photo in question from your site. That's take down request. It's not sufficient to remove only original, because all cuts still remain in result cache on thumbor and thumbor links will still return a cut of the image.

I guess it's not so big an issue after all to find out what file maps to what requested url, just have to consider BC break in that regard.

Actual removal could be simple `rm`.


Bigger concern is 2-level structure could be insufficient for a lot of files.

mpdude:
I’ll change this to create two directory levels based on the first four characters of the hash. That’s up to 65536 directories. AFAIK modern filesystems should be able to handle the order of 10K files per directory without problems. 

Do you think it would be worth the performance hit (?) to try to extract the host name from the source image URL and put that into the path as well? That could simplify removal of images from the cache by host... 🤔

cristiandean:
> I’ll change this to create two directory levels based on the first four characters of the hash. That’s up to 65536 directories. AFAIK modern filesystems should be able to handle the order of magnitude of 10K files per directory without problems.
> 
> Do you think it would be worth the performance hit (?) to try to extract the host name from the source image URL and put that into the path as well? That could simplify removal of images from the cache by host... 🤔

I think that wouldn't be the best approach, because many thumbor users use a single host to serve the images.

kkopachev:
@mpdude I think you are right. I did some searching and it seems performance is not as big of a problem now as it used to. Current proposed state with 2 chars per subdir and 2 levels should be enough. 

You can ignore my comment about takedown requests too. It's mostly just to provide a evidence for BC break. It is going to change directory structure anyway.

Extracting a host does not add any value, in my opinion, since it's not mandatory to have a host. For example when using S3 storage, only object key is in the request uri.

tl;dr: Please consider BC break.

mpdude:
Updated the cache to use two levels of directories.

mpdude:
@caeugusmao That’s great news!

As I said before, this is my first bit of Python code. So, do you have an example for me where else such a config and default setting exists? 

Thanks!

kkopachev:
Instead of doing it behind a flag and carrying a lot of if/else statements throughout the code, you might want to create another storage and keep old one intact. Then user can opt-in to using hash-storage when needed.

mpdude:
But that would mean to carry along a lot of copied/duplicated code in both versions of the class.

caeugusmao:
@kkopachev/@mpdude, 

I understand your point, but with hash or not, still continues being a file storage.

Maybe @mpdude can extend file_storage and implement `normalize_path` and `partition` methods, but I am not sure about that, because It still continues being a file storage. It is not a new kind of storage, is just a new directories organization using filer.

I don't know if we would have many if/else statements in this case. If we are able to implement with few if/else statements I would tend to have a conf.

mpdude:
Regarding BC, what is important to you - the exact way of how the cache is/was organized, or that you don’t want to possibly lose tens of thousands of converted images in some warm cache?

caeugusmao:
@mpdude 

> lose tens of thousands of converted images in some warm cache?

This is the main point for us. 

* To think
You can use https://pypi.org/project/mmh3/ instead of SHA1. Murmur is faster than SHA1 but with higher risk of collision than SHA1. Even this way, It is still low.

We have other features implemented using opt-in conf, for example AUTO_WEBP, you can take a look into Thumbor's code.

mpdude:
The downside with an opt-in flag is that there is no good way to urge users to make the upgrade or activate the flag even if they start from scratch. And those with big warm caches will keep the old setting as long as it is supported.

So, do you think it would be possible to switch to the new approach, but in case we have a cache miss, check the old location where the file would previously have been kept?

If we find it there, move it to the new location.

That way we could get a gradual migration. Old users keep their caches with a minimal performance impact on first hit for the old location. New users need do nothing. 

In some distant future, drop that compatibility layer if we deem that enough time for a „rolling“ migration has passed.

mpdude:
I've added a lazy migration so that cache entries will be moved from the old to the new location on first hit. That way, no need to configure anything – just give caches some time to migrate. Works for me 👍.

Still have to see if tests pass.

What's your policy regarding CHANGELOG entries? Just add one on top of the file? Or will you update the log upon release?

mpdude:
@caeugusmao What do you think – is this a feasible way of doing a smooth migration?

caeugusmao:
@mpdude 

I think is a good approach, but even this way, maybe we add a flag with default value True. This way I think we can preserve someone who does not want to migrate for while. 

Later in a BC version, we can remove this flag.

PR #1151, opened 2019-01-04
Fix the bug that put method is not asynchronous which will stuck the thumbor server

Tideorz:
Recently, found that the put method doesn't have tornado.coroutines, which will affect the performance 
of thumbor when some put requests getting into thumbor. And delete api also doesn't have yield, here's
a simple fix for this, each method in put/delete add yield to let them to be responded asynchronously.

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -747,6 +747,7 @@ class ImageApiHandler(ContextHandler):
             return False
         return True
 
+    @gen.coroutine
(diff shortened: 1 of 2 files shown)


heynemann:
Thank you very much for your contribution!

Tideorz:
@heynemann, could you please add a new tag for this bug fix, I see on the pypi, the latest version 6.7.0
doesn't include this changes.

PR #1159, opened 2019-01-24
Filter fill with blur

andrejguran:
# New parameter for filter fill - blur
inspired by this article: https://blog.imgix.com/2018/10/01/new-blur-parameter

a new accepted parameter for filter **Fill**: `fill(blur)`

http://localhost:8888/unsafe/fit-in/300x300/filters:fill(blur)/https://github.com/thumbor/thumbor/wiki/tom_before_brightness.jpg

## Almost identical output:
imgix: 
1548337166-8edf7d1e778848f927934c54ec5bae34
thumbor:
tom_fill_blur

## Works nicely also with transparent images
dice_transparent_background

## And animated gifs
ajax-loader


docs/filling.rst
```diff
@@ -23,6 +23,8 @@ Arguments
    If color is "auto", a color is smartly chosen (based on the image pixels)
    as the filling color.
 
+   If color is "blur", the missing parts are filled with blured original image.
```
scorphus: `s/blured/blurred/`

andrejguran:
> Nice one, @andrejguran! 👍 There's just a small typo that needs to be fixed. Please check my line comment.

typo fixed, good eye ;)

caeugusmao:
Tks @andrejguran.

thumbor/engines/pil.py
```diff
@@ -388,3 +388,6 @@ def strip_icc(self):
 
     def strip_exif(self):
         self.exif = None
+
+    def blur(self, radius):
+        self.image = self.image.filter(ImageFilter.GaussianBlur(radius))
```
kkopachev: This method has to be defined in parent class, otherwise it's unusable with non-pillow engines.
I am not sure what we could do about BC break in this case. 
Maybe we could use blur from filters? It works on raw RGB data.

andrejguran:
> Ouch, my bad for overlooking it. Would be great to have some test cases to catch that.
> 
> @andrejguran would you like and be willing to implement it for other engines as well?

yes sure. Can you point me, where the other engines are located as this repo contains only pillow, json, and gif engines?
Also for which of them does it make sense to implement the blur? I suppose imagemagic is fairly popular, what others?

kkopachev:
Maybe it would be easier to reuse blur method from blur filter? It's C-level implementation so it's engine-independent. That might be just one universal way of fixing it.

Defining new public method on base engine is BC break, I think, even if somebody will fix other known engines. There might be custom engines used nobody knows about.

andrejguran:
> Defining new public method on base engine is BC break, I think, even if somebody will fix other known engines. There might be custom engines used nobody knows about.

It's BC break only if they start using the new function. Otherwise, the new version will be still backward compatible. 
Not sure about C implementation of a blur, but maybe we can return black background for engines that didn't yet implement blur, so it won't be a breaking change.

heynemann:
@andrejguran thumbor has a blur filter (https://github.com/thumbor/thumbor/blob/master/thumbor/filters/blur.py), but we use the C convolution to do it. As @kkopachev said we already have the feature implemented and it's engine agnostic. That way this is not a BC issue.

We try hard not to break other engines people might be using in new thumbor releases. Out of the top of my head, I remember Wikipedia uses their own engine. And they have an engine for each file type. If we keep this PR going we'll be breaking their engines if they decide at any point to use this filter. And it will be a very hard issue to track.

Do you think you can open a new PR that uses the blur the same way the blur filter does? You can even import the blur in the blur filter.

scorphus:
@andrejguran I may help you with coming up with a fix. I'll be happy answering your questions. Once a new PR is created, even if it's a WIP, we start discussing it there.

caeugusmao:
@heynemann @kkopachev @scorphus @andrejguran 

Sorry guys, I didn't realize this change were a BC. It makes sense what @kkopachev and @heynemann said. I can revert this commit while this fix is being developed. What do you think?

scorphus:
Don't worry, @caeugusmao, I also missed that bit. I don't think it's really necessary to revert it though as it hasn't been released yet.

andrejguran:
> Hey, @andrejguran! Still on it? 😃

Hi, yeah. To sum up just change implementation of the blur through `thumbor/filters/blur.py` ?

scorphus:
Awesome! Yeah, that's pretty much it. Then, move the one `blur` from `thumbor/engines/pil.py` to `thumbor/engines/__init__.py` so it works no matter what engine is used.

scorphus:
New PR on its way.

PR #1167, opened 2019-02-08
Replace Raven with new Sentry SDK

tjb346:
Sentry says they are [phasing out Raven](https://github.com/getsentry/raven-python) for a new unified [SDK](https://github.com/getsentry/sentry-python). We wanted to start with the new SDK for our Thumbor server so we created the new error handler used in this PR. For our use it seems to provide more data and the request data better fits what Sentry expects as it is properly formatted in the Sentry portal.

Existing tests have been modified for the new SDK.


heynemann:
This looks good! Can you please rebase with master?

thumbor/error_handlers/sentry.py
```diff
@@ -1,87 +1,53 @@
...
-        import raven
+    def __init__(self, config):
+        self.config = config
 
         dsn = config.SENTRY_DSN_URL
+
         if not dsn:
             raise RuntimeError(
-                "If you set USE_CUSTOM_ERROR_HANDLING to True, and you are using thumbor.error_handlers.sentry, " +
+                "If you set USE_CUSTOM_ERROR_HANDLING to True, and you are using thumbor.error_handlers.sentry_python, " +
```
kkopachev: I think that has to stay as it was.

tjb346: Yes, you're right. I've pushed a commit to change it back.

tjb346:
There was a change in sentry-python version 0.8.0 that stopped adding quotes around the  "extra" dictionary string values. This was breaking the error handler test, so I added a commit to remove the expectation of quotes in the test values. It's actually more like to how it worked previously.

Issue #1156, opened 2019-01-21
test failure in test_redeye_applied (numpy >= 1.16.0)


marcelometal:
Running thumbor's test over `numpy >= 1.16.0`, it fails (numpy version is not locked in setup.py). Travis uses numpy version `1.13.3`.

```
======================================================================
ERROR: test_redeye_applied (tests.filters.test_redeye.RedEyeFilterTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/work/stemci01/debian/build/thumbor/thumbor-6.5.1/.pybuild/cpython2_2.7_thumbor/build/tests/filters/test_redeye.py", line 43, in test_redeye_applied
    config_context
  File "/work/stemci01/debian/build/thumbor/thumbor-6.5.1/.pybuild/cpython2_2.7_thumbor/build/tests/base.py", line 279, in get_filtered
    fltr.run(dummy_callback)
  File "thumbor/filters/__init__.py", line 202, in run
    results.append(self.runnable_method(*self.params))
  File "thumbor/filters/__init__.py", line 23, in wrapper
    return fn(self, *args2)
  File "thumbor/filters/redeye.py", line 38, in red_eye
    image.setflags(write=1)
ValueError: cannot set WRITEABLE flag to True of this array
-------------------- >> begin captured logging << --------------------
PIL.PngImagePlugin: DEBUG: STREAM 'IHDR' 16 13
PIL.PngImagePlugin: DEBUG: STREAM 'IDAT' 41 358237
PIL.PngImagePlugin: DEBUG: STREAM 'IHDR' 16 13
PIL.PngImagePlugin: DEBUG: STREAM 'IDAT' 41 65536
PIL.PngImagePlugin: DEBUG: STREAM 'IHDR' 16 13
PIL.PngImagePlugin: DEBUG: STREAM 'IDAT' 41 358944
--------------------- >> end captured logging << ---------------------```

cristiandean:
Thank you @marcelometal! I ran the tests with version 1.15.4 and it seems all right! #1171 

https://docs.scipy.org/doc/numpy/release.html#arrays-based-off-readonly-buffers-cannot-be-set-writeable

PR #1171, opened 2019-02-13
Fix numpy >= 1.16.0 array not writable

cristiandean:
#1156


marcelometal:
Hi @cristiandean , thanks!

But, why not fix it for new versions instead of fixing for old versions?

cristiandean:
Thanks for the comment @marcelometal. Since after version 1.16.0 of numpy we can not modify the array, so the workaround is to create a copy of it.

rafaeleyng:
Looks good to me

PR #1172, opened 2019-02-13
Fix red-eyes slice indices

cristiandean:
Some urls for the red-eye filter cause the slice indices to float. Causing the following error:

```
2019-02-12 23:26:44 tornado.application:ERROR Future exception was never retrieved: Traceback (most recent call last):
....
  File "/Users/cristiandean/projects/thumbor/thumbor/filters/redeye.py", line 44, in red_eye
    face_image = image[face_y:(face_y + face.height), face_x:(face_x + face.width)]
```

The exception can be reproduced by accessing the following url:

http://localhost:8888/unsafe/debug/fit-in/500x500/smart/filters:red_eye()/https://peladin.files.wordpress.com/2008/01/01.jpg

### DIFF
--- a/tests/test_point.py
+++ b/tests/test_point.py
@@ -33,17 +33,20 @@ class FocalPointTestCase(TestCase):
     def test_new_point_weighted(self):
         point = FocalPoint(x=10, y=20, height=1.0, width=3.0, weight=3.0)
         expect(point.weight).to_equal(3.0)
-        expect(str(point)).to_equal('FocalPoint(x: 10, y: 20, width: 3, height: 1, weight: 3, origin: alignment)')
+        expect(str(point)).to_equal(
+            'FocalPoint(x: 10, y: 20, width: 3, height: 1, weight: 3, origin: alignment)')
 
     def test_new_point_from_dict(self):
-        point = FocalPoint.from_dict({'x': 10.1, 'y': 20.1, 'z': 5.1})
-        expect(point.x).to_equal(10.1)
-        expect(point.y).to_equal(20.1)
-        expect(point.weight).to_equal(5.1)
+        point = FocalPoint.from_dict({'x': 10, 'y': 20, 'z': 5})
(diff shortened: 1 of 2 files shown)


kkopachev:
How come is this happening? Is face width/height float to begin with?

cristiandean:
image
Yes **face.width** and **face.height** are comming as float.

kkopachev:
@cristiandean Maybe it's better to cast them to int in place they are created? Is there a use case when face width could be fractions of pixel? @heynemann

cristiandean:
@kkopachev After checking where the points are created, I was able to solve the problem by doing the following changes:

`/thumbor/point.py`
**before:**
```
    @classmethod
    def from_dict(cls, values):
        return cls(
            x=float(values['x']),
            y=float(values['y']),
            weight=float(values['z']),
            width=float(values.get('width', 1)),
            height=float(values.get('height', 1)),
            origin=values.get('origin', 'alignment')
        )
```

**after:**
```
    @classmethod
    def from_dict(cls, values):
        return cls(
            x=int(values['x']),
            y=int(values['y']),
            weight=int(values['z']),
            width=int(values.get('width')),
            height=int(values.get('height')),
            origin=values.get('origin', 'alignment')
        )

```

Although I think maybe it's not the right way, there seems to be a reason for the focal points to be float. At least that's what some tests show:

image


-----

@heynemann, do you know if there is any reason for focal points to be floating values?

heynemann:
Well, the reason is they are centers of mass in the case of faces. Imagine there are several faces with different sizes and different centers. When you calculate the center of mass of those, it will be float. I don't think we lose anything by converting to int, since as you both noted, these are 2d pixel coordinates, right?

kkopachev:
I vote for casting to int. We can round too, but I think it’s overkill

cristiandean:
@kkopachev Fixed! 

It's necessary to apply this PR #1171 to test this solution.

Issue #1175, opened 2019-02-21
Please add tests to the sdist


jayvdb:
Please add tests to the sdist so they can be used for testing rpms, etc

PR #1179, opened 2019-03-10
MANIFEST.in: Add LICENSE and tests

jayvdb:
Fixes https://github.com/thumbor/thumbor/issues/1175

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,13 @@ recursive-include thumbor *.py
 recursive-include thumbor *.conf
 recursive-include thumbor *.xml
 recursive-include thumbor *.h
+
+include LICENSE README.mkd CHANGELOG
+recursive-include docs *
+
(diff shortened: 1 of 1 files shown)


heynemann:
I'm not sure we should include all tests in the manifest. What's the use case this enables? Would there be an easy way for people to run tests locally? I believe our tests have loads of dependencies right now.

jayvdb:
This enables building rpms from the sdist with tests run in the `%check` to verify all is well on the build environment.

e.g. https://build.opensuse.org/package/show/home:jayvdb:django/python-thumbor

While this might seem like a lot of dependencies to you, it is about average, comparable to something like https://build.opensuse.org/package/show/home:jayvdb:branches:devel:languages:python:numeric/opensesame which drags in half of all X libs and half of all python libraries.

heynemann:
Thanks for the clarification! :)

PR #1193, opened 2019-04-17
this should do it

okor:
So, we actually didn't have orientation enabled so this shouldn't even be running to begin with. Why it was is a bit of a mystery. But this short circuits EXIF retrieval/parsing if `RESPECT_ORIENTATION` isn't set to true. In our case, it's set to false. Which I didn't realize was true but it is. 

I've test this with the wackest EXIF data and of course, no memory explosion. Image is returned perfectly.


okor:
oops, should have been an internal PR

PR #1198, opened 2019-06-08
Remove unused Url, UrlSigner and old Crypto

kkopachev:
This is basically reverting https://github.com/thumbor/thumbor/commit/8af541279c5fd9b9928624b6e68a15529d1c112c but also adding fallbacks for BC purposes. 
Also removing old Crypto, which has not being used for few years now despite existance of `--old` option in `url_composer`


kkopachev:
any news here?

PR #1205, opened 2019-06-18
Prevent healthcheck from being cached

guilhermearaujo:
We run thumbor behind Cloudfront to cache the output images, without a direct, open route to the thumbor instances.

To monitor the application, we'd like to ping the health check route periodically, and since the only way for external tools to reach the server is going through Cloudfront, the health check route must no be cached.

This PR adds a cache control header with the `no-cache` policy to prevent Cloudfront (or whatever other service that may cache the result) from caching the response of the health check route.

### DIFF
--- a/tests/handlers/test_healthcheck.py
+++ b/tests/handlers/test_healthcheck.py
@@ -21,10 +21,12 @@ class HealthcheckHandlerTestCase(TestCase):
         response = self.fetch('/healthcheck')
         expect(response.code).to_equal(200)
         expect(response.body).to_equal("WORKING")
+        expect(response.headers.get('Cache-Control')).to_equal('no-cache')
 
     def test_can_head_healthcheck(self):
         response = self.fetch('/healthcheck', method='HEAD')
         expect(response.code).to_equal(200)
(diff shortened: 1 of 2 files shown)


kkopachev:
Curious how you made your thumbor only respond to cloudfront. In order for CF to talk to your origin, it has to be publicly accessible.

guilhermearaujo:
Using public load balancers and nginx, that re-route the requests to our applications in a private network.

But back to the PR, I can't understand why Travis is failing to compile `pycurl` 🤔 
Even the master branch seems to be failing. And ideas?

kkopachev:
I think load balancers should check for service health and remove unhealthy service from pool. Doing it on cloudfront level seems weird. https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/

kkopachev:
In terms of travis failure. It seems that travis is randomly deciding to run a test on ubuntu xenial instead of default trusty. I made a PR to fix it #1207

cristiandean:
Thank you @guilhermearaujo and @kkopachev!

PR #1210, opened 2019-07-17
Replace piexif with custom orientation-only exif reader/writer

### LINKED ISSUE #1204: EXIF parsing causing memory explosion
# Thumbor request URL
The request url for the issue you are having, you can swap the host name with a fake
`http://thumbor-host.com/unsafe/1400x0/filters:no_upscale()/file-host.com/GettyImages_539746540.jpg`

Here is the image with corrupted/invalid EXIF data used in the example request
[GettyImages_539746540.jpg.zip](https://github.com/thumbor/thumbor/files/3299001/GettyImages_539746540.jpg.zip)

### Expected behaviour
**Tell us what should happen**
An image should be returned

### Actual behaviour
**Tell us what happens instead**
No response is received, thumbor crashes due to running out of memory. Reverse proxy in front of thumbor may respond with a timeout code if that happens first.

### Operating system
**OSx, Ubuntu, etc**
Ubuntu

### Your thumbor.conf
```
## IMAGE PROCESSING ##
ENGINE = 'thumbor.engines.pil'

DETECTORS = [
    'thumbor.detectors.face_detector',
    'thumbor.detectors.feature_detector'
]

FILTERS = [
    'thumbor.filters.blur',
    'thumbor.filters.colorize',
    'thumbor.filters.extract_focal',
    'thumbor.filters.format',
    'thumbor.filters.focal',
    'thumbor.filters.no_upscale',
    'thumbor.filters.quality',
    'thumbor.filters.saturation',
    'thumbor.filters.fill',
]

OPTIMIZERS = [
    'thumbor.optimizers.jpegtran',
    'thumbor.optimizers.gifv',
    'thumbor_plugins.optimizers.autojpeg',
]
JPEGTRAN_PATH = '/usr/bin/jpegtran'
FFMPEG_PATH = '/usr/bin/ffmpeg'

ALLOW_ANIMATED_GIFS = True
USE_GIFSICLE_ENGINE = True
PRESERVE_EXIF_INFO = False
AUTO_WEBP = True

QUALITY = 80
WEBP_QUALITY = 80

AUTOJPEG_QUALITY='90'
AUTOJPEG_SUBSAMPLING='0'

GC_INTERVAL=60

ENGINE_THREADPOOL_SIZE=12
```

---

**I already know what is _causing_ the issue.** As documented here https://github.com/hMatoba/Piexif/issues/90 the piexif library has a bug which is causing the memory explosion. Both `piexif` and the older `pexif` libraries seem to have this bug.

We aren't even using `RESPECT_ORIENTATION` with our cluster but I found that thumbor tries to parse the EXIF data regardless of that setting - but it won't be used unless the setting is enabled. It's probably a _bad idea_ to parse the EXIF data if it's not necessary - it takes time to parse that data. But it's also not a good solution to disable EXIF parsing as a means of working around this problem ([though that is exactly what we have done as an interim patch](https://github.com/voxmedia/thumbor/pull/5/commits/7c4cb4c46be6ab71d150907bbe0c3b164fa9e786#diff-10f8610260b008de1dc1cd6b5df3d3cfR235))

I am also a bit perplexed as to why we are even using the `piexif` library at all. The `pillow` library can very easily return the EXIF data. The example below doesn't have orientation data but it would be in position 274 if it existed.
```
im = Image.open("GettyImages_539746540.jpg")
im._getexif()

/usr/local/lib/python2.7/site-packages/PIL/TiffImagePlugin.py:768: UserWarning: Possibly corrupt EXIF data.  Expecting to read 34225520648 bytes but only got 104. Skipping tag 33437
  " Skipping tag %s" % (size, len(data), tag))
/usr/local/lib/python2.7/site-packages/PIL/TiffImagePlugin.py:768: UserWarning: Possibly corrupt EXIF data.  Expecting to read 33685506 bytes but only got 0. Skipping tag 34850
  " Skipping tag %s" % (size, len(data), tag))
{36864: '0221', 37377: (9965784, 1000000), 37378: (4970854, 1000000), 36867: u'2008:02:21 14:18:14', 36868: u'2008:02:21 14:18:14', 37381: (3, 1), 41990: 0, 37383: 6, 37385: 16, 37386: (300, 1), 41986: 0, 270: u'Austin, TX February 21, 2008: Supporters of candidates Hillary Clinton and Barack Obama line up outside the Rec Sports Center at the Univeristy of Texas at Austin hours prior to the debate between the Democratic candidates Thursday.', 271: u'Canon', 272: u'Canon EOS 5D', 41987: 0, 33432: u'Bob Daemmrich Photography, Inc.', 37380: (-1, 3), 282: (300, 1), 283: (300, 1), 33434: (1, 1000), 34855: 160, 296: 2, 306: u'2008:02:21 14:46:13', 315: u'Bob Daemmrich', 41985: 0, 41486: (4368000, 1415), 41487: (2912000, 942), 41488: 2, 34665: 470}
```
^ As you can see, it encounters the same bad EXIF data but it doesn't OOM any servers which is a nice feature : D

** I think we should
1. Use pillow to extract EXIF orientation data (it's no more cryptic than piexif)
2. Remove piexif as a dependency
3. Stop parsing EXIF data unless it might be used in some later operation (reorientation)

^ I'm happy to code this up if folks feel like that is a good solution. I'd like to know what people think.

kkopachev:
Fixes #1204  
Closes #1209
(please see both for discussions)

Replacing piexif dependency with simple orientation-only reader. It ignores all other exif tags and does not traverse through IFDs other then IFD0 where orientation supposed to be.

### DIFF
--- a/thumbor/engines/extensions/exif_orientation_editor.py
+++ b/thumbor/engines/extensions/exif_orientation_editor.py
@@ -1,0 +1,75 @@
+import struct
+from io import BytesIO
+
+PREFIXES = [
+    b"MM\x00\x2A",  # Valid TIFF header with big-endian byte order
+    b"II\x2A\x00",  # Valid TIFF header with little-endian byte order
+    b"MM\x2A\x00",  # Invalid TIFF header, assume big-endian
+    b"II\x00\x2A",  # Invalid TIFF header, assume little-endian
+]
+
+EXIF_HEADER = b"Exif\x00\x00"
+
+
+class ExifOrientationEditor():
+    _endian = "<"
+    _offset = None
+
+    def __init__(self, data):
+        if data[:6] != EXIF_HEADER:
+            raise SyntaxError("not a TIFF file (header %r not valid)" % data[:4])
+
+        # Skip 6 bytes of exif header to simplify seeks in ImageFileDirectory
+        self.exif_buffer = BytesIO(data[6:])
+
+        header = self._read_header()
+        self._find_orientation_offset(header)
+
+    def _read_header(self):
+        header = self.exif_buffer.read(8)
+
+        if header[:4] not in PREFIXES:
+            raise SyntaxError("not a TIFF file (header %r not valid)" % header)
+        prefix = header[:2]
+        if prefix == b"MM":
+            self._endian = ">"
+        elif prefix == b"II":
+            self._endian = "<"
+        else:
+            raise SyntaxError("not a TIFF IFD")
+
+        return header
+
+    def _find_orientation_offset(self, header):
+        ifd_offset, = self._unpack("L", header[4:])
+        self.exif_buffer.seek(ifd_offset)
+
+        # Read tag directory
+        for _ in range(self._unpack("H", self.exif_buffer.read(2))[0]):
+            # Each tag is 12 bytes. HHL4s = tag, type, count, data
+            # Read tag and ignore the rest
+            tag, = self._unpack("H10x", self.exif_buffer.read(12))
+            if tag == 0x0112:  # Orientation tag
+                self._offset = self.exif_buffer.tell() - 4  # Back 4 bytes to the start of data
+                break
+
+    def _unpack(self, fmt, data):
+        return struct.unpack(self._endian + fmt, data)
+
+    def get_orientation(self):
+        if self._offset is None:
+            return None
+
+        self.exif_buffer.seek(self._offset)
+        return self._unpack('H', self.exif_buffer.read(2))[0]
+
+    def set_orientation(self, value):
+        if self._offset is None:
+            return
+
+        self.exif_buffer.seek(self._offset)
+        self.exif_buffer.write(struct.pack(self._endian + 'H', value))
+
+    def tobytes(self):
+        self.exif_buffer.seek(0)
+        return EXIF_HEADER + self.exif_buffer.read()
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,6 @@ http://<thumbor-server>/300x200/smart/s.glbimg.com/et/bb/f/original/2011/03/24/V
             "pycurl>=7.19.0,<7.44.0",
             "Pillow>=4.3.0,<6.0.0",
             "derpconf>=0.2.0",
-            "piexif>=1.0.13,<2.0.0",
             "statsd>=3.0.1",
             "libthumbor>=1.3.2",
             "futures",
--- a/tests/engines/test_base_engine.py
+++ b/tests/engines/test_base_engine.py
@@ -12,7 +12,6 @@ from __future__ import unicode_literals, absolute_import
 from struct import pack
 from os.path import abspath, join, dirname
 
-import piexif
 from unittest import TestCase
 from xml.etree.ElementTree import ParseError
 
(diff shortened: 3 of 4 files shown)


thumbor/engines/extensions/exif_orientation_editor.py
```diff
@@ -0,0 +1,66 @@
...
+
+class ExifOrientationEditor():
+    _endian = "<"
+    _offset = None
+
+    def __init__(self, data):
+        if data[:6] != EXIF_HEADER:
+            raise SyntaxError("not a TIFF file (header %r not valid)" % data[:4])
+
+        self.fp = BytesIO(data[6:])
```
okor: What is `fp`?

thumbor/engines/extensions/exif_orientation_editor.py
```diff
@@ -0,0 +1,66 @@
...
+        ifh = self.fp.read(8)
```
okor: What does `ifh` mean?

okor: It might be helpful to comment why this is only reading out 8 bytes. May be self documenting if `ifh` had a more descriptive name?

thumbor/engines/extensions/exif_orientation_editor.py
```diff
@@ -0,0 +1,66 @@
...
+]
+
+EXIF_HEADER = b"Exif\x00\x00"
+
+
+class ExifOrientationEditor():
+    _endian = "<"
+    _offset = None
+
+    def __init__(self, data):
```
okor: Generally speaking, this `__init__` method is quite dense and very difficult to tell what is going on ... at least for a non-pythonist/non-image expert. It may be helpful for future devs to explain what is going on here via comments or even named functions.

thumbor/engines/extensions/exif_orientation_editor.py
kkopachev: File pointer. 
Struggling to find better name. Any suggestions?

okor: Well, since ... I think ... what you are doing here is declaring a _named_, binary buffer for `data[6:]` range of bytes ... we might be able to call this `exif_buffer`?

heynemann: I agree with changing the names to more descriptive variables. Great work btw.

heynemann:
As soon as we get more descriptive names, I think this is good to merge. Thanks @kkopachev!!! Great work!

kkopachev:
Thank you for your feedback. Take a look now please.

thumbor/engines/extensions/exif_orientation_editor.py
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            raise SyntaxError("not a TIFF IFD")
+
+        return header
+
+    def _find_orientation_offset(self, header):
+        ifd_offset, = self._unpack("L", header[4:])
+        self.exif_buffer.seek(ifd_offset)
+
+        for i in range(self._unpack("H", self.exif_buffer.read(2))[0]):
```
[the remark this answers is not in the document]
kkopachev: Sure, updated.

thumbor/engines/extensions/exif_orientation_editor.py
```diff
@@ -0,0 +1,72 @@
...
+            tag, typ, count, data = self._unpack("HHL4s", self.exif_buffer.read(12))
```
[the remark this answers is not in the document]
kkopachev: figured we don't really need all these info, so just unpack tag id.

kkopachev:
Any updates here?

heynemann:
This looks pretty good! Thanks @kkopachev !!!

PR #1231, opened 2019-10-25
Optionally serve the original image if it is smaller.

ehlertij:
If the original image is smaller, we might as well respond with that instead of the processed one as long as we didn't run it through any filters.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -66,6 +66,8 @@ class Engine(BaseEngine):
 
     def create_image(self, buffer):
         try:
+            if self.context.config.SERVE_ORIGINAL_IF_SMALLER and hasattr(self.context.request.filters, '__len__') and len(self.context.request.filters) == 0:
+                self.original = BytesIO(buffer).getvalue()
             img = Image.open(BytesIO(buffer))
         except DecompressionBombExceptions as e:
             logger.warning("[PILEngine] create_image failed: {0}".format(e))
@@ -263,6 +265,14 @@ class Engine(BaseEngine):
 
         results = img_buffer.getvalue()
         img_buffer.close()
+
+        if self.context.config.SERVE_ORIGINAL_IF_SMALLER and hasattr(self.context.request.filters, '__len__') and len(self.context.request.filters) == 0:
+            logger.debug('Original length: %s', len(self.original))
+            logger.debug('Compressed length: %s', len(results))
+
+            if len(results) >= len(self.original):
+                return self.original
+
         self.extension = ext
         return results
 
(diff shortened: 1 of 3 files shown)


thumbor/engines/pil.py
```diff
@@ -263,6 +265,14 @@ def read(self, extension=None, quality=None):  # NOQA
 
         results = img_buffer.getvalue()
         img_buffer.close()
+
+        if self.context.config.SERVE_ORIGINAL_IF_SMALLER and hasattr(self.context.request.filters, '__len__') and len(self.context.request.filters) == 0:
```
kkopachev: Does not seem like a good idea to do it in one engine. Probably would be better to make it in handler somewhere near place where it tries to fit into bytes.

ehlertij: Good point, thanks. I'll see if I can find a better spot to put that!

ehlertij:
Alright, I moved the logic into `thumbor/handlers/__init__.py` and I think it makes a lot more sense there. Thanks for that feedback!

heynemann:
Also thumbor's transformation is not only filters. The only case where I can see returning the original being better is if no filters and no cropping and no resize was done, i.e.: nothing. Which is quite weird. What scenario are you using thumbor for?

ehlertij:
Even with no filters, cropping, or resizing thumbor is still optimizing the images and reducing their size, including converting them to more modern formats (WEBP, JPEG2000, etc). In this case, if the "optimized" image is larger than the original, that kind of defeats the purpose. Also, if the resize maintains the same aspect ratio, the same applies.

For apps/websites that are just trying to reduce page weight without changing the images, this can make a lot of sense.

As for other transformations, I'll have to figure out how to check for those as well.

alexandernst:
Can we get this merged?

heynemann:
Not really. The check for filters length is not enough as that does not check whether there were resize or cropping (auto or manual). The way this PR is right now, if the new image is larger than the original you would lose cropping and resizing.

heynemann:


```diff

Complexity increasing per file
==============================
- thumbor/handlers/__init__.py  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=4967911&bid=16414094)

heynemann:
Hey everyone, sorry this is taking so long. For me the issue here is with resize being thrown away. Even if the image maintains the aspect ratio and it is larger in KB, not having to do the resize while showing the image may be what you are after. In mobile devices, resizing images on the fly while rendering a page is very costly.

What I'm trying to convey here is that there is a contract to thumbor's resizing. If you ask for 300x200, you get a 300x200 image (unless you specify filters that modify dimensions). This change would kill that in a very unpredictable way.

If you do check for no cropping or resizing and no filters, we're good to go, but I believe that would defeat the purpose of this PR, wouldn't it?

PR #1232, opened 2019-10-29
Allow Pillow 6.x

andersk:
We should never use <= because that disallows security point releases. 6.1.x should have been allowed using `<6.2`.

It happens that 6.2.0 was also a security release, and it seems reasonable to allow all minor releases like we did in the past.

Pillow has announced that [7.0.0 will be the first release to drop Python 2 support](https://pillow.readthedocs.io/en/stable/releasenotes/6.2.1.html#python-2-7).


andersk:
It’d be nice to have a Thumbor release with this fix. Thumbor 6.7.0 requires Pillow < 6.0.0, making it incompatible with Pillow versions that include current security fixes, such as CVE-2019-16865.

PR #1000, opened 2017-10-26
Yield coroutines

kkopachev:
Reason for these changes is that sometimes, when exception happens inside coroutine which was not yielded, it got trapped inside Future object for timeout (of 60 seconds) and request hangs for that time (since handler methods are `@asynchronous`)

Unyielded coroutines create other issues as well, like [inability to write a test for fail case ](https://github.com/thumbor/thumbor/pull/932)

Also, crossing fingers, that such change could help with memory leaks.

Also would be nice to have #894 merged before this one.

PR #1235, opened 2019-11-16
Py3 compatibility

kkopachev:
please take a look, let me know.

### DIFF
--- a/thumbor/filters/watermark.py
+++ b/thumbor/filters/watermark.py
@@ -27,7 +27,7 @@ class Filter(BaseFilter):
             return pos
 
         sign, ratio = match.groups()
-        pos = "{sign}{pos}".format(sign=sign, pos=int(round(length * float(ratio) / 100 + 1e-5, 0)))
+        pos = "{sign}{pos}".format(sign=sign, pos=round(length * float(ratio) / 100 + 1e-5))
 
         return pos
 
--- a/thumbor/filters/watermark.py
+++ b/thumbor/filters/watermark.py
@@ -8,6 +8,7 @@
 # http://www.opensource.org/licenses/mit-license
 # Copyright (c) 2011 globo.com thumbor@googlegroups.com
 
+from __future__ import division
 from thumbor.ext.filters import _alpha
 from thumbor.filters import BaseFilter, filter_method
 from thumbor.loaders import LoaderResult
@@ -27,7 +28,7 @@ class Filter(BaseFilter):
             return pos
 
         sign, ratio = match.groups()
-        pos = "{sign}{pos}".format(sign=sign, pos=int(round(length * float(ratio) / 100, 0)))
+        pos = "{sign}{pos}".format(sign=sign, pos=int(round(length * float(ratio) / 100 + 1e-5, 0)))
 
         return pos
 
@@ -42,7 +43,7 @@ class Filter(BaseFilter):
         if not wm_max_height:
             wm_max_height = watermark_sz[1] * wm_max_width / watermark_sz[0]
 
-        if float(watermark_sz[0]) / wm_max_width >= float(watermark_sz[1]) / wm_max_height:
+        if watermark_sz[0] / wm_max_width >= watermark_sz[1] / wm_max_height:
             wm_height = round(watermark_sz[1] * wm_max_width / watermark_sz[0])
             wm_width = round(wm_max_width)
         else:
@@ -86,9 +87,9 @@ class Filter(BaseFilter):
         if not mos_x:
             repeat_x = (1, 0)
             if center_x:
-                x = (sz[0] - watermark_sz[0]) / 2
+                x = int((sz[0] - watermark_sz[0]) / 2)
             elif inv_x:
-                x = (sz[0] - watermark_sz[0]) + x
+                x = int((sz[0] - watermark_sz[0]) + x)
         else:
             repeat_x = divmod(sz[0], watermark_sz[0])
             if sz[0] * 1.0 / watermark_sz[0] < 2:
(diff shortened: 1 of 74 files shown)


thumbor/filters/watermark.py
```diff
@@ -27,7 +28,7 @@ def detect_and_get_ratio_position(pos, length):
             return pos
 
         sign, ratio = match.groups()
-        pos = "{sign}{pos}".format(sign=sign, pos=int(round(length * float(ratio) / 100, 0)))
+        pos = "{sign}{pos}".format(sign=sign, pos=int(round(length * float(ratio) / 100 + 1e-5, 0)))
```
cristiandean: Is `float` necessary here?

thumbor/handlers/__init__.py
```diff
@@ -39,6 +39,13 @@
 except NameError:
     basestring = str  # Python 3
 
+if sys.version_info.major >= 3:
```
cristiandean: I think it would be nice to keep a version check standard. Here we have `sys.version_info.major`, and here https://github.com/thumbor/thumbor/blob/6292cb58b866b8267fba1c0a5974a0f05d7963ac/thumbor/engines/extensions/pil.py#L86 with `six`.

kkopachev: Maybe we should just drop Py2 compatibility all together. Less than 2 months left for Py2 support anyway. 
Thoughts?

kkopachev:
Btw, I ran integration suit on python2, got content-length of response for each test case. Then compared content-length when on running python3.

thumbor/handlers/__init__.py
cristiandean: I agree, I think it's worth removing support for python 2.

cristiandean:
> Btw, I ran integration suit on python2, got content-length of response for each test case. Then compared content-length when on running python3.

Do the lengths match?

thumbor/filters/watermark.py
andersk: Presumably yes, since `ratio` is a string. Or maybe you’re asking if this could this be done with integer arithmetic: `(length * int(ratio) + 50) // 100`?

kkopachev:
> > Btw, I ran integration suit on python2, got content-length of response for each test case. Then compared content-length when on running python3.

> Do the lengths match?

Of course :100:

thumbor/filters/watermark.py
kkopachev: Had to keep `float` because `ratio` is indeed string, but managed to remove outer int.

kkopachev:
I think this is good for another round of review. 
I can't seem to understand what's wrong with Codacy and how to fix it possibly. Any suggestions welcome.

andersk:
Every “new” Codacy issue corresponds to a “fixed” Codacy issue on the previous version of the same line, so that’s probably not worth worrying about. It looks like Codacy has a way to mark issues ignored within the web interface; no idea if there’s a way to do so within the code itself.

cristiandean:
Awesome!!! LGTM, I wish someone else could take a look at this PR due to its size. I think we could release a beta for a while, what do you think @kkopachev ?

kkopachev:
I think we can do beta for some time. 
I was going to try this at work, but figured that aws loader/storage/result_storage does not support Python3 yet :( If someone is using local files can test, that would be awesome. I was going to upgrade tc_aws project too, but time limits hit me.

While we're breaking all the things, I think it's good time to upgrade deprecated `@return_future` and `@web.asynchronous` to `@gen.coroutine`. Maybe even upgrade straight to `async/await`.
Perhaps we can incorporate #1000 into new beta version.
Trying to balance here and take advantage of the fact that we're going to break BC anyway and not to change too many things at once.

alexandernst:
How is this going?

heynemann:
Thanks for all your hard effort @kkopachev!!! I took over this (with your commits included and rebased in the py3 branch in thumbor). Just merged this into master.

heynemann:
Btw I have a branch for asyncawait and that's shipping after py3 :)

kkopachev:
Awesome :100: 

@heynemann we should sync up on future moves. There are few APIs awaiting to be deprecated/removed given that it'll be new major version. Few places has to handle futures/async-awaits properly and etc. 
Maybe tracking issue or PR or just chat over gitter?

kkopachev:
@heynemann I think master branch is messed up with rebase. Might want to rebase it again and force-push.

heynemann:
Hey you mean force pushing the py3 branch to master? I'll do it tonight.

heynemann:
On the async/await bit, you can check it in the branch. I mean to modernize the whole of thumbor. The only part I have not changed to be async is the engine, since it already runs inside of threadpool if that's setup. I've upgraded tornado to the latest version as well and much of the old future/corouting stuff is gone from tornado 6.

kkopachev:
Oh, I though you didn't meant to merge py3 changes into master for quite some time. :man_shrugging: 

On async/await, could you open PR (perhaps draft) or point me to a branch?

heynemann:
Why do you think master is broken? I created a branch from previous master, pulled your changes in (git pull --rebase kkopachev py3) and then pushed them to master. Is there anything you feel is wrong? The build is green.

The branch for async/await is https://github.com/thumbor/thumbor/tree/asyncawait

kkopachev:
Oh, then no issues here :ok_hand: 
I just thought you didn't want to merge it all into master branch. 
Will look into asyncawait branch!

heynemann:
I created the branch and the build should be green. The process of converting everything to async/await brought to my attention how many things are lacking tests. Mainly filters and detectors :/

My next step with thumbor codebase is removing unnecessary filters/detectors/loaders/storages/result storages and even routes (Upload is going away). 

I want to create thumbor community projects for each of the removed items (or group of items). Thumbor's codebase should be as lean as possible. That is what will allow us to improve it (add more tests, refactor mercilessly).

heynemann:
The build for async/await branch is green. I still want to refactor the base tests into a thumbor.testing namespace, so thumbor plugin authors can reuse our test case and make their tests simpler. Other than that, this probably should be our python 3 release (since it breaks everything anyway). WDYT?

heynemann:
BTW I'm in thumbor's slack if any of you ever want to reach out. I'm also using the project view here in github to create what I feel are the next steps for thumbor. If anyone wants to pick a task just let me know.

heynemann:
Project View = https://github.com/thumbor/thumbor/projects/1

kkopachev:
I probably missed slack. Can you send an invite?

PR #1246, opened 2020-01-15
Added new loader "file_loader_404_fallback.py" for inexistent files

### LINKED ISSUE #1245: Add a fallback image for inexistent files
### Expected behaviour
When a request is made for a non existent file, I want to show a generic image. This way I can have  a fallback for ecommerce sites that show a default image and not an error. This must be optional, as it is an specific use case.

ediazcomellas:
Added a new loader "file_loader_404_fallback.py" that loads a fixed file "404.png" image in the file loader root folder if the requested image does not exist.

This is useful if you want to have a default image for all requests in a e-commerce site, for example.

closes  #1245

### DIFF
--- a/docs/image_loader.rst
+++ b/docs/image_loader.rst
@@ -69,6 +69,19 @@ you'll get an error.
 To use it you should set the **LOADER** configuration to
 **'thumbor.loaders.file\_loader\_http\_fallback'**.
 
+File loader with 404.png default image fallback
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some environments, like an e-commerce site, it is desirable to load a default
+image rather than throwing a 404 error. Yo can use this loader for this situation.
+
+This loader will try to to load the image from local storage. In case of an error, it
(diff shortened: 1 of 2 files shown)


heynemann:
I think this is really useful, but I don't think this should be in the core thumbor install. My advice is create a new project with just this loader or contribute to thumbor-community. Either way it's VERY easy to `pip install thumbor-file-loader-or-404` and add it to your `thumbor.conf`.

ediazcomellas:
> I think this is really useful, but I don't think this should be in the core thumbor install. My advice is create a new project with just this loader or contribute to thumbor-community. Either way it's VERY easy to `pip install thumbor-file-loader-or-404` and add it to your `thumbor.conf`.

That is a good point! I was not aware of thumbor-community. It definitely belongs to there, as it is a particular use case.

gabriel-kaam:
Hey there ; I'm trying to do something similar ; but instead of loading a new image , I'd like to generate a 301 Redirect .
I've been playing with a custom loader for a few hours now, but I've unsuccessful so far. 
Any idea on how I could do it ?

PR #1259, opened 2020-02-10
[IMPR] Simplify Pillow feature checks 

kkopachev:


heynemann:
This is great! Do you mind just changing the commit message to conform to the commit template I added? It will make it much easier to create release notes if all our commits use it.

kkopachev:
updated

PR #1261, opened 2020-02-13
[FEAT] URL Routers support

heynemann:
Is Backwards Compatible: yes

This commit adds support for a new class of plugins: Routers.
Routers are responsible for integrating new routes in thumbor's
App.

This is the missing extensiblity part in thumbor and one of the
reasons scope has crept in it (upload support).
After this has been released we'll start separating some of these
into their own modules.

One example apart from upload is healthcheck. While thumbor will
ship with a default healthcheck that's backwards compatible,
it will allow plugin builders to implement their own healthcheck
module (verifies if services that thumbor depends on are
available).

Creating a router is as simple as extending an abstract
`BaseRouter` and implementing the required methods.

Configuring which routers to use should be very familiar to any
thumbor user and is just a config value with the fully qualified
name of the class.

### DIFF
--- a/thumbor/app.py
+++ b/thumbor/app.py
@@ -12,11 +12,7 @@ import tornado.ioloop
 import tornado.web
 from libthumbor.url import Url
 
-from thumbor.handlers.blacklist import BlacklistHandler
-from thumbor.handlers.healthcheck import HealthcheckHandler
-from thumbor.handlers.image_resource import ImageResourceHandler
 from thumbor.handlers.imaging import ImagingHandler
-from thumbor.handlers.upload import ImageUploadHandler
 
 
 class ThumborServiceApp(tornado.web.Application):
@@ -26,22 +22,13 @@ class ThumborServiceApp(tornado.web.Application):
         super(ThumborServiceApp, self).__init__(self.get_handlers(), debug=self.debug)
 
     def get_handlers(self):
-        handlers = [
-            (self.context.config.HEALTHCHECK_ROUTE, HealthcheckHandler),
-        ]
-
-        if self.context.config.UPLOAD_ENABLED:
-            # Handler to upload images (POST).
-            handlers.append((r"/image", ImageUploadHandler, {"context": self.context}))
-
-            # Handler to retrieve or modify existing images  (GET, PUT, DELETE)
-            handlers.append(
-                (r"/image/(.*)", ImageResourceHandler, {"context": self.context},)
-            )
-
-        if self.context.config.USE_BLACKLIST:
-            handlers.append(
-                (r"/blacklist", BlacklistHandler, {"context": self.context})
+        handlers = []
+        for router in self.context.modules.importer.routers:
+            handlers.extend(
+                [
+                    (route.url, route.handler, route.initialize)
+                    for route in router(self.context).get_routes()
+                ]
             )
 
         # Imaging handler (GET)
--- a/thumbor/routers/__init__.py
+++ b/thumbor/routers/__init__.py
@@ -1,0 +1,15 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+BUILTIN_ROUTERS = [
+    "thumbor.routers.healthcheck.HealthcheckRouter",
+    "thumbor.routers.upload.UploadRouter",
+    "thumbor.routers.blacklist.BlacklistRouter",
+]
--- a/thumbor/routers/base.py
+++ b/thumbor/routers/base.py
@@ -1,0 +1,32 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional
+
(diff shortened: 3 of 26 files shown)


heynemann:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- thumbor/importer.py  1
- tests/routers/test_base_router.py  1
- tests/routers/test_blacklist_router.py  1
- thumbor/routers/blacklist.py  2
- tests/routers/test_healthcheck_router.py  1
- tests/routers/test_upload_router.py  1
- thumbor/routers/healthcheck.py  1
- thumbor/routers/base.py  1
- thumbor/routers/upload.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=5009545&bid=16524179)

heynemann:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- thumbor/routers/blacklist.py  2
- thumbor/routers/upload.py  2
- tests/routers/test_healthcheck_router.py  1
- thumbor/routers/healthcheck.py  1
- tests/routers/test_base_router.py  1
- tests/routers/test_upload_router.py  1
- thumbor/importer.py  1
- thumbor/routers/base.py  1
- tests/routers/test_blacklist_router.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=5009545&bid=16524179)

heynemann:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- tests/routers/test_upload_router.py  1
- thumbor/routers/blacklist.py  2
- thumbor/routers/healthcheck.py  1
- thumbor/importer.py  1
- thumbor/routers/base.py  1
- tests/routers/test_base_router.py  1
- tests/routers/test_healthcheck_router.py  1
- tests/routers/test_blacklist_router.py  1
- thumbor/routers/upload.py  2
         

Clones added
============
- thumbor/filters/__init__.py  1
         

Clones removed
==============
+ tests/handlers/test_base_handler.py  -2
+ thumbor/config.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=5009545&bid=16524179)

thumbor/app.py
```diff
@@ -26,22 +22,13 @@ def __init__(self, context):
...
-            # Handler to retrieve or modify existing images  (GET, PUT, DELETE)
-            handlers.append(
-                (r"/image/(.*)", ImageResourceHandler, {"context": self.context},)
-            )
-
-        if self.context.config.USE_BLACKLIST:
-            handlers.append(
-                (r"/blacklist", BlacklistHandler, {"context": self.context})
+        handlers = []
+        for router in self.context.modules.importer.routers:
```
kkopachev: Name `router`/`routers` seems weird. I think name `handlers` has better fit. After all you're putting routes into handlers.

thumbor/routers/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+BUILTIN_ROUTERS = [
+    "thumbor.routers.healthcheck.HealthcheckRouter",
+    "thumbor.routers.upload.UploadRouter",
```
kkopachev: is it a good time to move this into it's own repo?

thumbor/routers/base.py
```diff
@@ -0,0 +1,32 @@
...
+        self.handler = handler
+        self.initialize = initialize
+
+
+class BaseRouter(ABC):
+    def __init__(self, context: Any):
+        self.context = context
+
+    @abstractmethod
+    def get_routes(self) -> Optional[List[Route]]:
```
kkopachev: why `Optional`?

kkopachev:
@heynemann I am currently in such period of my dev life that I think it's beneficial to avoid extra classes and inheritance. 
As such, I don't see much value in having Router class and subclasses at all. Instead it all could be a collection of functions which receive context and return list of routes. Even routes themselves I am not sure about. It's good that they are value object and such, but not sure if it would be any worst if "routers" return list of handlers as how tornado expects them.

thumbor/app.py
heynemann: I just picked the name after tornado's routing namespace...

thumbor/routers/__init__.py
heynemann: Yes! :) If you feel like doing it i'm all for it. I plan on contributing with thumbor-aws to get it up and running before, though.

thumbor/routers/base.py
heynemann: Because you can return None or empty list if you check a configuration that disables the router. Maybe just List[Route] would be good enough as you can return an empty list, right?

thumbor/app.py
kkopachev: yes, but after very brief look into tornado documentation, it appears their routers are actually routing - finding appropriate handler. In thumbor case that's not what's happening. Thumbor always uses RuleRouter.

thumbor/routers/base.py
kkopachev: I am for empty lists instead of optionals.

heynemann:
I understand and please know that I deeply respect your opinion and I hope you keep giving it for free hehehe. In this instance I'm not sure I agree. We've been down that path with the loaders and it's harder to evolve them precisely because they are not classes.

It's harder to bake in private functionality and them feel confidence in changing it later on. If we do it must be a helper function that the user's are expected to call, thus leaking our implementation.

Regarding the Route object I'm thorn. Maybe you're right and I went a bit far. We could return tuples and be done with it :) Still time to change it. WDYT?

thumbor/routers/__init__.py
kkopachev: Let's sync up on aws stuff in slack. I had some thoughts.

PR #1273, opened 2020-02-29
[IMPR] Performance tests

heynemann:

This commit adds performance load testing to our build.

The reasons to have a good load testing suite abound:

1) This ensures that all our main functionality works. Think of it
as an acceptance test suite.
2) It allows us to easily compare the performance when making changes to
   thumbor.
3) It ensures that we won't completely break thumbor's performance in
   the future.


kkopachev:
Few questions on this.

Do you think travis's environment reliable enough?

What's acceptance criteria? I saw in the code, p95 should be below 3s, but in travis I see it's  above 6s
image

kkopachev:
Would also be nice to mention sources of images.

heynemann:
The threshold for travis is set in run.sh with the P95=8000 env var.
As for whether it is reliable enough, we'll see.
Even if we need to increase the p95, these tests check status 200, so they double as acceptance tests (in they go through the entire stack of thumbor, even cli and server).
Another benefit is running performance locally (if you want to try uvloop instead of tornado's ioloop, it's now easy to compare).
The source of the images is the same as in the docs, but you're right I should add it in the test as well.

PR #1281, opened 2020-03-13
[FIX] Ability to listen to a socket passed as file descriptor

### LINKED ISSUE #1269: Trying to launch thumbor with a unix socket ends with "No such device or address"

gael-ian:
Should fix #1269.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,7 @@ http://<thumbor-server>/300x200/smart/thumbor.readthedocs.io/en/latest/_images/l
             "derpconf>=0.2.0",
             "statsd>=3.0.1",
             "libthumbor>=1.3.2",
+            "socketfromfd>=0.2.0",
             "futures",
             "argparse",
             "pytz",
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -183,7 +183,7 @@ class ServerTestCase(TestCase):
         server_instance_mock = mock.Mock()
         server_mock.return_value = server_instance_mock
 
-        socket_from_fd_mock.fromfd.return_value = "socket mock"
+        socket_from_fd_mock.return_value = "socket mock"
 
         run_server(application, context)
         server_instance_mock.add_socket.assert_called_with("socket mock")
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -174,8 +174,8 @@ class ServerTestCase(TestCase):
         server_instance_mock.start.assert_called_with(1)
 
     @mock.patch.object(thumbor.server, 'HTTPServer')
-    @mock.patch.object(thumbor.server, 'socket')
-    def test_can_run_server_with_fd(self, socket_mock, server_mock):
+    @mock.patch.object(thumbor.server, 'socket_from_fd')
+    def test_can_run_server_with_fd(self, socket_mock, socket_from_fd_mock):
         application = mock.Mock()
         context = mock.Mock()
         context.server = mock.Mock(fd=11, port=1234, ip='0.0.0.0')
@@ -183,7 +183,7 @@ class ServerTestCase(TestCase):
         server_instance_mock = mock.Mock()
         server_mock.return_value = server_instance_mock
 
-        socket_mock.fromfd.return_value = "socket mock"
+        socket_from_fd_mock.return_value = "socket mock"
 
         run_server(application, context)
         server_instance_mock.add_socket.assert_called_with("socket mock")
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -175,7 +175,7 @@ class ServerTestCase(TestCase):
 
     @mock.patch.object(thumbor.server, 'HTTPServer')
     @mock.patch.object(thumbor.server, 'socket_from_fd')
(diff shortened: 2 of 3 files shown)


gael-ian:
I realized today that not only AF_UNIX sockets can be passed as file descriptors. AF_INET / AF_INET6 sockets created by systemd, Supervisor or another orchestrator will be passed the same way to dependent services.

In Python 3.8, we could have used `socket(fileno=fd_number)` to turn any file descriptor to a proper socket object but thumbor runs with Python 2.7 and there's no such option on `socket()`.

Instead of trying to guess the correct family with some complex code, my proposition is to let users pass it as an argument and reuse the `--port` argument for this (as there no sense to pass a file descriptor and a port at the same time).

Hence a thumbor invocation would look like:

* `thumbor --fd /run/thumbor.sock` (AF_UNIX socket passed by path)
* `thumbor --fd 3 --port 1` (AF_UNIX socket passed by file descriptor)
* `thumbor --ip 127.0.0.1 --port 8080` (AF_INET socket)
* `thumbor --fd 3 --port 2` (AF_INET socket passed by file descriptor)
* `thumbor --ip ::1 --port 8080` (AF_INET6 socket)
* `thumbor --fd 3 --port 10` (AF_INET6 socket passed by file descriptor)

1, 2 and 10 are numeric values respectively associated to Python's socket `AF_UNIX`, `AF_INET` and `AF_INET6` constants.

kkopachev:
Knowing to pass numbers is too magical? 

Another option to use https://github.com/tiran/socketfromfd which auto detects socket family and type.

gael-ian:
> Another option to use https://github.com/tiran/socketfromfd which auto detects socket family and type.

Much more interesting than argument cycling, indeed. I change the pull request to use it.

kkopachev:
That looks much cleaner, indeed. Would you fix the test please? Looks like it's trying to mock not existing anymore `socket`. That should now be `fromfd` mock instead

gael-ian:
> Would you fix the test please?

Tomorrow. It's too late here to dig now into what are pip, virtualenv and the whole Python (2...) ecosystem.

gael-ian:
Sorry, the history of this branch is a mess but I had to rely on TravisCi to run tests.

---

I tried to follow [contribution instructions](https://github.com/thumbor/thumbor/blob/master/CONTRIBUTING.md) but could not set up a development environment on my machine.

```shell script
$ python -V
Python 3.8.2
$ pip -V
pip 20.0.2
$ poetry --version
Poetry version 1.0.5
$ pip install -e .[tests]
[...]
ERROR: remotecv 3.0.0 has requirement Pillow<8.0.0,>=7.0.0, but you'll have pillow 6.2.2 which is incompatible.
$ make test                        
[... many failing tests ...]
make[1]: *** [Makefile:57: unit] Segmentation fault (core dumped)
make[1]: Leaving directory '[...]/thumbor'
make: *** [Makefile:39: test] Error 2
```

I noticed that, even on TravisCi, `remotecv` has unmet dependencies.

kkopachev:
we should probably limit remotecv version to be below 3 for thumbor 6

heynemann:
Should we cherry pick this to master?

kkopachev:
We should. Not sure if python's `socket.socket(fileno=)` works as expected in all pythons 3.

heynemann:
Do you mind cherry-picking it? If you can't, I'll find some time to do it.

Please be mindful of time spent here. **Your safety and getting your loved ones safe is the most important thing.** We can always work things out here later on! :) Hope you are safe and everyone you love as well.

PR #1284, opened 2020-03-21
[FEAT] thumbor-doctor

heynemann:

Thumbor doctor is a console app that users can run to verify their
thumbor install. That way they can send it with their issues in order to
make it easier for contributors to investigate what's going on.

The idea is that we keep adding checks to thumbor doctor over time.
It can be refactored to make it easier to add a new check, but this will
do for now.

The output when there's an error is like this:
```
Thumbor v7.0.0a4 (of 9-Mar-2020)

Thumbor doctor will analyze your install and verify if everything is working as expected.

Verifying libraries support...

❎ pycurl is not installed.
Thumbor works much better with PyCurl. For more information visit http://pycurl.io/.

✅ cv2 is installed correctly.
✅ pyexiv2 is installed correctly.
✅ cairosvg is installed correctly.

Verifying thumbor compiled extensions...

✅ _alpha
✅ _bounding_box
✅ _brightness
✅ _colorize
✅ _composite
✅ _contrast
✅ _convolution
✅ _curve
✅ _equalize
✅ _fill
✅ _nine_patch
✅ _noise
✅ _rgb
✅ _round_corner
✅ _saturation
✅ _sharpen

Verifying thumbor filters...

✅ thumbor.filters.brightness
✅ thumbor.filters.colorize
✅ thumbor.filters.contrast
✅ thumbor.filters.rgb
✅ thumbor.filters.round_corner
✅ thumbor.filters.quality
✅ thumbor.filters.noise
✅ thumbor.filters.watermark
✅ thumbor.filters.equalize
✅ thumbor.filters.fill
✅ thumbor.filters.sharpen
✅ thumbor.filters.strip_exif
✅ thumbor.filters.strip_icc
✅ thumbor.filters.frame
✅ thumbor.filters.grayscale
✅ thumbor.filters.rotate
✅ thumbor.filters.format
✅ thumbor.filters.max_bytes
✅ thumbor.filters.convolution
✅ thumbor.filters.blur
✅ thumbor.filters.extract_focal
✅ thumbor.filters.focal
✅ thumbor.filters.no_upscale
✅ thumbor.filters.saturation
✅ thumbor.filters.max_age
✅ thumbor.filters.curve
✅ thumbor.filters.background_color
✅ thumbor.filters.upscale
✅ thumbor.filters.proportion
✅ thumbor.filters.stretch

Verifying extension programs...

✅ jpegtran is installed correctly.
✅ ffmpeg is installed correctly.
✅ gifsicle is installed correctly.

😞 Oh no! We found some things that could improve... 😞

* No module named 'pycurl' - Thumbor works much better with PyCurl. For more information visit http://pycurl.io/.


If you don't know how to fix them, please open an issue with thumbor.
Don't forget to copy this log and add it to the description of your issue.
Open an issue at https://github.com/thumbor/thumbor/issues/new
```

### DIFF
--- a/thumbor/doctor.py
+++ b/thumbor/doctor.py
@@ -1,0 +1,174 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+import argparse
+import sys
+from importlib import import_module
+from shutil import which
+
+import colorful as cf
+
+from thumbor import __release_date__, __version__
+from thumbor.ext import BUILTIN_EXTENSIONS
+from thumbor.filters import BUILTIN_FILTERS
+
+
+def get_options():
+    parser = argparse.ArgumentParser(description="thumbor doctor")
+
+    parser.add_argument(
+        "-n",
+        "--nocolor",
+        action="store_true",
+        help="Disables coloring of thumbor doctor",
+    )
+
+    options = parser.parse_args()
+
+    return {
+        "nocolor": options.nocolor,
+    }
+
+
+def header(msg, color=cf.yellow):
+    print(color(msg))
+
+
+def subheader(msg, color=cf.bold_coral):
+    print(color(msg))
+    newline()
+
+
+def newline():
+    print()
+
+
+def check_filters():
+    newline()
+    subheader('Verifying thumbor filters...')
+    errors = []
+
+    for filter_name in BUILTIN_FILTERS:
+        try:
+            import_module(filter_name)
+            print(cf.bold_green('✅ %s' % filter_name))
+        except ImportError as error:
+            print(cf.bold_red('❎ %s' % filter_name))
+            errors.append(error)
+
+    return errors
+
+
+def check_compiled_extensions():
+    newline()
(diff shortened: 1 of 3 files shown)


heynemann:
Waiting for your call on whether to merge this to master @kkopachev. I believe this is going to help when users have problems and open issues. If they can send the log this tool produces it should be easier to diagnose what's going on. Would you like to add anything else to the log?

thumbor/doctor.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def check_modules():
+    newline()
+    subheader('Verifying pycurl support...')
+    errors = []
+
+    modules = (
+        ('pycurl', 'Thumbor works much better with PyCurl. For more information visit http://pycurl.io/.'),
+        ('cv2', 'Thumbor requires OpenCV for smart cropping. For more information check https://opencv.org/.'),
```
kkopachev: should also include exiv2 and cairosvg?

heynemann: Also added checks for programs (jpegtran, gifsicle and ffmpeg).

kkopachev: aren't ptah to ffmpeg (at least) has to be configured in config var? In that case, can't rely on `which`

heynemann: It may be added in config, but usually not being able to `which` it is a good indicator that it's not found. I plan on adding a `-c config_path` so we can get more stuff from the config and test it (are paths to storage readable and writable?). But for now I think this is much better than nothing :D

PR #1297, opened 2020-05-16
[FEAT] Storage initialization and shutdown

heynemann:

This change allows storage authors to add initialization and
shutdown methods to their implementations.

This is very useful in some cases like when you need to pre-allocate
resources like a pool, threads or the filesystem, and then
deallocate them when thumbor goes down.


heynemann:
Just formatted the whole codebase in master so now this PR is only actual changes. Less noise more signal. :)

heynemann:


```diff

Issues
======
+ Solved 7
- Added 4
           

Clones removed
==============
+ setup.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/heynemann/thumbor/pullRequest?prid=5536554&bid=18149840)

kkopachev:
Code looks good, but concept is weird.

(storage) modules are created during server start up, initialize called and then that module instance abandoned and never used.
When request comes in, ContextHandler [creates new Context](https://github.com/thumbor/thumbor/blob/master/thumbor/handlers/__init__.py#L787) which in turns creates new instance of storage.
That part of initialization on an object which is going to be abandoned feels weird.

Also, for usecase like tc_aws, we want to initialize connection pool for loaders and result_storage too. Not quite sure how well it's going to play with object-less loaders. 

There are other modules which needs initializers/finalizers. ThredPool already have shudown, Context object has `__exit__` and shuts down engine.

Perhaps it's better to generalize initializers/finalizers and make them work for all loaded modules. 

Also Context object on Server level and on Request level being  2 different objects makes things cumbersome and obscure.

heynemann:
Need to rethink this. Closing this PR.

PR #1317, opened 2020-10-05
Apply embedded image profile when converting to RGB

znerol:
This is an alternative approach for #927

Apply icc profiles embedded in images automatically when `image_data_as_rgb` is called.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -34,6 +34,15 @@ try:
 except ImportError:
     FILTERS_AVAILABLE = False
 
+try:
+    from PIL import ImageCms
+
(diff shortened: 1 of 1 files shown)


kkopachev:
Some people would like to keep images in original colorspace with attached profile. Converting all images to sRGB would mean lost data.

znerol:
Right, the profile should only be updated when converting from another color mode (e.g. `CMYK`).

heynemann:
Closing this PR since we had no activity in two years in the interest of keeping our PR list leaner. If this is something we still want to pursue, please reopen and I'll look into it. Thanks!

PR #1318, opened 2020-10-05
Pin py3exiv2 to a version < 0.8.0

znerol:


setup.py
```diff
@@ -25,7 +25,7 @@
     "mock==3.*,>=3.0.5",
     "numpy==1.*,>=1.18.1",
     "preggy==1.*,>=1.4.4",
-    "py3exiv2==0.*,>=0.7.1,!=0.7.2",
+    "py3exiv2==0.7.1",
```
scorphus, line 28:
```suggestion
    "py3exiv2==0.*,>=0.7.1,!=0.7.2,!=0.8.0",
```

scorphus (changes requested):
Thanks for contributing, @znerol.

Pinning is not ideal in situations like this. It would be better to exclude `0.8.0` in the hope that a future release won't be problematic. So, please consider my requested changes.

setup.py
znerol: I respectfully disagree for the following reason: Every new release of `py3exiv2` shipping with a broken setup script will cause automated builds to fail at unforeseeable points in time. So instead of hoping that some future release will include a fix, I suggest instead to monitor the situation and only lift the version constraints (i.e., exact version pinning) as soon as a working version of the library exists.

I understand that this subject has some potential for bike-shedding. As an author of both _applications_ and _libraries_, I found that my distributions are more robust if I *pin dependencies in applications* (for the sake of reproducibility) but on the other hand specify _looser constraints (ideally semver) in libraries_ (for the sake of flexibility downstream). A good explanation why this is desirable can be found in this [FAQ answer in the rust book](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries). A more [thorough discussion](https://nvie.com/posts/pin-your-packages/) from the author of [pip-tools](https://github.com/jazzband/pip-tools).

znerol: Note that I am not interested in a bike-shedding discussion. Hence if project maintainers prefer looser constraints, then that's okay with me.

scorphus: IMHO, that would be a topic for another issue/PR.

Here's why: I'd probably agree if that was a release dependency. In that case, one could even go at length and argue, for instance, about adopting something like [pip-compile](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile). Also, there would have to be some kind of automated way to do the monitoring and notifying job—otherwise it's just left forgotten. And that would be the case for every single dependency, not only py3exiv2.

znerol: Exactly.

Pushed the requested change.

scorphus: Thanks for the update and for adhering to the commit message pattern, much appreciated. 👍

scorphus (approved):

kkopachev (approved):

scorphus:
Thanks, @znerol!

PR #1321, opened 2020-10-29
Upgrade pillow and pytz requirements

andersk:
The breaking change in [Pillow 8](https://pillow.readthedocs.io/en/stable/releasenotes/8.0.0.html) is the desupporting of Python 3.5, but we already require 3.6.

Remove the upper version bound on pytz because pytz versions are calendar-based rather than semantic.


setup.py
```diff
@@ -113,8 +113,8 @@ def run_setup(extension_modules=None):
             "derpconf==0.*,>=0.8.3",
             "libthumbor==2.*,>=2.0.0",
             "opencv-python-headless==4.*,>=4.2.0",
-            "pillow==7.*,>=7.0.0",
-            "pytz==2019.*,>=2019.3.0",
+            "pillow==8.*,>=8.0.0",
```
kkopachev, line 116: I think we should allow 7 as well

andersk: Okay, updated. (Should we also allow pytz 2019.3.0?)

kkopachev: Yeah, sure. If older version still works and we don’t need anything from new version, there is no reason to restrict.

andersk: Updated again.

kkopachev (approved):

PR #1323, opened 2020-10-30
[FIX] Make fill with blur engine-agnostic

scorphus:
The `fill` filter, when used with color set to "blur", depended on the engine having the `blur` method. That broke the filter when the engine in use doesn't fulfill such dependency.

By using the already existing blur capabilities of the `blur` filter, `fill(blur)` now wokrs for engines other than PIL.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -17,7 +17,6 @@ from PIL import (
     Image,
     ImageDraw,
     ImageFile,
-    ImageFilter,
     ImageSequence,
     JpegImagePlugin,
     features as pillow_features,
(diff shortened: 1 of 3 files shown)


kkopachev (approved):

kkopachev:
This looks good, but I was thinking that instead of relying of C-defined blur that we maintain ourselves, maybe it's better just to hardcode use of PIL image for such operations like blur? That way burden of maintenance is on Pillow, it is required dependency to run already, so it has to be there.
Thoughts?

scorphus:
I agree that it could well be a good change. To start with, there would be less code to maintain. Or maybe not that much? I don't know exactly. Also, maybe Python code in filters becomes more complex? We should definitely investigate that and see if it's really worth it — also as part of another PR/issue.

heynemann:
Anything else needed here to merge?

scorphus:
@heynemann thanks for the reminder, I missed the 👍 on my last message. Thanks both for reviewing.

PR #1328, opened 2020-11-05
Make python binary configurable in Makefile

gi11es:
This allows overriding which version of python is run for the extension building via:

 `make PYTHON=python3 test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ run-prod: compile_ext
 	@thumbor -l error -c thumbor/thumbor.conf
 
 setup:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,7 +13,7 @@ setup:
 	@pip install -e .[tests]
```
kkopachev, line 13: Do you think it should be done the same way for pip3?

gi11es: Fixed in the follow-up commit

kkopachev (approved):

PR #1329, opened 2020-11-06
Fix HTTP 500 when request path points to directory

f90d83a8:
Originally if request's path points to directory then the server will fail
and return HTTP 500 status code. Correct behavior is to indicate issue with
request, not the server. Thanks to this fix 404 will be returned.

Returning 5xx status code affects potential Thumbor integration to larger systems,
where 500 response may be treated as Thumbor malfunction and will cause taking it
out of load balancing pool.

### DIFF
--- a/thumbor/loaders/file_loader.py
+++ b/thumbor/loaders/file_loader.py
@@ -10,7 +10,7 @@
 
 from datetime import datetime
 from os import fstat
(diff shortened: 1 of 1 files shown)


kkopachev (approved):

heynemann:
@kkopachev anything else needed here for merging?

PR #1324, opened 2020-10-31
Configure GitHub Actions

scorphus:
<details><summary>Edit: This PR is ready to be reviewed and accepted.</summary>

~~**DRAFT PR!** I'm opening this to collect early reviews.~~

</details>

TravisCI has been great since Thumbor started back in the day. Nowadays, however, it's lagging. It takes longer and longer to have a build runner available. Some builds seem to take forever.

This PR aims at moving the build over to GitHub Actions. These are the required functionalities:

- [x] Python 3.6
- [x] Python 3.7
- [x] Python 3.8
- [x] Lint tests
- [x] Unit tests
- [x] Integration tests
- [x] Performance tests (Python 3.8 only)
- [x] Caching of dependencies
- [x] Coveralls integration
- ~~Slack integration~~ (can be done later)
- ~~? Failure allowance (probably for Python 3.9? Keep reading)~~ Fixed

<details><summary>Edit: The original workflow was simplified with the addition of another one, specific for performance/load tests.</summary>

Some features seem to be missing; others are not fully supported. ~~For example, I couldn't find a _readable_/_maintainable_ way to have different kinds of tests for different versions of Python. On the other hand, if we decide to have performance tests for all available versions, the `matrix` tree can be simplified.~~
</details>

<details><summary>Edit: Step conditionals are working great.</summary>

~~Another thing I'm still investigating is the step conditional. It didn't work at first attempts. For now, there's a workaround as part of the commands in `run` (e.g. `[[ "${{ matrix.python.test }}" != "PERF_TEST" ]] && exit 0`).~~
</details>

<details><summary>Edit: Python 3.9 is now part of the build too. \o/</summary>

~~Python 3.9 could be added too, but some dependencies like [opencv-python-headless](https://pypi.org/project/opencv-python-headless/#files) still don't provide wheels for 3.9, and the build either fails or takes forever only to fail again. We should keep an eye on upcoming [improvements](https://github.com/skvark/opencv-python/issues/382) — I've subscribed.~~
</details>

Merging this PR means the actions will run on every push and PR — just like you see below. Once we cover all our needs, we can — and should IMO — disable the build on Travis.

Any feedback is greatly appreciated.

Happy Halloween!

PR #1333, opened 2020-11-19
Extend the PIL test timeout

gi11es:
The default of 5 seconds can be too short for testing the underlying 472 requests.

Without this, if def_single_params times out it results in an obscure concurrent.futures._base.CancelledError just for the affected request.

I had this issue happen non-deterministically in a Docker container running the test suite.

### DIFF
--- a/integration_tests/pil_test.py
+++ b/integration_tests/pil_test.py
@@ -9,7 +9,7 @@ from .urls_helpers import UrlsTester, single_dataset
 class PILTest(EngineCase):
     engine = "thumbor.engines.pil"
 
-    @gen_test
+    @gen_test(timeout=30)
     async def test_single_params(self):
         if not self._app:
             return True


kkopachev:
Oh, sorry that you got this issue. @scorphus and I spent some time to figure it out couple of weeks ago and he did a fix in his `gh-actions-2` branch: https://github.com/thumbor/thumbor/pull/1324/commits/6ebc631fce2d08067edd8043126aaa7be0356b3c

gi11es:
Sounds good, either way fixes the issue! I can abandon this PR as soon as the other branch gets merged 🙂

scorphus:
I'm sorry to learn that you stumbled upon that issue, @gi11es. I hope it isn't a blocker for you. I've just finished the work on GitHub Actions (ref #1324) which includes the abovementioned fix. Any feedback is super appreciated.

heynemann:
@scorphus I'd say it's best to split the two things. The tests should run properly locally right? It should not be tied into github actions. Maybe I misunderstood...

scorphus:
They still run ok locally, but that depends on how fast your hardware is. Sometimes they pass on GH Actions, other times they fail. Like @gi11es said, 5s seems just too short a timeout. Relaxing it allows for "slightly slower" executions to succeed.

gi11es:
I hit the limit locally running those tests in a Docker container on a fairly recent MBP...

heynemann:
There's no need to do this. Tornado supports an env var to configure this timeout, so you can run the integration tests like this:

```
env ASYNC_TEST_TIMEOUT=30 make integration_run
```

What do you think?

scorphus:
That fix is already included in #1324 and it would help any user who is not aware of that env var or gets lost due to the absence of more clear error messages (like Guilles and I did back then).

IOW this PR can be closed as it is a dupe of part of #1324

scorphus:
Closing this one as #1324 is now merged. Thanks for your contribution, @gi11es. I'm sorry again that you ran into this issue.

PR #1335, opened 2020-11-30
add head_frame filter

return200:
-  1.Make an image to circle
-  2.Add a frame picture out of the circle image
-  3.eg: http://www.foo.com/unsafe/fit-in/100x100/filters:head_frame(www.foo.com/bar.png,100):format(png)/www.foo.com/bar.jpg.png
-  4.Due to my poor knowledge of python, the code may be so ugly or unsuitable function may be used.


return200:
result like this:
https://image.200ok.site/2020/11/30/5fc4ca0c066b4.png

return200:
change the base branch from master to branch-6.x

heynemann:
I'm sorry you went to all this trouble without any guidance from us. And it took us long to reply here. We are not going to be adding filters, loaders, storages, result storages or detectors to thumbor as it is super simple to distribute this as a library.

You can just create a new repository with your plugin that contains the filter and whenever someone wants to use they can just pip install it and add to their configuration file.

Here's an example of a filter plugin repository: https://github.com/globocom/thumbor-distributed-collage-filter

Thanks a lot for being part of our community and once more sorry for the colaboration experience not being a good one.

PR #1340, opened 2021-01-24
[FIX] Exclude py3exiv2 version 0.9.3

scorphus:
Yet another version excluded. Will they ever get it right?


scorphus:
Tests are passing: https://github.com/scorphus/thumbor/actions/runs/507165218

PR #1342, opened 2021-01-24
[IMPR] Compatibility loader,storage,res storage

heynemann:

With this change we introduce compatibility loader, storage and
result storage.

Users can now keep using the libraries they are already used to
(like tc_aws) to load, store and cache results.

At the same time, thumbor's codebase can keep clean with async
functions in all these structures.

We have, intentionally, skipped providing compatibility layers for
filters and detectors as these are not in the way of adopting
thumbor in Python 3, but may review this decision in the future.

Successful Build: https://app.circleci.com/pipelines/github/thumbor/thumbor/30/workflows/3a2e79db-73bc-4942-8b84-72ec4c2f4e3a

### DIFF
--- a/thumbor/compatibility/compat.py
+++ b/thumbor/compatibility/compat.py
@@ -1,0 +1,33 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
+# Licensed under the MIT license:
+# http://www.opensource.org/licenses/mit-license
+# Copyright (c) 2011 globo.com thumbor@googlegroups.com
+
+import asyncio
+
+
+async def compatibility_get(*args, **kw):
+    func = kw.pop("func", None)
+    if func is None:
+        raise RuntimeError(
+            "'func' argument can't be None when calling thumbor's compatibility layer."
+        )
+
+    loop = asyncio.get_event_loop()
+    queue = asyncio.Queue()
+
+    def put(*args):
+        loop.call_soon_threadsafe(queue.put_nowait, args)
+
+    func(*args, **kw, callback=put)
+
+    result = await queue.get()
+    if isinstance(result, (tuple, list)) and len(result) == 1:
+        result = result[0]
+
+    return result
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,9 @@ jobs:
     steps:
       - checkout
       - run:
+          name: Update dependencies
+          command: make setup
+      - run:
           name: Compile Extensions
           command: make compile_ext
       - run:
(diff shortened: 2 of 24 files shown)


thumbor/compatibility/compat.py
```diff
@@ -0,0 +1,33 @@
...
+            "'func' argument can't be None when calling thumbor's compatibility layer."
+        )
+
+    loop = asyncio.get_event_loop()
+    queue = asyncio.Queue()
+
+    def put(*args):
+        loop.call_soon_threadsafe(queue.put_nowait, args)
+
+    func(*args, **kw, callback=put)
```
kkopachev, line 27: There is a good chance I don't understand what's going on...
But this call here is synchronous and blocking. The fact that you put it's  result into queue does not solve anything. Consider the following example where my func is syncronous:
```python
def func(callback):
    sleep(10)
    callback('Great success')
```
in my understanding, such `func` would block the loop until it is finished sleeping. Then it'll call callback, which would put `Great success` string into a queue and then on next loop iteration it'll be popped from queue. However, that does not matter, since plain call to `func` is blocking right here.

heynemann: Hey, thanks for the review! I think maybe we are mixing two different concepts here: non-blocking and async (as in AsyncIO async).

The compatibility layer is not trying to turn blocking storages, loaders and result storages into non-blocking ones. If the storage is blocking it will remain blocking.

If the storage uses non-blocking code with callbacks, it should remain non-blocking. If it is not the case, it should remain blocking. Maybe picking the legacy file loader was not the best scenario as it is blocking. I think I'll change to use the http loader instead as it has non-blocking callback based code.

heynemann: Hmm just tried. Using the legacy http loader is not possible as the callback based fetch has been removed from tornado's http client.

Just to clarify further, this change is to enable loaders, storages and result storages that were working with thumbor 6.x to keep working with thumbor 7 with as small a change as possible (no need to use AsyncIO in any calls).

I'm not sure on how feasible that is given that Python 3 has different requirements on code and modern Tornado has changed quite a bit. It Anyway I believe it was worth the shot. If we can test with tc_aws and it makes it easier it would be great.

heynemann:
@kkopachev just finished testing this compatibility layer. A few clarifications:

1. It does not asyncify anything. If a loader, storage or result storage is currently blocking, it remains blocking.
2. It does allow for callback-based loaders, storages and result storages to be used with thumbor 7.
3. If you are using thumbor's compatibility layer, and tornado does not have `tornado.concurrent.return_future` decorator, we monkey patch it to be a noop. This is just meant to make it easier on transitioning to thumbor 7.

Currently the build is failing due to a problem with gifv and ffmpeg that I'm having a hard time figuring it out. Would you mind approving this PR so I can merge and release thumbor 7.0.0b?

heynemann:
One thing that I can't do for thumbor's users is python 2 support. The storages, loaders, result storages, detectors must be python 3 compliant.

PR #1346, opened 2021-02-08
Update atime in file-based result storage

mpdude:
I am using Thumbor with the file-based result storage. A "M/Monit" job monitors the `result_storage` directory size and will remove files when its size exceeds a given threshold.

This PR proposes to "touch" files in the result storage every time they are used to serve a response. This will change the "access" timestamp only, not the "modified" timestamp.

With this change, the files least recently used (LRU) could easily be found by looking at the `atime` timestamps. Deleting these files first would help to improve cache efficiency.

### DIFF
--- a/thumbor/config.py
+++ b/thumbor/config.py
@@ -371,6 +371,9 @@ Config.define(
 Config.define(
     'RESULT_STORAGE_STORES_UNSAFE', False,
     'Indicates whether unsafe requests should also be stored in the Result Storage', 'Result Storage')
+Config.define(
+    'RESULT_STORAGE_TOUCH_FILES', False,
+    'When enabled, files retrieved from the result storage will have their "last accessed" file information updated', 'Result Storage')
 
 # QUEUED DETECTOR REDIS OPTIONS
 Config.define('REDIS_QUEUE_SERVER_HOST', 'localhost', 'Server host for the queued redis detector', 'Queued Redis Detector')
--- a/thumbor/result_storages/file_storage.py
+++ b/thumbor/result_storages/file_storage.py
@@ -13,8 +13,10 @@ from uuid import uuid4
 from shutil import move
 import pytz
 import hashlib
+import time
 
 from os.path import exists, dirname, join, getmtime, abspath, isdir, isfile
+from os import utime
 
 from thumbor.engines import BaseEngine
 from thumbor.result_storages import BaseStorage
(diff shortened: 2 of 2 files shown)


kkopachev:
I am not sure that this is good change.
That would cause write for every read of a stored file, which might be not desirable. 
I went looking why open/read calls do not update atime and [figured that FS mounted with `relatime` option](https://superuser.com/questions/464290/why-is-cat-not-changing-the-access-time). 
I think you could achieve same behavior you propose here with remounting your FS with `strictatime` option.

mpdude:
@kkopachev that‘s true.

I do know, however, a few situations where filesystems are mounted with `noatime` to explicitly disable this feature and the costs of a write access on every file read. In these situations, a change like the one proposed here could still be useful.

So, would you support the change if it were an opt-in feature/toggle?

kkopachev:
I don't have strong opinion here. Could be added behind a flag, disabled by default. 
Could you mount FS just for thumbor results storage with strictatime option? I am a bit hesitant on introducing new features which would be carried for long time.
On the other hand it's not that big of a deal - just a flag and few lines of code :shrug:

mpdude:
@kkopachev I'm not a Python programmer – does it make sense this way, or am I missing something?

kkopachev (approved):

mpdude:
@kkopachev Should I do anything else to get this merged?

mpdude:
@kkopachev Can I do anything to help getting this merged?

heynemann:
TBH I don't think this should be merged. Thumbor already has many different configuration properties. If all you need is the touch capability it is super easy to just write your own result storage that does it and use that. I'm wary of adding the branching here. I'm sorry it took so long for me to chip in here.

polarathene:
You could probably add a filesystem mount just for this storage as suggested already. That doesn't require making a new partition on the disk, you can use a large file and mount this or even use RAM (probably not a great idea for image storage if scarce).

See [zram-config](https://github.com/ecdye/zram-config), it will create an OverlayFS mount for local directory to RAM (compressed zram) and sync back to the host filesystem on start/stop, there is some mention of writes but seems to lack documentation on the subject. Mount options don't seem to be configurable, but you could perhaps [modify the source](https://github.com/ecdye/zram-config/blob/63132dd913c44a1d265b552c89eabe3c605e73c9/zram-config#L77) or request support for it by raising an issue?

[zram-generator](https://github.com/systemd/zram-generator/issues/89) (_mount options support not yet supported, but there is a recent PR_) might make that easier for you to setup and manage.

---

If you prefer a shell script approach, you could probably just write all files to a text file (1 path per line) and have a timestamp before each file path entry.

Then grep the filepath from that file and update the timestamp with sed each time you would have modified the actual file metadata. For finding images to remove you'd just pipe that file list with timestamps into `sort -u` and truncate the N items for filepaths to remove.

Perhaps rather than the approach in your PR, you'd get more approval by reworking it to be a hook that can be used to call custom logic like you want, be that python or a shell script command.

heynemann:
Closing this PR since we had no activity for a while in the interest of keeping our PR list leaner. If this is something we still want to pursue, please reopen and I'll look into it. Thanks!

mpdude:
@heynemann Yes, I do. But since you wrote

> TBH I don't think this should be merged. 

… I spent no more time on this. If you think this has any chance of getting merged let me know!

heynemann:
I still think it's super easy to just have a different result storage that does it and feature that as a community plugin. WDYT?

PR #1350, opened 2021-03-17
Preserves image metadata (exif, iptc, xmp and comment)

luizender:
Added a way to preserves image metadata (exif, iptc, xmp and comment) in Pillow engine using the exiv2. If there is exiv2 in the environment and the configuration PRESERVE_IMAGE_METADATA is enabled, the image metadata read in the load method will be copy to the new image generated. So, the result of read method will be the buffer of ImageMetadata.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -286,7 +286,7 @@ class Engine(BaseEngine):
         if METADATA_AVAILABLE and self.context.config.PRESERVE_IMAGE_METADATA:
             results_metadata = ImageMetadata.from_buffer(results)
             results_metadata.read()
-            self.metadata.copy(results_metadata)
+            self.metadata.copy(results_metadata, exif=False)
             results_metadata.write()
             results = results_metadata.buffer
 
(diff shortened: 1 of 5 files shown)


thumbor/engines/pil.py
```diff
@@ -275,6 +282,14 @@ def read(self, extension=None, quality=None):  # NOQA
 
         results = img_buffer.getvalue()
         img_buffer.close()
+
+        if METADATA_AVAILABLE and self.context.config.PRESERVE_IMAGE_METADATA:
+            results_metadata = ImageMetadata.from_buffer(results)
+            results_metadata.read()
+            self.metadata.copy(results_metadata)
```
kkopachev, line 289: I am afraid it's a bit more complex than that. 
If image was rotated according to exif, we should change that in exif written here, otherwise image will have incorrect orientation.
If request called for strip_icc, this code should also remove icc profile.

luizender: I can disable the copy of exif (by passing exif=False in ImageMetadata copy method). So the exif will be the same that was setting in the image object.

In my tests, the icc_profile will be keep from image object because de ImageMetadata does not handle icc profile.

heynemann:
Closing this PR since we had no activity for a while in the interest of keeping our PR list leaner. If this is something we still want to pursue, please reopen and I'll look into it. Thanks!

PR #1376, opened 2021-11-24
Updated min pillow version

heynemann:
Just modernizing min pillow version.


cristiandean (approved):

heynemann:
Thanks @cristiandean! Great to have you with us!!!

PR #1377, opened 2021-11-24
fix(gifv): Adding webm required bitrate parameter

guilhermef:
The bitrate parameter was missing when converting Gif to Webm.


heynemann (approved):
Thanks for the catch!

PR #1386, opened 2022-01-21
thumbor.conf, black it!

marcelometal:


thumbor/thumbor.conf
```diff
@@ -1,260 +1,221 @@
...
 
-# the max width of the resized image
-# use 0 for no max width
-# if the original image is larger than MAX_WIDTH x MAX_HEIGHT,
-# it is proportionally resized to MAX_WIDTH x MAX_HEIGHT
+## The domains that can have their images resized use an empty list for allow
+## all sources
+# ALLOWED_SOURCES = ["mydomain.com"]
+
+## The max width of the resized image # use 0 for no max width if the original
```
mvhirsch, line 19: Please re-read the comments you squashed. This sentence makes no sense any more (since previously they were multiple sentences).

thumbor/thumbor.conf
```diff
@@ -1,260 +1,221 @@
...
+
+
+## This configuration indicates whether thumbor should use a custom error
+## handler. Defaults to: False
+# USE_CUSTOM_ERROR_HANDLING=False
+# ERROR_HANDLER_MODULE="thumbor.error_handlers.sentry"
+# SENTRY_DSN_URL=""
+# SENTRY_ENVIRONMENT=None
+
+FILTERS = BUILTIN_FILTERS
```
mvhirsch, line 174: Not sure if `FILTERS` should be set explicitly. Currently it automatically would load "all by default" anyway.

Does the `thumbor-doctor` depend on this, too?

@heynemann what do you think?

thumbor/thumbor.conf
```diff
@@ -1,260 +1,221 @@
...
-#   # Add your handler list modules here
+#    # Add your handler list modules here
 # ]
 
-#JPEGTRAN_PATH = '/usr/local/bin/jpegtran'
+# JPEGTRAN_PATH = "/usr/local/bin/jpegtran"
 
-# If the CurlAsyncHTTPClient should be used
-# Defaults to: False
+## If the CurlAsyncHTTPClient should be used Defaults to: False
```
mvhirsch, line 196: For a quick read, defaults are now much harder to read. Please consider using a newline for them.

mvhirsch (changes requested):
I like your approach for cleaning up the thumbor.conf file, but please keep in mind that the documentation in there needs to be scanned quickly (especially the "defaults").

thumbor/thumbor.conf
heynemann: Agreed. Thumbor.conf is a documentation source for many users.

thumbor/thumbor.conf
heynemann:
```suggestion
# FILTERS = BUILTIN_FILTERS
```

I like that now there's just the BUILTIN_FILTERS, but it could be commented out as you suggest.

marcelometal:
Hi @mvhirsch, thank you for the quick review!

mvhirsch:
LGTM :+1:

PR #1389, opened 2022-01-22
Updating cairosvg

heynemann:
If anyone has concerns over updating cairosvg, please let's discuss them here.


mvhirsch:
LGTM :+1:

PR #1390, opened 2022-01-22
Updated pillow version due to security issues

heynemann:
Given pillow <9 has a security problem, let's take the opportunity and update pillow to the most recent version.


heynemann:
Hmmm seems like we'll have to drop python 3.6 for that. Seems like a sensible choice anyway.

mvhirsch:
> Hmmm seems like we'll have to drop python 3.6 for that. Seems like a sensible choice anyway.

Support for python 3.6 has ended last year. I too think this is a sensible choice.
https://www.python.org/dev/peps/pep-0494/#lifespan

heynemann:
Maybe @kkopachev can help here? Any objections to moving to new pillow?

kkopachev (approved):
I think it's fine. We might want to release minor version for this change?

heynemann:
Yep! 👍 Thanks!

marcelometal:
https://github.com/thumbor/remotecv/blob/master/setup.py#L39
 https://github.com/thumbor/remotecv/pull/34

Issue #1384, opened 2022-01-18
Thumbor 7.0.0 upsizing all JPEG images to MAX_WIDTH/MAX_HEIGHT


bjaworski3:
### Thumbor request URL
`unsafe/filters:format(jpeg)/https://raw.githubusercontent.com/thumbor/thumbor/master/docs/images/blur_before.jpg`

### Expected behaviour
Thumbor 7.0.0 should retain the original image size and convert it to a jpeg.

### Actual behaviour
Thumbor is instead upsizing the image up to the MAX_WIDTH or MAX_HEIGHT set when there isn't a specified target size.

MAX_WIDTH and MAX_HEIGHT are set to 5000 on my server so in the above image example it returns an image of 3326x5000. Only seeing this issue after upgrading to thumbor 7.0.0 with no changes to config. On Thumbor 6.7.5 it retains the original source image's dimensions

When MAX_WIDTH and MAX_HEIGHT are not set at all Thumbor 7.0.0 will retain the source image's dimensions properly

### Operating system
Debian 11

### Your thumbor.conf
Only included settings changed from default:

```
MAX_WIDTH = 5000
MAX_HEIGHT = 5000
AUTO_WEBP = True
MAX_AGE = 315360000
ENGINE_THREADPOOL_SIZE = 2
STORAGE = 'thumbor.storages.no_storage'

# Tried with optimizer on and off and made no difference
OPTIMIZERS = [
    'thumbor.optimizers.jpegtran'
]

USE_CUSTOM_ERROR_HANDLING = True
ERROR_HANDLER_MODULE = 'thumbor.error_handlers.sentry'
SENTRY_DSN_URL = redacted
```

heynemann:
Working on it now.

heynemann:
Hey, I tried to reproduce both by running thumbor 7 as well as added a unit test to the suite that proves this issue and in both thumbor worked as expected. Can you help me dive further? Maybe provide the images you are using to test? The commit with the unit test is here: https://github.com/thumbor/thumbor/commit/5d733892bcc33bd0123d507c59bda7b13b0fed10

mvhirsch:
I've used @bjaworski3 example as input: https://raw.githubusercontent.com/thumbor/thumbor/master/docs/images/blur_before.jpg

[...]

PR #1391, opened 2022-01-22
Bug with normalize during refactor in thumbor 7

heynemann:
Maybe @kkopachev remembers?


heynemann:
Relates to #1384

heynemann:
I think this may be an ancient bug as this code is from 2011. 😢

heynemann:
Going to merge as I have to release a new version since this is a high-pri bug.

PR #1393, opened 2022-01-23
Smarter thumbor doctor

heynemann:
Now users can pass their thumbor.conf file in. If thumbor.conf has been passed in we validate some of the modules, like storage, result storage and loader. We can definitely add more as time goes by.

An example of output with an invalid filter in thumbor.conf can be seen in the test snapshot at tests/snapshots/snap_test_doctor.py.


README.mkd
```diff
@@ -93,7 +93,17 @@ If you experience any troubles, try running:
...
 ```
 
-If you still need help, please [raise an issue](https://github.com/thumbor/thumbor/issues).
+If you have a `thumbor.conf` file, you can use that to help thumbor-doctor:
+
+```bash
+thumbor-doctor -c thumbor.conf
+```
+
+If you still need help, please [raise an issue](https://github.com/thumbor/thumbor/issues). Remember to send your `thumbor-doctor` output in the issue:
```
mvhirsch, line 102: This should be documented in the github issue template instead.

thumbor/doctor.py
```diff
@@ -83,37 +158,63 @@ def check_compiled_extensions():
...
-        (
-            "cv2",
-            "Thumbor requires OpenCV for smart cropping. "
-            "For more information check https://opencv.org/.",
-        ),
-        (
-            "pyexiv2",
-            "Thumbor uses exiv2 for reading image metadata. "
-            "For more information check https://python3-exiv2.readthedocs.io/en/latest/.",
+            (
```
mvhirsch, line 195: I'm no python dev, but curious: why is `pycurl` formatted as is, and the next module `cairosvg` not the same?

thumbor/doctor.py
```diff
@@ -123,93 +224,169 @@ def check_modules():
...
+    warn_modules = [
+        (
+            "pyexiv2",
+            (
+                "Thumbor uses exiv2 for reading image metadata.\n"
+                "Don't worry. If you don't have exiv2 installed, "
+                "Thumbor will still work.\nIt just means your images won't "
+                "have EXIF metadata stored properly.\n"
+                "For more information check "
+                "https://python3-exiv2.readthedocs.io/en/latest/."
```
mvhirsch, line 238: Please be simple:

```suggestion
                "If you do not need EXIF metadata, you can safely ignore this.\n"
                "For more information visit "
                "https://python3-exiv2.readthedocs.io/en/latest/."
```

README.mkd
heynemann: Maybe in both?

thumbor/doctor.py
heynemann: Not sure either. Fixing anyway... Good catch!

thumbor/doctor.py
heynemann: Thanks!

heynemann:
Now I'm happy with the output of thumbor-doctor when a config file is present. Definitely helps users in need.

heynemann:
Next in my plate for doctor is to allow extensions to also add checks :) That would be awesome!

PR #1394, opened 2022-01-23
Added pre-commit

marcelometal:


Makefile
```diff
@@ -2,6 +2,7 @@ PYTHON = python
 .PHONY: docs build perf
 
 OS := $(shell uname)
+HAS_GIT := $(shell command -v git 2> /dev/null)
```
heynemann, line 5: Instead of doing this magic :) let's add precommit to setup.py in the dev dependencies. I assume that if you are developing thumbor you should abide by these rules, right?

heynemann: And there's no need to install for every user anyway, right?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
...
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.1.0
+  hooks:
+  - id: trailing-whitespace
+  - id: end-of-file-fixer
+  - id: check-yaml
+- repo: https://github.com/psf/black
+  rev: 21.12b0
+  hooks:
+  - id: black
```
heynemann, line 11: Please add the following:

```
  - repo: https://github.com/pycqa/flake8
    rev: ''  # pick a git hash / tag to point to
    hooks:
      - id: flake8
  - repo: https://github.com/pycqa/pylint
    rev: pylint-2.6.0
    hooks:
      - id: pylint
```

heynemann (approved):
Please perform the requested changes before merging.

heynemann:
Did the same for thumbor-aws in https://github.com/thumbor/thumbor-aws/commit/9f8a6598e308f10b9ca543e3b2b92ae43c6d8c13

heynemann:
Since this is WIP I converted it to draft @marcelometal :) Hope you don't mind.

.pre-commit-config.yaml
marcelometal: ok, I will add it

Makefile
marcelometal: There is a error when running `pre-commit install`, because there is no .git folder or git command in CI, for that reason I did this magic =)

marcelometal: is there another better way to install hooks?

heynemann: I was actually looking into a way to check whether a commit was run by pre-commit. There's a pre-sign pre-commit plugin that embeds in the commit a signature that proves that the user has actually run pre-commit. We could check this in PRs. I'm not sure we can enforce people running pre-commit anyway.  WDYT?

marcelometal: Instead of run `pre-commit install` in `make setup` we can show a message asking user to run it, what do you think?!

heynemann: I like it :)

scorphus (approved):
LGTM 👍 

Thanks, @marcelometal

PR #1395, opened 2022-01-24
Remove SQS detection support

cristiandean:
Do we already have this feature in another project? If not, I think it would be a good idea to migrate it to [this one](https://github.com/thumbor/thumbor-aws/). 

I can do that if needed.

P.S. Do you think it would be better to mark the detector as @deprecated? So we could remove it after a while.


heynemann:
Yes and yes :) thanks for helping out!

heynemann:
I mean yes let's remove from thumbor into thumbor-AWS.

heynemann:
Hey @cristiandean I think before removing we need to add it to thumbor-aws. What do you think?

cristiandean:
@heynemann  Yeah, definitely! I'm working on this.

PR #1397, opened 2022-01-25
doctor: fix errors when there is no config

marcelometal:


heynemann:
Hey good catch! Do you mind adding a test for when there's no config? I missed that one. There's a test file that tests doctor outputs.

heynemann (approved):
As mentioned above, pending only adding a unit test.

heynemann:
Gotta fix another bug :( Sorry. There's a sys.exit if no errors found and that sys.exit should also respect the exit flag.

heynemann:
Just need to update the snapshot and you should be good to go :)

heynemann:
You force-pushed over my change :(

heynemann:
Just remove the sys.exit from the print message method.

marcelometal:
> You force-pushed over my change :(

I'm sorry!

> Just need to update the snapshot and you should be good to go :)

Did you add `--snapshot-update` in `make test`?

heynemann:
Nope and you shouldn't, otherwise tests will never break. The idea of snapshots is that you look at the snapshot and it is what you expect, then it never changes (unless the functionality changes). This ensures the result is always the same.

Issue #1380, opened 2021-12-16
allowed_sources being ignored


luison:
Having issues with allowed_domains ignoring our configuration and returning images from any unauthorised url.

### Expected behaviour
https://our.thumborserver.org/t/1925x800/smart/static.externaldomain.com/img/arc_25699_g.jpg
(not real url)
should not be returning an image

### Actual behaviour
any domain seems accepted by thumbor even if not included in allowed_sources

ALLOWED_SOURCES = ['.+.domain1.com', '.+.domain2.net', '.+.domain3.es']

### Operating system
thumbor vo 6.5 on LXC debian container behind a nginx reverse proxy

### Your thumbor.conf
Only including non-default public values:
```
MAX_WIDTH = 2400
MAX_HEIGHT = 3000
MAX_PIXELS = 35000000.0

#ALLOWED_SOURCES = #    [
#    ]
ALLOWED_SOURCES = ['.+.domain1.com', '.+.domain2.net', '.+.domain3.es']
QUALITY = 75

AUTO_WEBP = True

RESULT_STORAGE = 'thumbor.result_storages.file_storage'

LOADER = 'thumbor.loaders.file_loader_http_fallback'

FILE_LOADER_ROOT_PATH = '/home/webs-static/thumbor'


STORAGE_EXPIRATION_SECONDS = 300

FILE_STORAGE_ROOT_PATH = '/cache/thumbor'


DETECTORS = ["thumbor.detectors.feature_detector"]

OPTIMIZERS = [
  'thumbor.optimizers.jpegtran'
]

## Path for the jpegtran binary
## Defaults to: '/usr/bin/jpegtran'
JPEGTRAN_PATH = '/usr/bin/jpegtran'

FILTERS =     [
        'thumbor.filters.autojpg',
        'thumbor.filters.brightness',
        'thumbor.filters.colorize',
        'thumbor.filters.contrast',
        'thumbor.filters.rgb',
        'thumbor.filters.round_corner',
        'thumbor.filters.quality',
        'thumbor.filters.noise',
        'thumbor.filters.watermark',
        'thumbor.filters.equalize',
        'thumbor.filters.fill',
        'thumbor.filters.sharpen',
        'thumbor.filters.strip_exif',
        'thumbor.filters.strip_icc',
        'thumbor.filters.frame',
        'thumbor.filters.grayscale',
        'thumbor.filters.rotate',
        'thumbor.filters.format',
        'thumbor.filters.max_bytes',
[...]

PR #1399, opened 2022-01-26
Allowed sources work in file_loader_http_fallback

heynemann:
This commit fixes a bug with the
file_loader_http_fallback loader in that it
would not validate the requests being made


heynemann:
Tested and fixed. Will release version soon. Hopefully fixes #1380

PR #1400, opened 2022-01-27
Feature/optional opencv

heynemann:
Now OpenCV is optional and detectors are smart to
skip if cv2 could not be imported.
Also refactored face detector a bit to make it more
maintainable.
Now thumbor can be installed with
pip install thumbor
pip install thumbor[all]
pip install thumbor[opencv]
pip install thumbor[tests]

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -62,7 +62,14 @@ Decide which installation option you want to use.
 ### Option 1: pip
 
 ```bash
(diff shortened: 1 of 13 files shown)


heynemann:
Just removed circleci from our build. Doesn't make any sense to keep using it.

PR #1402, opened 2022-01-29
Update Contributing.md with informations about how to use virtualenv

nicolasgabrielt:
I add some basic information how to install and use virtualenv, because I assume that many developers may never have used virtualenv, so it will enable them to install and use it in the project in a fast and simple way.


heynemann:
Great work @nicolasgabrielt! We're off to a great start. Virtualenv was required to be installed as a dependency, like you showed but starting with python 3.3 it is now part of the framework: https://docs.python.org/pt-br/3/library/venv.html

Since thumbor only supports Python 3.7+ we can just assume it is there, so you can just remove the parts to check it's there, where we ask people to use pip install and update the part where it says `virtualenv <my_env_name>` to read `python3 -m venv /path/to/new/virtual/environment` (exactly like in Python docs).

After you change that we can squash and merge!

nicolasgabrielt:
@heynemann thanks for the feedback. I already update the doc following your feedback.

CONTRIBUTING.md
```diff
@@ -24,13 +24,32 @@ There are a few things you'll need in order to properly start hacking on it.
...
 to "turn them on".
 
-Other than that, you'll also need [redis-server](https://redis.io)
-installed (for queued detector unit tests).
+The project requires Python 3.7+, and in this version virtualenv is already installed by default, to create a virtual enviroment follow the next steps:
+
+
+1. Create a virtual enviroment
+```
+$ python3 -m venv /path/to/new/virtual/environment
```
heynemann, line 40: Maybe to help people out, just use the same path in both commands?
I would change the path to "~/.venv/<my_env_name>" as otherwise virtualenv will create the folder in the current directory.
WDYT?

heynemann (approved):

CONTRIBUTING.md
nicolasgabrielt: Yeah, I agree i will change that. Thanks @heynemann

CONTRIBUTING.md
```diff
@@ -24,13 +24,32 @@ There are a few things you'll need in order to properly start hacking on it.
...
 [virtualenv](http://pypi.python.org/pypi/virtualenv) since it will keep
 your environment clean of thumbor's dependencies and you can choose when
 to "turn them on".
 
-Other than that, you'll also need [redis-server](https://redis.io)
-installed (for queued detector unit tests).
+The project requires Python 3.7+, and in this version virtualenv is already installed by default, to create a virtual enviroment follow the next steps:
+
+
+1. Create a virtual enviroment, in the folder .venv, located in the root
```
heynemann, line 38: Last nit:

```suggestion
1. Create a virtual enviroment, in the folder .venv, located in the user's home folder
```

heynemann: Btw nit = nitpick, usually what reviewers will tell you that's not a requirement for merging, but a good idea to change.

heynemann (approved):

heynemann:
As soon as you change the root wording, and the workflows pass I'll merge! Thank you very much for this!!!

nicolasgabrielt:
> As soon as you change the root wording, and the workflows pass I'll merge! Thank you very much for this!!!

It is done, thanks for all feedbacks.

heynemann:
Just a reminder Nicolas, configure your git to have your name and email - https://git-scm.com/book/pt-br/v2/Come%C3%A7ando-Configura%C3%A7%C3%A3o-Inicial-do-Git

heynemann:
Other than that 🎉! Your first thumbor contrib!!!

PR #1404, opened 2022-01-31
Changing pyexiv2

heynemann:
Let's see if that can be done. Would make our docker image much smaller and not depend on boost.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ EXTRA_LIBS_REQUIREMENTS = [
     # Going to update in a proper commit
     "cairosvg>=2.5.2",
     "pycurl==7.*,>=7.43.0",
-    "py3exiv2==0.*,>=0.7.1,!=0.7.2,!=0.8.0,!=0.9.3",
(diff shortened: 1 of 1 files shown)


marcelometal:
there are some `py3exiv2` references

https://github.com/thumbor/thumbor/blob/f73aea633da026700004058bc35951320b020277/docs/metadata.rst

heynemann:
Yeah gotta change the docs, but first I wanted to make sure tests are passing with the new pyexiv2... Seems to be the same thing.

kkopachev:
metadata test was skipped because it could not import `ImageMetadata` class: https://github.com/thumbor/thumbor/runs/5008775364?check_suite_focus=true#step:7:239.
Not so easy :(

heynemann:
That's odd, localy it's not skipping and apparently it exists in pyexiv2 https://github.com/LeoHsiao1/pyexiv2/blob/master/docs/Tutorial.md

kkopachev:
@heynemann its ImageData vs Imagemetadata ne difference

PR #1405, opened 2022-02-03
[Bootcamp Task]  Add static typing to point.py (FocalPoint class)

nicolasgabrielt:
**Issue related**:
[Implement static typing for thumbor/point.py
](https://github.com/thumbor/thumbor-bootcamp/issues/8)

**Documentation Related**:
I took some decisions for give the static type that it was based in the documentations bellow:

- [Class Objects](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-type-of-class-objects) In the ```@classmethods``` for the ```cls``` arguments I used this reference.

- [Number type](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower) As some arguments can be float and int, following the documentation I could use ```Union[int,float]``` , but I was reading the [PEP 484 Type Hints](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower) and they specifically states that: 
> ""Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.""

**Problems that I encounter during the task**:
-  [make test gives an error](https://github.com/thumbor/thumbor-bootcamp/issues/24)
- [make setup gives an error](https://github.com/thumbor/thumbor-bootcamp/issues/23)


thumbor/point.py
```diff
@@ -19,7 +23,7 @@ class FocalPoint:
         "bottom": 1.0,
     }
 
-    def to_dict(self):
+    def to_dict(self: T) -> dict:
```
heynemann, line 26: No need to type `self` and `cls` as they are implicitly typed. :)

thumbor/point.py
```diff
@@ -30,7 +34,7 @@ def to_dict(self):
         }
 
     @classmethod
-    def from_dict(cls, values):
+    def from_dict(cls: Type[T], values: dict) -> T:
```
heynemann, line 37: Here you can use `-> FocalPoint`.

thumbor/point.py
```diff
@@ -9,6 +9,10 @@
 # Copyright (c) 2011 globo.com thumbor@googlegroups.com
 
 
+from typing import TypeVar, Type
+T = TypeVar('T', bound='FocalPoint')  # pylint: disable=invalid-name
```
heynemann, line 13: No need for this as well...

heynemann (changes requested):

heynemann:
The build is failing in the CI due to linting. In your change you moved some of the #pylint disable lines. Run `make flake pylint` to check the problems.

thumbor/point.py
nicolasgabrielt: When I use FocalPoint I get this error message:
```
/thumbor/point.py:62:10: F821 undefined name 'FocalPoint'
```

Should I ignore it too ?

nicolasgabrielt:
> The build is failing in the CI due to linting. In your change you moved some of the #pylint disable lines. Run `make flake pylint` to check the problems.

I was running it, in my machine was not getting any error, but when I send it github shows me this errors.

heynemann:
Hey @nicolasgabrielt seems you're almost there, just missing lint errors. How can we help?

heynemann:
You can just use `# pylint:disable=C0103` in the two lines where the lint error is happening...

PR #1411, opened 2022-02-12
ext/filters/_convolution: Apply convolution to alpha channel too

scorphus:
Not applying convolution to the alpha channel of images with transparency results in an image where the colors are blurred but the alpha mask is not. This change adds the alpha channel to the mix – when it exists – fixinng the issue.

Both `blur` and `fill` filters are affected, hence the new tests and fixtures.

ps.: the new test case for `fill` reflects one of the usage examples in [the docs](https://thumbor.readthedocs.io/en/latest/filling.html#example-2).


thumbor/ext/filters/_convolution.c
```diff
@@ -88,12 +89,18 @@ _convolution_apply(PyObject *self, PyObject *args)
                     sum_r += copy_buffer[tmp_idx + r_idx] * kernel_value;
                     sum_g += copy_buffer[tmp_idx + g_idx] * kernel_value;
                     sum_b += copy_buffer[tmp_idx + b_idx] * kernel_value;
+                    if (num_bytes > 3) {
```
scorphus, line 92:
```suggestion
                    // Change alpha channel only when available
                    if (num_bytes > 3) {
```

scorphus (commented):
Hey, @heynemann! Thanks for your input. I followed your hint and added a comment to make the purpose of that line (and the one below) a bit clearer. Please LMK if it can be done any better.

thumbor/ext/filters/_convolution.c
heynemann: LGTM!

heynemann (approved):

PR #1423, opened 2022-03-14
Moving gifv to thumbor-plugins

guilhermef:
The Gifv code was moved to https://github.com/thumbor/thumbor-plugins/blob/master/thumbor_plugins/optimizers/gifv/__init__.py

We can remove the dependency on the next major release, and install it by default to keep it compatible with the current release.


thumbor/optimizers/gifv.py
```diff
@@ -8,114 +8,15 @@
...
-            return webcolors.normalize_hex(color_string)
-        except ValueError:
-            pass
-
-        if color_string:
-            logger.error("background_color value could not be parsed")
-
-        return None
+class Optimizer(gifv.Optimizer):
+    logger.warning("""
```
scorphus, line 17:
```suggestion
    logger.warning(
        """
```

thumbor/optimizers/gifv.py
```diff
@@ -8,114 +8,15 @@
...
+    This optimizer will be deprecated on the next major version.
+    Add thumbor-plugins-gifv as a dependency, and
+    change thumbor.optimizers.gifv to thumbor_plugins.optimizers.gifv
+    on the OPTIMIZERS list on you thumbor.conf.
+    """)
```
scorphus, line 22:
```suggestion
    """
    )
```

scorphus: This code doesn't look to me like it was *blackified*. We were *blackifying* all code, IIRC. I'd say we just accidentaly stopped checking it as part of the workflow. If that's really the case, I will happily fix it in a PR.

scorphus (approved):
Good job, @guilhermef, thanks!

LGTM, except for the comments below.

PR #1427, opened 2022-03-27
Replace 'colorful' with 'colorama'

marcelometal:
We don't need two or more color packages, so I'm replacing `colorful` with `colorama`. 
I'm choosing `colorama` because there is a Debian package for it.

### DIFF
--- a/thumbor/doctor.py
+++ b/thumbor/doctor.py
@@ -17,7 +17,7 @@ from os.path import abspath
 from shutil import which
 from typing import List
 
-import colorful as cf
+from colorama import Fore, init
 
 from thumbor import __release_date__, __version__
 from thumbor.config import Config
@@ -58,13 +58,29 @@ def get_options():
     }
 
 
-def header(msg, color=cf.yellow):
-    print(color(msg))
+def header(msg):
+    print(f"{Fore.YELLOW}{msg}")
 
 
-def subheader(msg, color=cf.bold_coral):
-    print(color(msg))
+def subheader(msg):
+    print(f"{Fore.MAGENTA}{msg}")
     newline()
+
+
+def print_error(msg):
+    print(f"{Fore.RED}{msg}")
+
+
+def print_success(msg):
+    print(f"{Fore.GREEN}{msg}")
+
+
+def print_warning(msg):
+    print(f"{Fore.YELLOW}{msg}")
+
+
+def print_info(msg):
+    print(f"{Fore.CYAN}{msg}")
 
 
 def newline():
@@ -147,9 +163,9 @@ def check_extensibility_modules(cfg):
         for module in modules:
             try:
                 import_module(module)
-                print(cf.bold_green(f"{CHECK} {module}"))
+                print_success(f"{CHECK} {module}")
             except ImportError as error:
-                print(cf.bold_red(f"{CROSS} {module} - {error_message}"))
+                print_error(f"{CROSS} {module} - {error_message}")
                 errors.append(
                     format_error(
                         module,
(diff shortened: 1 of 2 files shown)


scorphus (approved):
So far so good 👍

scorphus:
May I suggest rebasing on top of current master and see if the checks become green?

marcelometal:
> May I suggest rebasing on top of current master and see if the checks become green?

Nops...  But I will fix it soon

```
thumbor/doctor.py:420:8: W0[6](https://github.com/thumbor/thumbor/runs/5778889615?check_suite_focus=true#step:11:6)03: Using the global statement (global-statement)
thumbor/doctor.py:420:8: C0103: Constant name "Fore" doesn't conform to UPPER_CASE naming style (invalid-name)
```

thumbor/doctor.py
```diff
@@ -405,9 +425,10 @@ def load_config(config_path):
 
 
 def configure_colors(nocolor):
-    cf.use_style("solarized")
+    init(autoreset=True)
```
scorphus, line 428:
```suggestion
```

thumbor/doctor.py
```diff
@@ -405,9 +425,10 @@ def load_config(config_path):
...
     if nocolor:
-        cf.disable()
+        global SHOULD_USE_COLOR  # pylint: disable=global-statement
+        SHOULD_USE_COLOR = False
```
scorphus, line 431:
```suggestion
        init(strip=True, convert=False)
        init(strip=True, convert=False)
```

thumbor/doctor.py
```diff
@@ -29,6 +29,8 @@
 ERROR = "⛔"
 QUESTION = "❓"
 
+SHOULD_USE_COLOR = True
+
```
scorphus, line 33:
```suggestion
```

thumbor/doctor.py
```diff
@@ -58,15 +60,35 @@ def get_options():
     }
 
 
-def header(msg, color=cf.yellow):
-    print(color(msg))
+def print_message(color, msg):
+    print(f"{color}{msg}" if SHOULD_USE_COLOR else msg)
```
scorphus, line 64:
```suggestion
    print(f"{color}{msg}")
```

thumbor/doctor.py
scorphus, line 431:
```suggestion
        init(strip=True, convert=False)
    else:
        init(autoreset=True)
```

scorphus (commented):
Let me suggest another approach for turning colors off. What do you think, @marcelometal?

scorphus:
Here's the code snippet that helped me play with `colorama`'s `init`:

```python
from colorama import Back, Fore, Style, init

# init(autoreset=True)
init(strip=True, convert=False)

print(Fore.RED + 'some red text')
print(Back.GREEN + 'and with a green background')
print(Style.DIM + 'and in dim text')
print(Style.RESET_ALL)
print('back to normal now')
```

marcelometal:
> Let me suggest another approach for turning colors off. What do you think, @marcelometal?

Nice! code updated, thanks!

thumbor/doctor.py
```diff
@@ -58,15 +58,35 @@ def get_options():
     }
 
 
-def header(msg, color=cf.yellow):
-    print(color(msg))
+def print_message(color, msg):
+    print(f"{color}{msg}")
```
scorphus, line 62: Now this is a nit: perhaps now we can get rid of this method. That means one less function call. One less function in the code... Would you see any benefit in maintaining it?

scorphus (approved):

thumbor/doctor.py
marcelometal: done!

scorphus (approved):
🤘 :shipit:

PR #1430, opened 2022-04-01
Improving URL composer

marcelometal:


thumbor/url_composer.py
```diff
@@ -220,6 +220,8 @@ def main(arguments=None):
     """Converts a given url with the specified arguments."""
 
     parsed_options, arguments = get_options(arguments)
+    if not arguments:
+        sys.exit(1)
```
heynemann, line 224: Maybe print a message saying arguments are required?

heynemann (approved):

thumbor/url_composer.py
marcelometal: There is a message in [get_options()](https://github.com/thumbor/thumbor/blob/master/thumbor/url_composer.py#L163), I will move it to `main()`
```
Error: The image argument is mandatory. For more information type thumbor-url -h
```

thumbor/url_composer.py
```diff
@@ -221,6 +218,12 @@ def main(arguments=None):
 
     parsed_options, arguments = get_options(arguments)
 
+    if not arguments:
+        sys.stdout.write(
+            "Error: The image argument is mandatory. For more information type thumbor-url -h\n"
+        )
+        return sys.exit(1)
```
scorphus, line 225: `sys.exit` raises an exception, I deem `return` unnecessary — and kind of unreadable, as it may trick the reader into trying to understand what is being returned there and why is it the case. WDYT?

```suggestion
        sys.exit(1)
```

thumbor/url_composer.py
```diff
@@ -234,7 +237,7 @@ def main(arguments=None):
             "Error: The -k or --key argument is mandatory."
             " For more information type thumbor-url -h\n"
         )
-        return None
+        return sys.exit(1)
```
scorphus, line 240: Dito.
```suggestion
        sys.exit(1)
```

thumbor/url_composer.py
```diff
@@ -244,8 +247,7 @@ def main(arguments=None):
     url = crypto.generate(**thumbor_params)
     sys.stdout.write("URL:\n")
     sys.stdout.write(f"{url}\n")
-
-    return url
+    return None
```
scorphus, line 250: This line is also unnecessary.
```suggestion
```

scorphus (approved):
LGTM, with some remarks. Please consider them.

PR #1438, opened 2022-05-12
Skip test from #1435 for python3.7 version

RaphaelVRossi:
Need to skip test from #1435 because on python3.7 doesn't work with AsyncMock.

See [unittest.mock.patch decorator doesn't work with async functions #81177](https://github.com/python/cpython/issues/81177) for more information.


scorphus (approved):
Nice catch! 👍

Issue #1314, opened 2020-09-10
AVIF support


allantatter:
Could [AVIF](https://jakearchibald.com/2020/avif-has-landed/) support expected to be added at some point similarly to `webp` in [format](https://thumbor.readthedocs.io/en/stable/format.html#format) filters?

franek:
Hi,
We would really appreciate that Thumbor does support AVIF format. 
* How could we help to achieve that ? 
* What are the requirements ? Is this PR on Pillow required https://github.com/python-pillow/Pillow/pull/4782 ? 
Thanks,

fdintino:
I opened python-pillow/Pillow#5201, which would make it possible for thumbor to support the format.

Tsouef:
Any news ?

gladson:
https://github.com/libvips/libvips

It supports a good range of image formats, including JPEG, JPEG2000, JPEG-XL, TIFF, PNG, WebP, HEIC, AVIF, FITS, Matlab, OpenEXR, PDF, SVG, HDR, PPM / PGM / PFM, CSV, GIF, Analyze, NIfTI, DeepZoom, and OpenSlide.

scorphus:
You can try the [ImagaMagick engine](https://github.com/scorphus/thumbor-wand-engine). You'll need Thumbor from master, though. Let me know if it works, please.

fdintino:
It's been taking a while to get the pillow PR merged. To be fair to the pillow developers, it's not a trivial new feature. Until that PR lands I've created a standalone package, [pillow-avif-plugin](https://github.com/fdintino/pillow-avif-plugin), that adds AVIF support to Pillow. Also, unlike my pillow PR—which only works in python 3—my standalone package has wheels built against python 2. This is because the latest stable version of thumbor is still python 2-only.

I'm currently testing out an extension of thumbor and [thumbor-video-engine](https://github.com/theatlantic/thumbor-video-engine) on https://www.theatlantic.com/ that adds auto-AVIF support. If you visit the site in chrome, you'll see a bunch of AVIF images being served in the Network tab in the web inspector. The source code for that isn't publicly available, but if it goes well I'm planning on creating a package for it.

vanchung1995:
@fdintino I have used pillow avif plugin, it worked nice. But I dont know how to apply it into thumbor. Example I want to crop image, then convert it to avif image. I read a lots but nothing found. Can you give some guide? Thanks so much

mvhirsch:
[...]

PR #1446, opened 2022-05-17
proof of concept: plugin architecture

fdintino:
This is a first draft at a plugin architecture that would allow me to build auto-AVIF support (see #1314) using an extensible framework. I'm 100% open to any suggestions for changing it...I've tried my best to avoid over-engineering it, but I'm not completely sure I've succeeded. It would of course need tests and documentation, but I'll save that until I have confidence about the general direction. I'll use the rest of this PR description to describe the mechanics. The code itself should also be somewhat expository, as I've refactored some of the thumbor core functionality into plugins (specifically: auto-png-to-jpeg, the gifsicle engine, auto-webp, and the implementation behind the max_bytes filter).

I tried to restrict myself to making as few changes to the core thumbor code as possible. Where there are changes, they should be entirely backwards-compatible.

### Anatomy of a plugin

A plugin is a module, which can optionally include any of the following classes or functions:

- `ImagingHandlerPlugin`
- `ResultStoragePlugin`
- `UploadHandlerPlugin`
- `ResourceHandlerPlugin`
- `EnginePlugin`
- `TransformerPlugin`
- `validate_config`
- `modify_context_importer`
- `modify_import_modules`

Plugins can also define configuration values as needed.

The functions are the most straightforward: any `validate_config` functions defined in plugins get called at the end of the `validate_config` function in `thumbor.server`, and  `modify_import_modules` and `modify_context_importer` get called with an Importer and ContextImporter instance, respectively. They allow for plugins to mutate those objects as needed.

The plugin classes can have special methods defined on them (designated by the use of the `@plugin_method` decorator) that allow them to hook into methods of the class they are modifying. These all take the general form:

```python
@plugin_method(target="attr_of_target_method_on_class")
def override_method(self, next_fn, instance, *args, **kwargs):
    # Add logic to run before the target method runs. The arguments can be
    # changed, or there can be a return statement, which would prevent the
    # target method from executing
    retval = next_fn(*args, **kwargs)
    # Add logic to run after the target method, optionally mutating or
    # replacing the return value
    return retval
```

The argument `next_fn` will either call the target method or the next plugin's override method (these are evaluated in the order they're defined in the `PLUGINS` setting). The `instance` argument is what would be `self` in the target method (e.g. the `Engine` or `ImagingHandler` instance).

The [`ImagingHandlerPlugin` in the auto_webp plugin](https://github.com/fdintino/thumbor/blob/93f63af178429703db077e5dc29bbe1fa4bce9fd/thumbor/plugins/auto_webp.py) is a pretty good example of what this looks like in practice.

I've created a base plugin class for every thumbor class that can be extended, with what I felt were sensible aliases for some of the methods. But plugins aren't limited to just the methods I've included on these base classes. _Any_ method on the target class can be overridden by using the `@plugin_method` decorator.

### Other changes made in this PR

Besides the logic that moved things from thumbor core into plugins, there are a few other changes made to facilitate the plugin functionality and to make auto-format-conversion more extensible.

I changed `RequestParameters.accepts_webp` to `RequestParameters.accept_formats`, so that plugins can append other values (e.g. `"avif"` for `image/avif`, `"gifv"` for `video/*`, etc.). Similarly I've added `RequestParameters.vary_formats`, which is used in `ImagingHandler` as part of the determination of whether to send a `Vary: Accepts` response header. It is also used in the result storage class as part of the path. If auto-avif and auto-webp were both enabled, then the result storage class for browsers with `Accepts: image/webp`, `Accepts: image/avif`, and `Accepts: image/avif, image/webp` would all need separate prefixes. In this PR they would be, respectively, `auto_webp`, `auto_avif`, and `auto_webp_avif`.

I made the ImagingHandler, UploadHandler, and ResourceHandler classes all overridable in settings. This was not strictly necessarily, but I don't see why they shouldn't be if the `APP_CLASS` can be overridden.


heynemann:
One suggestion I would make for this PR is that it is too big. Can we make
one thing extensible via plugins instead of all there is? I'd love to have
users try out this new plugin architecture and see its impacts on
performance before rolling out on all extension points. WDYT?

Bernardo Heynemann

fdintino:
I like that as an incremental approach. It would only reduce the diff by 300 lines, so it will still be a large PR. But it's not nothing, and definitely ought to be more manageable. Also the testing surface would be much smaller. I can move all of the actual plugin implementation code back to where it was originally, but keep this code around if and when we want to refactor them into plugins.

fdintino:
@heynemann are there any other changes you'd like me to make, or is there anything else I can do to make this easier for you to review? Thanks!

RaphaelVRossi:
@fdintino Could you rebase and resolve the conflicting above?

I'll test this PR later.

@heynemann Could you help me to review this PR later?

fdintino:
@RaphaelVRossi I'm working on a rebase. It's complicated by the auto-avif feature added in #1518. My auto-avif plugin ensured that `Vary: Accept` was added to the response and that it worked with result storage, whereas the implementation in #1518 does not. Given the holidays it may take me a few more days to finish reconciling the two implementations.

fdintino:
Bumping this so that it doesn't auto close

fdintino:
This work got away from me. There are enough changes after the rebase that I'm fine opening it as a new PR when I have the time to finish it.

PR #1439, opened 2022-05-12
Added Redis Sentinel support

RaphaelVRossi:
Added Redis Sentinel support.

To connect into sentinel cluster, you need to use these following variables:

```python
REDIS_QUEUE_MODE=sentinel
REDIS_QUEUE_SENTINEL_INSTANCES=sentinel_host1:26379,sentinel_host2:26379
# optional
REDIS_QUEUE_SENTINEL_PASSWORD=password
REDIS_QUEUE_SENTINEL_MASTER_INSTANCE=master_instance
# optional
REDIS_QUEUE_SENTINEL_MASTER_PASSWORD=password
# optional
REDIS_QUEUE_SENTINEL_SOCKET_TIMEOUT=10.0
```

To test locally you can use:

```sh
make test # will run all unit tests
# or
make redis
make unit
make kill_redis
```

Variable `REDIS_QUEUE_MODE` is optional and has a default value equals `single_node`, so if you are using a redis single node you doesn't have to change anything.

PR #1450, opened 2022-05-30
Fix library test example in docs and use python3

tchollingsworth:
The signing code has moved in thumbor and it no longer supports Python 2

```
% python3 -c 'from thumbor.crypto import Signer; signer = Signer("secret"); print(signer.signature("url"))'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'thumbor.crypto'
```
```
% python3 -c 'from libthumbor.url_signers.base64_hmac_sha1 import UrlSigner; signer = UrlSigner("secret"); print(signer.signature("url").decode("utf-8"))'
jjTzhyRdkQ3WHRCJofpLTjuYCZI=

```

### DIFF
--- a/docs/libraries.rst
+++ b/docs/libraries.rst
@@ -124,7 +124,7 @@ Here's how it was implemented in Ruby:
 
     def sign_in_thumbor(key, str)
         #bash command to call thumbor's decrypt method
-        command = "python -c 'from thumbor.crypto import Signer; signer = Signer(\"" << key << "\"); print signer.signature(\"" << str << "\")'"
+        command = "python3 -c 'from libthumbor.url_signers.base64_hmac_sha1 import UrlSigner; signer = UrlSigner(\"" << key << "\"); print(signer.signature(\"" << str << "\").decode(\"utf-8\"))'"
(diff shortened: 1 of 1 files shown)


mvhirsch:
@heynemann looks like there is a problem with the CI. Maybe a caching issue of dependencies, since Sentinel support just got added in a non-merged branch here? https://github.com/thumbor/thumbor/pull/1439

scorphus:
@mvhirsch, latest releases of `remotecv` require recent versions or `redis`. `remotecv` had to be locked in `3.1.x` but that's not the case:

https://github.com/thumbor/thumbor/blob/1784b200107225581379f06eb688bbe681d6fe64/setup.py#L36

This should be fixed as soon as the referenced PR gets merged.

mvhirsch:
> This should be fixed as soon as the referenced PR gets merged.

@scorphus the PR https://github.com/thumbor/thumbor/pull/1439 was merged. Could you please re-trigger the CI?

heynemann:
Can you rebase so it gets run again?

Issue #1455, opened 2022-06-16
Watermark url are not subject to ALLOWED_SOURCES


FelixMartel:
### Thumbor request URL
http://localhost:8888/unsafe/filters:watermark(http://domain2/foo,0,0,0)/http://domain1/bar

### Expected behaviour
Watermark request should respect ALLOWED_SOURCES.

### Actual behaviour
Arbitrary URL can be supplied and will be fetched.

### Operating system
Ubuntu

### Your thumbor.conf
ALLOWED_SOURCES = ['domain1']

devppjr:
Hi @FelixMartel, it was tested and I confirmed watermark filter doesn't respect the allowed sources configuration. I'm working on it!

PR #1457, opened 2022-06-20
added new validation in watermark filter to respect allowed sources configuration

maiagripp:
It was add a new validation and now watermarks filters respects allowed sources configuration, fixing https://github.com/thumbor/thumbor/issues/1455


RaphaelVRossi (approved):
LGTM

thumbor/filters/watermark.py
```diff
@@ -207,3 +210,20 @@ async def watermark(self, url, x, y, alpha, w_ratio=False, h_ratio=False):
...
+        if not is_valid:
+            logger.warning(
+                (
+                    'Request denied because the specified url "%s" was '
+                    "not identified by the loader as a valid url"
+                ),
+                url,
+            )
+
+        return is_valid
```
scorphus, line 229: * Nit 1: The warning message is long, especially when compared to other logged messages. Also, a request might not make sense in the context of a filter. Something like “watermark source not allowed” would make more sense, IMO.

* Nit 2: The extra variable (and its added cognitive load) makes it a bit slower (and less easy to read) than returning a boolean instead, e.g.:

  ```suggestion
          if not self.context.modules.loader.validate(self.context, url):
              logger.warning('watermark source not allowed: "%s"', url)
              return False
  
          return True
  ```
  
* Nit 3: I know, this code is duplicated from the following source:

  https://github.com/thumbor/thumbor/blob/1784b200107225581379f06eb688bbe681d6fe64/thumbor/handlers/__init__.py#L746-L761

  And it would be desirable to have the URL validation functionality in a single place.

So, I'd say we merge this as is, and solve the aforementioned nits in a future pull request.

scorphus (approved):

thumbor/filters/watermark.py
maiagripp: Thanks for your insights.
I implemented the suggestions and commited again. 
Could you please re-review?

scorphus (approved):
Thanks for contributing, @maiagripp

thumbor/filters/watermark.py
```diff
@@ -207,3 +210,13 @@ async def watermark(self, url, x, y, alpha, w_ratio=False, h_ratio=False):
...
             raise tornado.web.HTTPError(500)
+
+    def validate(self, url):
+        if not hasattr(self.context.modules.loader, "validate"):
+            return True
+
+        if not self.context.modules.loader.validate(self.context, url):
+            logger.warning('watermark source not allowed: "%s"', url)
+            return False
+
```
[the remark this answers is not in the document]
maiagripp: Thanks for the insight again! :)
I configured my editor to trim the trailing spaces.
Can you review again?

RaphaelVRossi (approved):

PR #1460, opened 2022-06-30
readme: Update Wikipedia logo

Krinkle:
The logo was redesigned in 2010. That's before this image was added to the readme, but for some reason the old logo was still used.

Ref https://en.wikipedia.org/wiki/Wikipedia_logo.

I used the file from <https://commons.wikimedia.org/wiki/File:Wikipedia-logo-v2-en.svg> and for consistency with the other files, gave it a white bg and expanded the space to a 256px square.

----

It seems the current URLs are pointing to a branch, which makes updating them tricky since updating the main branch doesn't affect the readme, and updating the branch leaves one without a preview. I suspect it was done this way as part of an experiment but then left that way. I don't know if relative paths are intentionally avoided, but it's something that works fine as far as I can tell, and seems to be what other projects do in their readme. (Within the GitHub online viewer, relative paths can be confusing since when naively interpreted they will point ot the HTML page representing the file as opposed to the image file itself; but this is a problem specific to GItHub and GitHub's own markdown processor expands these fine to the raw URL they should be; when viewing them outside GitHub or when viewing them locally in a typical markdown generator or static server, they naturally point to the actual file and also work fine; this also means PRs and diff previews will work as expected).

If desired, I can apply the same changes to the other logos as well.


mvhirsch:
@heynemann I think this can be merged. Updating to more recent Wikipedia Logo seems fine to me!

mvhirsch (approved):

PR #1466, opened 2022-07-18
aarch64 docker image build scripts added

kevinsingapore:
**key points**

1 aarch64 Dockerfile added;

2 thmbor conf file and docker scripts added.


mvhirsch:
Thanks for contributing @kevinsingapore. Currently there are [many Docker related thumbor repositories](https://github.com/thumbor/awesome-thumbor#hosting-thumbor). 

`thumbor/thumbor` concentrates on the core code to run the application. This PR requires to be built on an arm processor, while hard-coding version constraints, too. To be a suitable contribute either this needs much more improvement: 

- supporting x64 and arm64
- use configurable version string (build-arg?) allowing easy maintenance
- github workflow definition, so on release this will be built automatically

WDYT @heynemann ?

heynemann:
Same. I would advise you to create a new repository and we will be glad to advertise it in awesome-thumbor :) But core thumbor should be smaller, not bigger. We still need to remove some functionality here.

Issue #1472, opened 2022-08-08
Difference between crop images from Thumbor 6.0 to Thumbor 7.0


maiagripp:
### Thumbor request URL
```http://localhost:8888/unsafe/180x180/smart/filters:strip_icc()/https://user-images.githubusercontent.com/44818833/183480806-6df9d1b8-c1c9-4399-a93f-2d118c792d66.png```

Original:
![original](https://user-images.githubusercontent.com/44818833/183480806-6df9d1b8-c1c9-4399-a93f-2d118c792d66.png)

### Expected behaviour
After the migration from Thumbor 6.0 to Thumbor 7.0, we were looking for the expected response were same as actual response, but instead we get the crop a little bit different.

After these alterations:
This first alteration, forced an int 
[Point.py](https://github.com/thumbor/thumbor/commit/2f15b1550c386bfe31719c5eafc419d30dd8a4cc?diff=split#diff-fad6e661cd85308e55fb7b8f9a4e06daa4fb461e4701951c78c42ad2e2515df0)
And this alteration, added a round(). In Python3 round() don't behave the same way as in Python2
[Transformer.py](https://github.com/thumbor/thumbor/commit/2f15b1550c386bfe31719c5eafc419d30dd8a4cc?diff=split#diff-436623fd381383c31e437b38ae41349401702890e01522b010e656a2c3b18f31)
[Round Behaviour](http://python3porting.com/differences.html?highlight=round#rounding-behavior)
So the center of the images became different from expected to actual.

Expected:
![expected](https://user-images.githubusercontent.com/44818833/183480750-2a8269f4-4fff-4748-b40c-3c7460f0c91a.png)

The image is more centralized
![icon-globo](https://user-images.githubusercontent.com/44818833/183496481-8d27b1c9-e9d2-45c4-868c-2b625fd26d7d.png)

### Actual behaviour
Thumbor now is cropping more to the right side then the left side. 
Calling Thumbor by the request url above, we expected the response to be equal as the expected image.
Actual image is cropped more to the right
![icon-globo-gcp](https://user-images.githubusercontent.com/44818833/183496685-3a477a4e-6011-4ad6-89c4-a9fcb19492e4.png)

### Operating system
We are currently running Thumbor inside a docker debian with 512Mb RAM, and python version 3.9

```
cd thumbor/

docker build --platform linux/amd64 -f TestDockerfile39 -t thumbor-test-39 .

[...]

PR #1475, opened 2022-08-16
Correct difference crop between Python2 and Python3

maiagripp:
## Motive
After the migration from Thumbor 6.0 to Thumbor 7.0, we were looking for the expected response were same as actual response, but instead we get the crop a little bit different.

## Changes Proposed
Remove the forced cast to ```int``` and adding a customize ```round``` method which maintain the same behaviour as ```round``` in Python2.

## How to validate this PR
```
cd thumbor/

docker build --platform linux/amd64 -f TestDockerfile39 -t thumbor-test-39 .

docker run --rm -it --name thumbor-test --memory 512mb -p 8888:8888 --platform linux/amd64 -v "$(pwd):/app" thumbor-test-39 make setup run

docker exec -it thumbor-test bash

make unit
```

Closes https://github.com/thumbor/thumbor/issues/1472


mvhirsch:
Thank you for contributing @maiagripp!

@heynemann this looks well tested and sounds reasonable to me. Could you please review these changes?

thumbor/point.py
```diff
@@ -49,19 +49,19 @@ def __init__(
...
         self.origin = origin
 
     @classmethod
     def from_square(
         cls, x, y, width, height, origin="detection"
     ):  # pylint: disable=invalid-name
-        center_x = x + width // 2
-        center_y = y + height // 2
+        center_x = x + (width // 2)
+        center_y = y + (height // 2)
```
scorphus, line 64:
```suggestion
        center_x = x + width // 2
        center_y = y + height // 2
```

scorphus (commented):
I still need to dig into this one myself to better understand it. From a first look it seems it needs a more solid fix, if it is indeed an issue.

thumbor/point.py
maiagripp: hey @scorphus !
thanks for the suggestion. It was already implemented.

maiagripp:
@scorphus ping🙂

scorphus (commented):
I don't think there is an issue here. I think The image used here is misleading. It being so small and one pixel wider than its height — 37x36 — makes this seem like an issue.

Consider the following samples and their “croppings” by both Thumbor 6 and 7:

## Sample image of size 36x36

Both cropped images are identical. That's because the width, 36, is an even number.

### Original image

thumbor-issue-1475-36x36

### Cropped by Thumbor 6

thumbor-issue-1475-36x36-thumbor-6

### Cropped by Thumbor 7

thumbor-issue-1475-36x36-thumbor-7

## Sample image of size 37x36

The cropped images are different, as 37 is odd. Thumbor 6 crops it one pixel less from the left, Thumbor 7 from the right.

### Original image

thumbor-issue-1475-37x36

### Cropped by Thumbor 6

thumbor-issue-1475-37x36-thumbor-6

### Cropped by Thumbor 7

thumbor-issue-1475-37x36-thumbor-7

## Sample image of size 38x36

Again, both cropped images are identical.

### Original image

thumbor-issue-1475-38x36

### Cropped by Thumbor 6

thumbor-issue-1475-38x36-thumbor-6

### Cropped by Thumbor 7

thumbor-issue-1475-38x36-thumbor-7

Some could argue that Thumbor 6 and 7 should behave the same. I argue Thumbor 6 had it wrong. The image should be cropped from the right.

heynemann:
Amazing investigation Pablo!!! Thanks for that and 100% agree!!!

heynemann:
Closing this as the fix does not seem to be correct. If you still feel we need to pursue this re-open with the reasoning.

PR #1476, opened 2022-08-16
feat: Support AVIF format encoding

### LINKED ISSUE #1314: AVIF support

fdintino:
Fixes #1314 

This PR makes it possible to pass `avif` to the format filter and have it return AVIF images if [pillow-avif-plugin](https://pypi.org/project/pillow-avif-plugin/) is installed.


heynemann:
This PR looks good to me with a few additions required:
* Rebase to fix conflict
* Add a test with an image (a photo or something) to ensure we do not get regressions
* Add an integration test for it
Removing stale from it

fdintino:
My latest push is rebased off of the latest commits on the master branch. I've added a simple integration test without any mocks. Otherwise, this PR has 100% test coverage…I've tested everything I can think of. If you have something specific in mind, could you point me to an example of the sort of test you're looking for?

Thanks!

heynemann:
The integration test I was mentioning is this one https://github.com/thumbor/thumbor/blob/master/integration_tests/urls_helpers.py

This way our CI will help ensure that we don't break AVIF filtering in the future (combined with your testing, thanks for that) :) Basically you just add a new filter there `filters:format(avif)` and that should be it.

thumbor/engines/pil.py
```diff
@@ -280,6 +292,46 @@ def read(  # noqa
         if options["quality"] is None:
             options["quality"] = self.context.config.QUALITY
 
+        if ext == ".avif":
```
RaphaelVRossi, line 295: I was testing these changes locally, and I didn't install `pillow-avif-plugin` just for test purpose. 

And Thumbor returned error 500 with this log message:

```python
  File "/usr/local/lib/python3.9/dist-packages/thumbor/handlers/__init__.py", line 571, in finish_request
    result = await self.context.thread_pool.queue(
  File "/usr/local/lib/python3.9/dist-packages/thumbor/threadpool.py", line 55, in queue
    return await self._execute_in_pool(operation, *args)
  File "/usr/local/lib/python3.9/dist-packages/thumbor/threadpool.py", line 49, in _execute_in_pool
    return await loop.run_in_executor(self.pool, operation, *args)
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.9/dist-packages/thumbor/handlers/__init__.py", line 491, in _load_results
    results = context.request.engine.read(image_extension, quality)
  File "/usr/local/lib/python3.9/dist-packages/thumbor/engines/pil.py", line 213, in read
    img_buffer = BytesIO()
  File "/usr/local/lib/python3.9/dist-packages/PIL/Image.py", line 2307, in save
    save_handler = SAVE[format.upper()]
KeyError: 'AVIF'
```

Maybe here, We can validate if `HAVE_AVIF` is True, before accept AVIF extension, or change to default JPEG.

What do you think @fdintino / @heynemann ?

fdintino: oh yes, that's definitely an oversight on my part. I'll add the check and a regression test for this situation.

RaphaelVRossi:
I talked to @scorphus  and we thought of a better possibility for this change.
Instead of leaving it directly in Thumbor, we can create a plugin, using [thumbor-plugins ](https://github.com/thumbor/thumbor-plugins/blob/master/thumbor_plugins/optimizers/gifv/__init__.py) as an example. So we didn't add as much complexity to the Thumbor code and left it isolated in case someone wanted to use this format.

What do you guys think about it, @fdintino  @heynemann ?

fdintino:
We use thumbor with AVIF on www.theatlantic.com, and have been for over a year now (we actually have an auto-AVIF feature that works the same as auto-webp). And we do it only by extending thumbor. The end result is far more complex than what we have in this pull request. It requires that we write a custom engine that extends the pil engine, which becomes tricky if you want it to play nice with something like [thumbor-video-engine](https://github.com/theatlantic/thumbor-video-engine). There are limits to where a plugin can hook into the control flow, and so it often involves duplicating code from thumbor or the overriding of protected methods. I opened #1446 as a proposal for how the plugin architecture can be formalized.

The next version of Safari will have AVIF support, and Firefox and Chrome already have support for AVIF (in the case of Chrome, it's been supported for over a year). It's compression is so superior to JPEG and WebP that for many users it will be a must-have for an image optimization service. Once this is merged it will allow me to add auto-avif support to #1446 in a way that can work alongside the existing auto-webp feature (and even auto-gif-to-h264, which is also something The Atlantic has in our custom thumbor server). That strikes me as the sort of thing that would make sense as a separate plugin.

frasko23:
hello
does this feature support the fallback mechanism for google images / bot since Google does not (!) support avif format?

if google crowl the pages with avif images it need to have this fallback or the avif images are not seen by google bot/serp

PR #1482, opened 2022-09-22
Set seed for the noise filter test

scorphus:
This makes the test result consistent regardless of C implementation used in place — e.g., this test now passes in both Linux and macOS.


scorphus:
🤔

PR #1483, opened 2022-09-23
Use random and srandom instead of rand and srand

scorphus:
The C standard library offers much more robust `random` and `srandom`, which behave the same in different platforms — such as Linux and macOS.

With this change, the noise filter test stops failing on macOS.

I'm not so sure about non-POSIX counterparts, though.


heynemann (approved):

kkopachev (approved):

scorphus:
Thanks for the approvals, I overlooked this and ended up not merging it. Doing it now.

PR #1493, opened 2022-10-17
Adding docs for cover filter

maiagripp:
This PR adds documentation for the `cover()` filter.
I had to make some changes on `configuration.srt`, since this filter isn't skipped when `USE_GIFSICLE_ENGINE` is set to `TRUE`


docs/cover.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Example
+-------
+
+.. image:: images/animated.gif
+    :alt: Gif before cover filter
+
+::
+
+.. code::
```
scorphus, line 26: This line can be removed.

```suggestion
```

docs/cover.rst
```diff
@@ -0,0 +1,30 @@
+Cover
+========
+
+Usage: `cover()`
+
+Description
+-----------
+
+This filter is used in gifs to extract the first frame of them as an image to be used as cover.
```
scorphus, line 9: I suggest GIFs as a more accurate term (Wikipedia also uses it) and a small rewording:

```suggestion
This filter is used in GIFs to extract their first frame as the image to be used as cover.
```

docs/cover.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Obs: This filter will only function when `USE_GIFSICLE_ENGINE = True` are set to true on `thumbor.conf`
```
scorphus, line 11: How about a Note as in other documents?

```suggestion
..note::
    This filter will only function when `USE_GIFSICLE_ENGINE = True` are set to true on `thumbor.conf`
```

scorphus (changes requested):

docs/cover.rst
maiagripp: Removed :)

docs/cover.rst
maiagripp: The suggestions were applied. Thank you!

docs/cover.rst
maiagripp: Changed to .. note::

RaphaelVRossi (approved):
LGTM! 🚀

docs/cover.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. image:: images/animated.gif
+    :alt: Gif before cover filter
+
+::
+
+    `http://localhost:8888/unsafe/filters:cover()/http://server.my/animated_static.gif`
+
+.. image:: images/animated_static.gif
+    :alt: Gif after cover filter
```
scorphus, line 30: Sorry for missing this one.

👉 Tip: please, use the "Commit suggestion" button It's easier and the extra commit gets squashed once we're all happy with the changes 😊

```suggestion
Cover
=====

Usage: `cover()`

Description
-----------

This filter is used in GIFs to extract their first frame as the image to be used as cover.

.. note::

    This filter will only function when ``USE_GIFSICLE_ENGINE`` are set to ``True`` in ``thumbor.conf``:

.. code:: python

    USE_GIFSICLE_ENGINE = True

Arguments
---------

No arguments.

Example
-------

.. image:: images/animated.gif
    :alt: Gif before cover filter

::

    `http://localhost:8888/unsafe/filters:cover()/http://server.my/animated_static.gif`

.. image:: images/animated_static.gif
    :alt: Gif after cover filter
```

scorphus (changes requested):

scorphus:
Good catch, @maiagripp 👍 

For some reason, the docs stopped building almost three years ago: https://readthedocs.org/projects/thumbor/builds/ 🤷🏻‍♂️

I've updated the settings and now it's back building: https://thumbor.readthedocs.io/en/latest/cover.html 🎉 

Thank you for reporting! 🙌

PR #1507, opened 2022-10-25
keep exif copyright information

devppjr:
### Motivation

In light of European Union Directive 790, we need to ensure that copyright owners are being properly attributed, and are not being accidentally removed from their works by overzealous optimizations.

### Solution
 
This PR keeps the image's copyright information, even if the `strip_icc` and `strip_exif` filters are used.

ps: libraries `py3exiv2` and `exif` have been tested, but with no success. 😞

fixes https://github.com/thumbor/thumbor/issues/1502  🎉


guilhermef:
I would use this opportunity to replace the current library for EXIF, `py3exiv2 ` is unmaintained, and it's almost impossible to install successfully.
https://pypi.org/project/exiv2 seems to be in a good state, and it has wheels for most operational systems.

devppjr:
> I would use this opportunity to replace the current library for EXIF, `py3exiv2 ` is unmaintained, and it's almost impossible to install successfully. https://pypi.org/project/exiv2 seems to be in a good state, and it has wheels for most operational systems.

Hi @guilhermef, I agree with your tip, but I believe this could be implemented in another PR, because `exiv2` doesn't have support to `load` and `dump` exif information from/to image bytes which is needed in this PR. What do you think?

devppjr:
ping @guilhermef @scorphus @heynemann 🙃

devppjr:
ping @guilhermef @scorphus @heynemann 🙃

RaphaelVRossi (approved):
LGTM!

guilhermef (changes requested):

guilhermef:
I would also consider adding an integration test for this feature.

devppjr:
> I would also consider adding an integration test for this feature.

thanks for the suggestions, can you review again?

scorphus (changes requested):
`piexif` should remain an extra requirement. A new dependency has dramatic implications (such as packaging) and should be avoided.

Thumbor by itself has no obligation in preserving image copyright in EXIF data. Of course, it should provide means of achieving that should it be of the user's interest and, in that sense, #1502 and this PR are super valid. But `piexif` should remain optional and a configuration variable should be created to activate such feature and, consequently, log any warnings or errors resulting from a missing optional dependency.

guilhermef:
@scorphus since piexif is a pure python package, I don't see any problem having it as a normal dependency.
We used to have this problem with py3exiv2 because of the libboost dependency.

scorphus:
If we can avoid having a hard dependency while doing so with maintainable code, all the better! And I'm totally up for it. I can happily hop in and help coding this. What you guys think?

devppjr:
ping! @scorphus @guilhermef 🙃

tests/engines/test_pil.py
```diff
@@ -13,8 +13,8 @@
 from io import BytesIO
 from os.path import abspath, dirname, join
 from unittest import TestCase, mock
-import piexif
 
+import piexif
```
guilhermef, line 17: Given what @scorphus mentioned, piexif might not be available in all platforms.
We can make it optional so it doesn't fail.

marcelometal: Quote from [documentation](https://piexif.readthedocs.io/en/latest/about.html#dependency):  "Piexif doesn’t depend on any third library"

There are packages for it:

- [Debian](https://packages.debian.org/sid/python3-piexif)
- [Fedora](https://src.fedoraproject.org/rpms/python-piexif)

scorphus: Thanks for the extra info, @marcelometal.

So, my take is that, with `piexif` depending on no third party lib and being readily available and already packaged in many distros, I think we're good to go with it being a hard dependency. That outweighs the extra logic needed to workaround it being unavailable. We need to remember this when drafting a new release, with the proper information about the new dependency.

Now, speaking about the feature. I believe that not every user would want it — much the opposite, if I'd have to guess. Perhaps, most users would like to strip **all** exif information, regardlessly, leaving no traces behind, as they don't need to keep any at all.

So, I still think the feature of "keeping copyright information when stripping exif data" should stay behind a configuration entry, with default to false/inactive.

devppjr: Thanks guys. @scorphus Makes sense to revert the feature that was added in https://github.com/thumbor/thumbor/pull/1524 in this PR?

scorphus: That can and should definitely be gone granted we do not draft a new release until the issue it fixed is solved otherwise. 😊

heynemann (approved):
After all the changes, looks good to merge.

guilhermef (approved):

thumbor/engines/pil.py
```diff
@@ -128,6 +135,18 @@ def create_image(self, buffer):
 
         return img
 
+    def get_exif_copyright(self):
+        exifs = piexif.load(self.image.info.get("exif"))
+        copyright_exif = {}
+        if exifs is None or "0th" not in exifs or exifs["0th"].items() is None:
```
scorphus, line 141: I may be missing something here, but I tried and failed to think of (and reproduce) cases where dict.items() could be `None`. So I believe this line should be corrected:
```suggestion
        if exifs is None or "0th" not in exifs or exifs["0th"] is None:
```

thumbor/engines/pil.py
```diff
@@ -128,6 +135,18 @@ def create_image(self, buffer):
...
+            return None
+
+        for exif_cod, exif in exifs["0th"].items():
+            if exif_cod in KEEP_EXIF_COPYRIGHT_TAGS:
+                copyright_exif[exif_cod] = exif
```
scorphus, line 146: How about iterating over the potentially shorter collection?
```suggestion
        for copyright_tag in KEEP_EXIF_COPYRIGHT_TAGS:
            if copyright_tag in exifs["0th"]:
                copyright_exif[copyright_tag] = exifs["0th"][copyright_tag]
```
This also saves a bit of time — albeit marginal — doing lookups on a dict (constant time) instead of a list (linear).

thumbor/config.py
```diff
@@ -248,6 +248,13 @@
     "Imaging",
 )
 
+Config.define(
+    "PRESERVE_EXIF_COPYRIGHT_INFO",
+    False,
+    "Preserves exif copyright information in generated images.",
```
scorphus, line 254: Nit:
```suggestion
    "Preserves Exif copyright information in generated images.",
```
According to https://en.wikipedia.org/wiki/Exif:
> [...] officially Exif, according to JEIDA/JEITA/CIPA specifications [...]

scorphus (changes requested):
Nice work, @devppjr. Please consider my suggestions below:

scorphus (approved):
Thanks a bunch, @devppjr! 🙌 🚀 :shipit:

PR #1496, opened 2022-10-18
Add Thumbor official Docker image

guilhermef:
This image will be used to create the official helm chart.

PR #1517, opened 2022-10-31
workflows/release: Do not run `docker` on pull requests

scorphus:


scorphus:
Seems like the directives to build on `pull_request` were leftovers from the attempts of making #1496 work.

guilhermef:
@scorphus I've added the release on pull_request, to test if the changes are not breaking the release process, but we only push on tags: https://github.com/thumbor/thumbor/pull/1517/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R117

scorphus:
Fair enough. I missed the if condition. How about not running the `docker` job on pull requests?

scorphus:
If it already built successfully for a previous release (on a previous tag push) it would continue building successfully for that same release, over and over, on every pull request.

guilhermef:
It depends. If you change a dependency or an extension it might break the docker build.

Issue #1504, opened 2022-10-24
Thumbor should respect accept-header


maiagripp:
### Thumbor request URL
```
curl -i http://localhost:8888/unsafe/https://raw.githubusercontent.com/thumbor/thumbor/6837e8e69d98fe6702a26ee0a49704231ab83faf/tests/fixtures/images/image.webp -H "accept: image/jpeg"
HTTP/1.1 200 OK
Server: Thumbor/7.2.0
Content-Type: image/webp
```

### Expected behaviour
If a `WebP`/`AVIF` image is explicitly requested, but it's not proper for rendering based on the `Accept header`, 
then the response should be “downgraded” to a supported image format.

### Actual behaviour
If `WebP` is explicitly requested or the original source, Thumbor doesn't do the "downgrade".

PR #1518, opened 2022-11-01
Define image format by accept header

devppjr:
## Motivation

Thumbor should consider the accept header to define image format, and sometimes the response should be “downgraded” to a supported image format.

## Solution

This PR adds three settings to define the image format based on the accept header.

- `AUTO_JPG`: Specifies whether JPG format should be used automatically if the request accepts it (via Accept header)

- `AUTO_AVIF`: Specifies whether Avif format should be used automatically if the request accepts it (via Accept header) and pillow-avif-plugin is enabled

- `AUTO_HEIF`: Specifies whether Heif format should be used automatically if the request accepts it (via Accept header) and pillow-heif is enabled.

closes https://github.com/thumbor/thumbor/issues/1504

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -343,8 +343,15 @@ well as static gifs with the smallest possible size.**
 WARNING: When using gifsicle engine, filters will be skipped, except for `cover()` filter. thumbor
 will not do smart cropping as well.
 
+AUTO_*
+~~~~~~~~~~~~
+
+These configurations indicates that thumbor will try to automatically convert
+the image format to a lighter image format, following a better compression order
+`WEBP -> AVIF -> JPG -> HEIF`.
+
 AUTO\_WEBP
-~~~~~~~~~~
+^^^^^^^^^^
 
 This option indicates whether thumbor should send WebP images
 automatically if the request comes with an "Accept" header that
@@ -354,8 +361,19 @@ specifies that the browser supports "image/webp".
 
    AUTO_WEBP = True
 
+AUTO\_AVIF
+^^^^^^^^^^
+
+This option indicates whether thumbor should send Avif images
+automatically if the request comes with an "Accept" header that
+specifies that the browser supports "image/avif" and pillow-avif-plugin is enabled.
+
+.. code:: python
+
+   AUTO_AVIF = True
+
 AUTO\_PNG\_TO\_JPG
-~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^
 
 This option indicates whether thumbor should transform PNG images
 automatically to JPEG. If the image is a PNG without transparency and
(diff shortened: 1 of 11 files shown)


devppjr:
ping @guilhermef @scorphus @heynemann 🙃

RaphaelVRossi (approved):
LGTM!

guilhermef:
@devppjr When the AUTO_WEBP was implemented, it was a way to save bandwidth by automatically converting it to WEBP.
Do we get the same benefit by having an `AUTO_*` flag with these formats?

devppjr:
> @devppjr When the AUTO_WEBP was implemented, it was a way to save bandwidth by automatically converting it to WEBP. Do we get the same benefit by having an `AUTO_*` flag with these formats?

Yes, with this changes we get the same benefit as AUTO_WEBP. To use this feature we need these configs are enabled and have the accept header. Below you can see the tests we did using the magick tool:

The original image is `example.jpg`, and we use the follow commands to convert image:

```
magick example.{jpg,webp}
magick example.{jpg,avif}
magick example.{jpg,heif}
magick example.{jpg,png}
```

Results:
```
51K 11 Nov 15:59 example.webp
22K 11 Nov 16:00 example.avif
213K 12 Mai  2022 example.jpg
86K 11 Nov 16:01 example.heif
1,5M 11 Nov 16:01 example.png
```

devppjr:
@guilhermef @scorphus @heynemann ping? 🙃

guilhermef:
Sorry for the delay @devppjr.
Maybe we should clarify that the flags are exclusive, or even check if you have more than one enabled at once and alert it.

devppjr:
@guilhermef By default all these configuration are disabled. This PR aims to allow all of these configurations to be used simultaneously. This update will only be useful if the format filter (`...filters:format()/...`) is not included in the url, the request needs to contain the accept header for the format the client wants, and the plugins are installed. We also log this automatic conversion, like: `"Image format set by AUTO_AVIF as.`

devppjr:
ping! @scorphus @guilhermef 🙃

guilhermef:
@RaphaelVRossi I support the idea of automatically converting to a more suited format depending no the header.
My main concern is not being clear to the users what would happen if the request `Accept` header matches multiple `AUTO_*` flags.

I would be wise to clarify in the Config what the behavior would be if multiple `AUTO_*` flags were enabled simultaneously.

devppjr:
> @RaphaelVRossi I support the idea of automatically converting to a more suited format depending no the header. My main concern is not being clear to the users what would happen if the request `Accept` header matches multiple `AUTO_*` flags.
> 
> I would be wise to clarify in the Config what the behavior would be if multiple `AUTO_*` flags were enabled simultaneously.

@guilhermef thanks for the ideia. We move all this `AUTO_*` configuration to a subtopic in [configuraiton.rst](https://github.com/thumbor/thumbor/pull/1518/files#diff-68a31565c7f34fbe6ac80bbc0b4287ee3711fa615ce5b973c1f83f285d0d28faR346), were indicates what is the sequence that thumbor will to try to convert the image format. What do you think?

devppjr:
!ping @guilhermef 🙃

guilhermef:
@devppjr it looks good to me, but let's check with @heynemann

devppjr:
> @heynemann could you review this PR? 🙃

@heynemann 🏓

devppjr:
> > @heynemann could you review this PR? 🙃
> 
> @heynemann 🏓

With no response from @heynemann , could we merge it to master @guilhermef?

mpdude:
I think this misses setting the `Vary: Accept` response header, see #1563.

heynemann:
@guilhermef do you think you can take a look at this? Otherwise I'll look into it as soon as I can.

PR #1521, opened 2022-11-08
Move Thumbor tests to run on host

guilhermef:
This will remove the need for separate Dockerfiles per Python version and simplify the test workflow.


heynemann:
This is how it was previously. And every thumbor build took several minutes. After we started running tests in docker the builds started taking seconds. I'm all for improving the build, but time is key here. We can't have slow builds.

guilhermef:
This change wouldn't impact the build times. We're already installing the build time dependencies on Docker during the action execution, but it will make it easier to cache.

heynemann:
So we're once again taking long for building :/ How can we make builds fail in a couple seconds? Then if they succeed we update the docker images? That's why I added running tests in docker. Making it easier to contribute.

guilhermef:
The whole test process takes around 2 minutes and, the environment setup takes less than 30 seconds.
It's a good trade-off since we would be able to validate that library updates aren't preventing Thumbor from being built.

PR #1519, opened 2022-11-02
feat(thumbor): Replace pyexiv2 with piexif

guilhermef:
Installing pyexiv2 has been a constant issue on Docker, and it's not compatible with Python 3.11.
This library that @devppjr used on #1507, makes it easier to maintain.

PR #1524, opened 2022-11-10
fix(engines): Try to import piexif and use it when available

scorphus:
As packages specified in `extras_require` are not installed by default, `piexif` may be unavailable. Thus the try/except around the import line and the ensuing check of its availability before use.

Ref.: 937628d29ae6e38c0a7aebde7cff74062eb41136
Ref.: #1519

### DIFF
--- a/tests/engines/test_base_engine.py
+++ b/tests/engines/test_base_engine.py
@@ -315,3 +315,18 @@ class BaseEngineTestCase(TestCase):
         expect(self.image).to_equal(((1, 2), (3, 4)))
 
         expect(self.engine.get_orientation()).to_equal(1)
+
+    @mock.patch("thumbor.engines.piexif")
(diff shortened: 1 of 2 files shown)


RaphaelVRossi (approved):
Nice catch! Thanks!

scorphus:
Let me suggest an additional test here, in a bit.

devppjr:
> Let me suggest an additional test here, in a bit.

in this PR https://github.com/thumbor/thumbor/pull/1507 `piexif `will be a required lib. maybe this implementation isn't necessary, what do you think?

scorphus:
I think piexif shouldn't be a requirement. A new requirement has many implications and should be avoided.

guilhermef:
@scorphus you're right, but we should consider moving piexif to the normal requirements since it doesn't require libboost as py3xiv2 used to.

guilhermef (approved):

PR #1530, opened 2022-12-06
Revert "fix(engines): Try to import piexif and use it when available …

maiagripp:
As discussed in https://github.com/thumbor/thumbor/pull/1507 we reverted the #1524 commit, since the import of pyexif lib verification is no longer required.


maiagripp:
@scorphus Hey! As discussed with @devppjr in https://github.com/thumbor/thumbor/pull/1507 we reverted your #1504 commit. 
As soon as you review it, we are going to draft a new release.

scorphus (approved):
I was going to do that, felt like my responsibility 😊 Thanks for doing the work!

:shipit:

Issue #1540, opened 2023-01-02
Thumbor shouldn't convert png to jpeg when image have transparency.


devppjr:
### Thumbor request URL
http://domain.org/unsafe/100x100/http://mydomain.org/image.jpeg

### Expected behaviour
Thumbor shouldn't convert png to jpeg when image have transparency.

### Actual behaviour
Thumbor is converting png to jpeg when image have transparency.

PR #1541, opened 2023-01-02
fix: auto convert image with transparency

devppjr:
# Motivation

Thumbor shouldn't convert images with transparency to jpeg format.

# Solution 

Was added a verification if image has transparency.

ps: Was tested with others formarts (`image/avif`, `image/heif` and `image/webp`) and this problem doesn't happened.

# Issue

closes https://github.com/thumbor/thumbor/issues/1540


RaphaelVRossi (approved):
Thanks! @devppjr 🎉

PR #1542, opened 2023-01-05
Fixed piexif load when exif tag is broken

RaphaelVRossi:
When the exif tag is broken, and the `PRESERVE_EXIF_COPYRIGHT_INFO` variable is True, all exif data will be kept.


fbtravi:
LGTM

devppjr (approved):
Nice catch! Thanks @RaphaelVRossi 🚀

Issue #1543, opened 2023-01-10
Thumbor should raise exception


devppjr:
### Expected behaviour
Thumbor should raise exception when loading results [handlers/__init__.py#L637](https://github.com/thumbor/thumbor/blob/master/thumbor/handlers/__init__.py#L637) 

### Actual behaviour
Thumbor is handling error when loading results and because of that sentry can't capture event.

### Operating system
OSx, Ubuntu, etc

### Your thumbor.conf
Your config file (WARNING: replace your SECURITY_KEY key with a dummy key or your key will become public)

devppjr:
We are thinking in two solutions:

1 - just raise exception

```python
        try:
            result = await self.context.thread_pool.queue(
                operation=functools.partial(self._load_results, context),
            )
        except Exception as error:  # pylint: disable=broad-except
            logger.exception("[BaseHander.finish_request] %s", error)
            self._error(500, f"Error while trying to fetch the image: {error}")

            raise
```

2 - use behind a configuration to handle the error, like  [thumbor/transformer.py#L172](https://github.com/thumbor/thumbor/blob/master/thumbor/transformer.py#L172)

```python
        try:
            result = await self.context.thread_pool.queue(
                operation=functools.partial(self._load_results, context),
            )
        except Exception as error:  # pylint: disable=broad-except
            logger.exception("[BaseHander.finish_request] %s", error)
            self._error(500, f"Error while trying to fetch the image: {error}")

            if self.context.config.USE_CUSTOM_ERROR_HANDLING:
                self.context.modules.importer.error_handler.handle_error(
                    context=self.context,
                    handler=self.context.request_handler,
                    exception=sys.exc_info(),
                )
            return
```

what do you think @RaphaelVRossi @scorphus @guilhermef @maiagripp @heynemann @kkopachev ?

PR #1544, opened 2023-01-11
Thumbor should send error to error handler

devppjr:
## Motivation

The loading results func is one of the most importants funcs in Thumbor, and we should handle errors that happens inside this func. Thumbor should send errors to error handler when loading results. 

## Problem

When Thumbor handle error with try/except, the error handler (sentry) can't capture event

## Solution

This PR sends error to custom error handler when is enabled

ps: was tested the first solution in issue, but will log the same exception 3 times, so because of that we choose the solution 2

closes https://github.com/thumbor/thumbor/issues/1543


RaphaelVRossi:
Nice catch @devppjr 🎉

RaphaelVRossi (approved):

PR #1546, opened 2023-01-23
Handle exception when loading exif data

devppjr:
## Motivation

Thumbor should return image when can't load image exif data

## Problem

Thumbor raises error when can't load image exif data. (`PRESERVE_EXIF_COPYRIGHT_INFO` config is enabled)

## Solution

Thumbor will return image even can't load image exif data.


RaphaelVRossi (approved):
🎉

PR #1568, opened 2023-05-31
fix check_extensibility_modules : RESULT_STORAGE default value is None

gdegoulet:


thumbor/doctor.py
```diff
@@ -161,18 +161,19 @@ def check_extensibility_modules(cfg):
...
-                    format_error(
-                        module,
-                        str(error),
-                        error_message,
+            if module is not None:
+                try:
+                    import_module(module)
+                    print_success(f"{CHECK} {module}")
+                except ImportError as error:
+                    print_error(f"{CROSS} {module} - {error_message}")
```
scorphus, line 169: Would you consider decreasing the level of indentation of the code and continuing the loop earlier?

```suggestion
            if not module:
                continue
            try:
                import_module(module)
                print_success(f"{CHECK} {module}")
            except ImportError as error:
                print_error(f"{CROSS} {module} - {error_message}")
                # ...
```

That would make the code read more easily while behaving the same.

scorphus (changes requested):
Hey, thanks for your contribution intent!

1. Do you know under which circumstances `module` is `None` on line 164?
2. Would you be willing to write a test for that?

In the meantime, please consider my suggestion below.

devppjr:
ping @gdegoulet

PR #1578, opened 2023-07-01
fix: Ensure result storage is passed results in bytes

### LINKED ISSUE #1577: Why is a file generated every time a meta url is refreshed?
Why is a file generated every time a meta url is refreshed?

url：
http://127.0.0.1:8888/unsafe/meta/http://127.0.0.1:9999/test/5d8cce994c0190cbdb2f1e0914286c0a

These files are all from the same meta url.

Shouldn't only one file be correct?

example

scorphus:
When results are text, such as JSON content for `/meta/` endpoints, ensure the results are converted to bytes before passing them to a result storage.

Fix #1577


devppjr (approved):
🚀

scorphus:
Thanks for the review, @devppjr!

PR #1580, opened 2023-07-12
Add security layer upload images

MacKey-255:
- Added a simple way to remove expired images
- Added a security layer to upload images


digwa-ing:
Nice pull request. I have been waiting for upload security for a while. Thank you. I hopw this will make it into the main branch.

PR #1610, opened 2023-10-28
Enhanced image URL validation to prevent malicious requests

ninj4c0d3r:
This pull request introduces crucial updates to the URL validation process, ensuring that only genuine image URLs are processed and forwarded.

## Background:
Previously, the application was vulnerable to forwarding any URL without adequate validation, potentially enabling an attacker to interact with internal applications. An example of such a malicious request could be: `http://thumbor/unsafe/http://0.0.0.0:6000/app.php%3Fexec=system%28%22touch%20%2Ftmp%2Frce%22%29%3B` where the system would not validate if the requested resource was an actual image, thus sending a potentially harmful request to the target URL.

## Changes:
- Enhanced the URL validation logic to strictly validate and forward only image URLs.
- Added checks to ensure that the URL path ends with valid image extensions.
- Conducted thorough testing to ensure the resilience of the updated validation against bypass attempts and other malicious scenarios.

It is highly recommended to merge these changes to fortify the application's security posture and mitigate potential exploitation risks.

### DIFF
--- a/thumbor/loaders/http_loader.py
+++ b/thumbor/loaders/http_loader.py
@@ -11,7 +11,7 @@
 import datetime
 import re
 import socket
-from typing import Pattern
+from typing import Callable, List, Union, Pattern
 from urllib.parse import quote, unquote, urlparse
 
 import tornado.httpclient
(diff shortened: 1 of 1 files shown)


devppjr:
Hi @ninj4c0d3r , thanks for your contribution ❤️ . 

It is possible that some image upstreams do not include the image extension in their URL, and with this change these images will no longer be processed. Furthermore, we believe that if we accept your contribution we will be restricting the types of images on which Thumbor can perform its functions.

Issue #1563, opened 2023-05-02
`Vary: Accept` header for `AUTO_AVIF`?


mpdude:
When the image format is selected as WebP based on the `AUTO_WEBP` setting and the value of the `Accept` header, the `Vary` response header is set accordingly. This is necessary for downstream caches, to make them aware that for other `Accept` request header values, a previously cached response must but be reused.

https://github.com/thumbor/thumbor/blob/9b54c904015789705577169ba61342b008b24d64/thumbor/handlers/__init__.py#L713-L727

When a similar feature was implemented in #1518 for AVIF and HEIF, this condition was not amended for the new formats. I think that's a bug that may cause problems with downstream caches.

mpdude:
I still think that’s a valid issue.

mpdude:
I still think that’s a valid issue.

PR #1626, opened 2023-11-30
Should have vary header with auto heif and avif

devppjr:
This PR aims to add `Vary` header in Thumbor response to avoid problems with downstream caches.

Closes https://github.com/thumbor/thumbor/issues/1563
thanks @mpdude


RaphaelVRossi (approved):
Thanks! 🎉

mpdude:
Thank you!

PR #1643, opened 2024-01-23
feat(engines/pil.read_multiple): Use Pillow to save multipage images as GIF

scorphus:
The functionality provided by the images2gif module is [part of Pillow since 2015][1]. It also renders `gifsicle` unnecessary in this context.

[1]: https://pillow.readthedocs.io/en/stable/releasenotes/3.0.0.html


scorphus:
With this change, we're free to remove the version constraints for Pillow:

https://github.com/thumbor/thumbor/blob/812ed4c6c532154d79ce2094523ffd716f53ba45/setup.py#L153-L154

scorphus:
Please, check #1644 before commenting on lint failure.

thumbor/engines/pil.py
```diff
@@ -452,46 +451,22 @@ def read(  # noqa
...
-            converted_images.append(image.convert("RGB"))
-            coordinates.append((0, 0))
-            dispose.append(1)
-
-        loop = int(self.image.info.get("loop", 1))
-
-        images = gif_writer.convertImagesToPIL(converted_images, False, None)
-        gif_writer.writeGifToFile(
-            img_buffer, images, duration, loop, coordinates, dispose
+        _, tmp_file_path = mkstemp()
```
guilhermef, line 454: @scorphus, you can also create the temp file using a context manager: https://docs.python.org/3/library/tempfile.html#examples

scorphus: Seems like I overlooked that one. Thanks, @guilhermef, good catch!

guilhermef (approved):

thumbor/engines/pil.py
```diff
@@ -452,50 +450,23 @@ def read(  # noqa
...
-
-        command = ["gifsicle", "--colors", "256", tmp_file_path]
-
-        popen = Popen(  # pylint: disable=consider-using-with
-            command, stdout=PIPE
-        )
-        pipe = popen.stdout
-        pipe_output = pipe.read()
-        pipe.close()
+        with NamedTemporaryFile() as tmp_file:
```
guilhermef, line 453: We might not need to write a file for gifscicle:
```
gifsicle reads and processes GIF input files in order. If no GIF input file is given, or you give the special filename ‘-’, it reads from the standard input.
```

thumbor/engines/pil.py
```diff
@@ -452,50 +450,23 @@ def read(  # noqa
...
+                format="GIF",
+                save_all=True,
+                append_images=images[1:],
+                duration=[im.info.get("duration", 80) / 1000 for im in images],
+                loop=int(self.image.info.get("loop", 1)),
+            )
 
-        if popen.wait() == 0:
-            results = pipe_output
+            command = ["gifsicle", "--colors", "256", tmp_file.name]
```
guilhermef, line 463: It seems that Pillow improved Gif support, we might not need gifsicle anymore https://github.com/search?q=repo%3Apython-pillow%2FPillow+gif+language%3Arst+path%3Adocs&type=code

scorphus: Yes! I'm coming to the exact same conclusion. After some experimentation, I'm seeing hardly any improvements to GIFs with `gifsicle`, to a point where it doesn't make much sense to keep supporting it, especially by running it as a sub process.

scorphus:
There we go, @guilhermef. I believe we're much better of having no `gifsicle` in `pil.read_multiple`. I'd appreciate another round of review.

guilhermef (approved):

Issue #1692, opened 2024-09-20
ModuleNotFoundError: No module named 'setuptools' while running github actions


RaphaelVRossi:
https://github.com/thumbor/thumbor/actions/runs/10958697099/job/30429460166

```
Run docker exec thumbor_test make compile_ext
Traceback (most recent call last):
  File "/app/setup.py", line 1[5](https://github.com/thumbor/thumbor/actions/runs/10958697099/job/30429460166#step:8:6), in <module>
    from setuptools import Extension, setup
ModuleNotFoundError: No module named 'setuptools'
make: *** [Makefile:1[7](https://github.com/thumbor/thumbor/actions/runs/10958697099/job/30429460166#step:8:8): compile_ext] Error 1
```

PR #1696, opened 2024-10-10
fix: ModuleNotFoundError: No module named 'setuptools'

Ronkiro:
Probably this fixes https://github.com/thumbor/thumbor/issues/1692 needs testing though

### DIFF
--- a/TestDockerfile
+++ b/TestDockerfile
@@ -22,4 +22,4 @@ RUN apt-get update -y && \
     apt-get install -y --reinstall --no-install-recommends --no-install-suggests build-essential && \
     apt-get clean
 
-RUN pip install --upgrade pip
+RUN pip install --upgrade pip setuptools


marcelometal:
Root cause: https://github.com/docker-library/python/issues/952

RaphaelVRossi:
Hey @Ronkiro thanks for contributing! :smile: 

I was testing it too, so maybe it's better if we install setuptools using [setup.py](https://github.com/thumbor/thumbor/blob/master/setup.py#L148). Either way, we need to install it to run it locally with Python 3.12. :thinking:

RaphaelVRossi:
@Ronkiro Inside the `install_requires`. Something like: `"setuptools>=70.0.0",`

https://github.com/thumbor/thumbor/blob/5b19638385bc11b3aedfa6130a9a29438ab89b48/setup.py#L148

Ronkiro:
@RaphaelVRossi  yeah, i mean, if we only do it we get an error because the code is tied to setuptools itself.

i.e. (the code below has setuptools inside `install_requires`)

```sh
python setup.py install
  File "C:\Programas\thinkai\trendy\gpt_job\setup.py", line 15, in <module>
    from setuptools import Extension, setup
ModuleNotFoundError: No module named 'setuptools'
```

RaphaelVRossi:
:thinking: but Thumbor's Makefile is using pip itself.

`python3 -m pip install -e .[tests]`

see: https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead

Ronkiro:
I didn't know about that, ty! Closing for now

PR #1701, opened 2024-11-04
Filter Watermark - allow h_ratio and w_ratio as a real number

asokani:
In order to set the exact size of the watermark, I need to have the percentage value as a real number. There is no reason not to. This is backward compatible


asokani:
Hey @heynemann  I think this patch is useful.

RaphaelVRossi:
Hey @asokani, thank you for your interest in contributing to Thumbor!

Could you please explain this regex? It seems a bit like magic to me.

Additionally, we should update the documentation for the watermark filter with this new use case. The source code for it is here: https://github.com/thumbor/thumbor/blob/master/docs/watermark.rst

asokani:
Hello @RaphaelVRossi. Thanks for asking. The regex comes straight from https://github.com/thumbor/thumbor/blob/master/thumbor/filters/__init__.py#L132
We can't use it in this context "as is" though, because we also need to support string "none". In short, there are four options: DDD.optional_DDD | optional_DDD.DDD | DDD | none (D means digit).

I've also updated the docs. I explicitly pointed out that percentages can be real numbers. In fact - percentages are ALWAYS real numbers. I don't know why there was an artificial restriction to only whole integers. That never made sense.

asokani (commented):
thanks ai

asokani:
Hey @heynemann, I opened this request on Nov 4, 2024 just to share my patch with others. But no one seems to care. This is so frustrating

heynemann:
Sorry for that @asokani . We should have picked up this earlier. Approving now once the workflows pass.

heynemann:
Actually I think in order to approve this PR you need to include at least a few tests here https://github.com/thumbor/thumbor/blob/master/tests/filters/test_watermark.py to ensure we are properly processing watermarks for floating point numbers, right?

asokani:
@heynemann I've added the requested tests

asokani:
@heynemann not stale

tests/filters/test_watermark.py
```diff
@@ -223,6 +223,71 @@ async def test_watermark_filter_simple_60p_80p(self):
...
+
+    @gen_test
+    async def test_watermark_filter_fractional_float_w_ratio(self):
+        image = await self.get_filtered(
+            "source.jpg",
+            "thumbor.filters.watermark",
+            "watermark(watermark.png,30,-50,20,33.33)",
+        )
+        assert image is not None
+        assert len(image) > 0
```
marcelometal, line 279: `get_filtered()` returns a `PIL.Image.Image`, so `len(image)` raises `TypeError: object of type 'Image' has no len()`. The same call at line 289 also fails, leaving both fractional-ratio tests red. Please replace these checks with assertions on the rendered watermark's size/content. A 600px-wide source with `w_ratio=9.5` should visibly contain a 57px-wide watermark; checking only the output image's existence or overall size would also pass when the watermark filter is skipped.

tests/fixtures/watermark_fixtures.py
```diff
@@ -75,6 +75,16 @@
     (200, 300),
     (25, 50),
     (50, 25),
+    # Float ratios
+    (33.33, None),
```
marcelometal, line 79: Adding fractional cases exposes integer-only assumptions in `test_watermark_filter_calculated_resizing`. The first failure is `33.33%` of `800px`: the computed limit is `266.64`, the renderer rounds it to `267px`, and `assert_fits_into` rejects `267 <= 266.64`. Please update the test expectations to account for pixel rounding, including the exact equality used by the subsequent `fill out` check, while still checking the expected watermark dimensions.

docs/watermark.rst
```diff
@@ -35,10 +35,12 @@ Arguments
    (fully opaque) and 100 (fully transparent).
 -  ``w_ratio`` - percentage of the width of the image the watermark should fit-in, defaults to 'none'
    (without the single quotes) which means it won't be limited in the width on resizing but also won't
-   be resized based on this value
+   be resized based on this value. Sharpen amount. Can be integer or float number between :math:`0.0` and
```
marcelometal, line 38: Please remove the copied `Sharpen amount` text and correct the stated `0.0–100.0` range for both ratios. Values above 100 are already supported for upscaling and covered by the existing fixtures, and this patch adds no upper-bound validation. The new example should also explain rounding to whole pixels rather than promise `perfect accuracy`.

marcelometal (changes requested):
Rebased this PR onto the current master (`2a486080`) while preserving the four original author commits. I also added a separate compatibility commit converting the remaining `preggy` assertions to plain `assert`, matching master's test infrastructure.

The shared decimal regex enables the requested fractional ratios, including the 9.5% / 600px = 57px case from #1702. However, the rebased PR still has three reproducible test failures that need to be addressed before merge.

Validation on Python 3.14:

- Watermark tests: 20 passed, 3 failed.
- `make unit`: the same two `TypeError`s and one rounding assertion failure; 9 skips and 2 expected failures.
- Pre-commit checks, including Black, Flake8, isort and Pylint: passed.

Please address the inline comments. In particular, the fractional-ratio tests should verify the actual rendered watermark dimensions, not just that an image object exists, so they catch the original behavior where the filter was silently skipped.

PR #1703, opened 2024-11-27
Fix setting minimal height for no_upscale filter

crisp-tweakers:
It now uses the intrinsic image width iso height; this is a bug caused by https://github.com/thumbor/thumbor/commit/44f86442ba341b9b7cccc392976e9c8191082281


ericcitaire:
up !

PR #1708, opened 2024-12-17
Change document default port to 80

ghorbani-ali:
Based on [this doc](https://thumbor.readthedocs.io/en/latest/hosting.html#running-with-docker), the port default value must be changed to 80 from 8888.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -22,7 +22,7 @@ to *0.0.0.0* (listening on localhost and current IP).
 ~~~~~~~~~~~~
 
 The port that Tornado will listen for incoming request. It defaults to
-*8888*.
+*80*.
 
 -c or --conf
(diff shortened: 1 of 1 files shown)


OndrejIT:
@ghorbani-ali The default port in the thumbor is 8888, [/thumbor/console.py#L32](https://github.com/thumbor/thumbor/blob/master/thumbor/console.py#L32).
You are referring to a docker image that has nginx in front of the thumbor that listens on 80 [/nginx-proxy-cache/nginx.tmpl#L236](https://github.com/MinimalCompact/thumbor/blob/master/nginx-proxy-cache/nginx.tmpl#L236)

ghorbani-ali:
That is absolutely right; I didn't realize this.
Thanks

PR #1712, opened 2025-01-31
Update file_storage.py

jan-panoch:
we need to find-out in which file is located result for a url, so we added debug logging of url-filename pair.


aoqfonseca (approved):

RaphaelVRossi:
hey @jan-panoch first things first, thank you for contributing to Thumbor! I approved the GitHub action and saw the lint was failing, so could you run black to reformat all lines?

PR #1724, opened 2025-05-02
chore: updated ubuntu release version

RaphaelVRossi:
Since 2025-04-15 github actions drops the support for ubuntu 20.04, as you can see in the following error message 

`This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101`


.github/workflows/release.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-22.04
```
marcelometal, line 48: Other jobs are already using `ubuntu-latest`, so I believe we can safely use it here as well for consistency. 
What do you think?

PR #1775, opened 2025-11-14
Add support to Python 3.14

marcelometal:


scorphus (approved):
LGTM 👍

guilhermef:
@marcelometal we can update Pillow and Pillow-heif, the newer version already has wheels for Python 3.14

marcelometal:
> we can update Pillow and Pillow-heif, the newer version already has wheels for Python 3.14

remoteCV is the main problem. I’ll release a new version of it and then update this PR.

PR #1781, opened 2025-12-22
Use rand crate instead of unsafe libc random() in noise filter

scorphus:
Replaces unsafe libc FFI with the `rand` crate in the `_noise` filter. The crate was already a dependency but wasn't being used.

The test fixture `noise.png` was regenerated to match the new random output.

Note: seed parameter type changed from `u32` to `u64`. This is unlikely to break anything since seeds are typically small integers.


scorphus:
Copilot's review is nonsense.

marcelometal:
Incorporated the _noise rand-based RNG change into the squashed convert_filters_to_rust patch, including the updated noise.png fixture. Closing this PR as superseded. Thanks!

PR #1806, opened 2026-03-19
Add metrics for filters

marcelometal:
Introduce success, error, and timing metrics for filters at the filter execution level, including multi-frame images. Use time.perf_counter() for timing, add coverage for multi-frame engines, and document the metric behavior.

When a metrics backend is enabled, each filter execution emits a ``filter.<filter_name>.time`` timing. Successful executions also emit ``filter.<filter_name>.count``, while failed executions emit ``filter.<filter_name>.error``. The ``<filter_name>`` portion comes from the decorated method name, so it should stay stable if you rely on dashboards or alerts.


scorphus (approved):
Nice addition for observability 🙌 Thanks!

Clean implementation ✨ Good test coverage 👍 Docs updated 📖

Once #1801 gets merged, this will need a rebase and the tests should use sane, native pytest assertions.

PR #1807, opened 2026-03-24
fix package version metadata

marcelometal:
fix regression 43ca793


guilhermef (approved):

scorphus (approved):
Fixes version duplication. Single source of truth restored. 👍

PR #1819, opened 2026-04-29
Respect ALLOWED_SOURCES in the frame filter

4390c336:
Add the same loader validation used by the watermark filter before loading frame assets, so disallowed external frame URLs are rejected with HTTP 400.

### DIFF
--- a/tests/filters/test_frame.py
+++ b/tests/filters/test_frame.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+# thumbor imaging service
+# https://github.com/thumbor/thumbor/wiki
+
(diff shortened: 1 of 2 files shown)


marcelometal:
@4390c336 thanks for the PR!

The change makes sense: `frame()` should respect `ALLOWED_SOURCES`, just like the main image request and `watermark()` do.

Could you add a small test for allowed/disallowed frame sources before we merge?

One note: validation happens only on cache miss, matching `watermark()` behavior.

4390c336:
@marcelometal done. I followed the same code in test_watermark.py

marcelometal:
Thank you @4390c336

PR #1858, opened 2026-08-02
feat: parse the Accept header instead of searching substrings

scorphus:
👉 Note: Third of five PRs splitting up #1746.

`RequestParameters` now parses the `Accept` header into media ranges with quality values, handling quoted parameters, whitespace and case. It exposes `accepts_avif`, `accepts_heif`, `accepts_png` and `accepts_jpeg` next to the existing `accepts_webp`, and `accepts_mime_type` answers from the parsed values.

Three behavior changes, all on purpose:

- A format listed with `q=0` is rejected. Before, `image/webp;q=0` still counted as accepting WebP because only the substring was checked.
- `image/*` accepts every image format. Before it matched nothing.
- `*/*` keeps its old meaning: it only enables JPEG, never the other formats, so existing `AUTO_WEBP` deployments keep the same behavior.

New keyword arguments were appended after the existing ones, so callers that pass positional arguments keep working. Handlers that build their own request objects keep the old substring lookup.

Big thanks, @marcelometal!

### DIFF
--- a/thumbor/context.py
+++ b/thumbor/context.py
@@ -14,6 +14,95 @@ from thumbor.metrics.logger_metrics import Metrics
 from thumbor.threadpool import ThreadPool
 
 
+def _split_http_header(value, delimiter):
+    parts = []
+    current_part = []
+    quoted = False
+    escaped = False
+
+    for character in value:
+        if escaped:
+            current_part.append(character)
+            escaped = False
+            continue
+
+        if quoted and character == "\\":
+            current_part.append(character)
+            escaped = True
+            continue
+
+        if character == '"':
+            quoted = not quoted
+        elif character == delimiter and not quoted:
+            parts.append("".join(current_part))
+            current_part = []
+            continue
+
+        current_part.append(character)
+
+    parts.append("".join(current_part))
+    return parts
+
+
+def _parse_accept_quality(value):
+    try:
+        quality = float(value.strip())
+    except ValueError:
+        return 0.0
+
+    return quality if 0 <= quality <= 1 else 0.0
+
+
+def _parse_accept_media_range(media_range):
+    parts = [part.strip() for part in _split_http_header(media_range, ";")]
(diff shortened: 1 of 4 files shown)


marcelometal (approved):

thumbor/context.py
```diff
@@ -14,6 +14,95 @@
 from thumbor.threadpool import ThreadPool
 
 
+def _split_http_header(value, delimiter):
```
guilhermef, line 17: @scorphus werkzeug has some functions to help with header parsing

https://werkzeug.palletsprojects.com/en/stable/http/

scorphus: Thanks, @guilhermef, TIL!

I looked into it — werkzeug isn't among our dependencies, not even transitively, so it would mean adding werkzeug + MarkupSafe just for this one header. It would also tie thumbor to whatever werkzeug version distros ship (Debian stable can lag years, and werkzeug's API moves).

Since the parser is ~90 lines and the tests cover the cases we care about (`q=0`, `image/*`, quoted parameters), I'd rather keep it in-house.

What is your take? Either way, I'm happy to revisit if we ever pull werkzeug in for something bigger.

/cc @marcelometal for your thoughs and your angle on Debian packaging

guilhermef: Sorry @scorphus, I checked it on my phone.
I thought Tornado had Werkzeug as a dependency; you're right, it's not worth it for just the header parser.

scorphus: No worries, I had to double-check it myself. So many frameworks 😄

marcelometal: Reiterating what I mentioned to Pablo in private, IMHO, avoiding a new dependency makes sense here, especially if it would be introduced only for this header parsing.

We could create an issue to revisit this in the near future if werkzeug becomes relevant for other parts of the project or if maintaining our own parser becomes a concern.

guilhermef (approved):

PR #1859, opened 2026-08-05
feat: separate result storage entries by accepted image formats

scorphus:
👉 Note: Fourth of five PRs splitting up #1746.

Result storage keyed every response by URL only, but with `AUTO_AVIF`, `AUTO_JPG`, `AUTO_HEIF` or `AUTO_PNG` enabled the same URL produces different formats for different clients — so a cached AVIF could be served to a client that never accepts AVIF. Two commits:

1. **Format decisions read the parsed `Accept` values.** `can_auto_convert_to_avif/heif/jpg/png` now use the `accepts_*` attributes from #1858, so the decision and the cache key always see the same reading of the header. Handlers that override `accepts_mime_type` keep working, and an explicit `q=0` rejection of JPEG wins over such an override.

2. **Result storage entries are separated by accepted formats.** The new `thumbor.auto_image_format` module computes a cache-key discriminator; the file result storage uses it as its namespace. `AUTO_WEBP`-only setups keep the legacy `auto_webp`/`default` namespaces and their migration. Setups with the other `AUTO_*` flags or `AUTO_PNG_TO_JPG` get an isolated namespace, because their old `default` namespace can hold a different format for the same URL. When a handler overrides `accepts_mime_type` and one of those formats is active, result storage is skipped for that request — no safe key can be computed before the engine exists.

⚠️ Upgrade note: deployments running `AUTO_AVIF`, `AUTO_JPG`, `AUTO_HEIF`, `AUTO_PNG` or `AUTO_PNG_TO_JPG` start with an empty result-storage namespace after this change — every cached result is regenerated once. Old entries are not removed automatically. Custom result storages should adopt `get_auto_image_format_cache_key` (see the new docs section).

Big thanks, @marcelometal 🤘

### DIFF
--- a/thumbor/handlers/__init__.py
+++ b/thumbor/handlers/__init__.py
@@ -31,6 +31,10 @@ from thumbor.utils import CONTENT_TYPE, EXTENSION, logger
 
 HTTP_DATE_FMT = "%a, %d %b %Y %H:%M:%S GMT"
 
+MIMETYPE_IMAGE_WILDCARD = "image/*"
+MIMETYPE_JPEG = "image/jpeg"
+MIMETYPE_JPG = "image/jpg"
+
 DEFAULT_AUTO_IMAGE_FORMAT_PREFERENCE = ("webp", "avif", "jpg", "heif", "png")
 
 LEGACY_AUTO_IMAGE_FORMAT_METHODS = {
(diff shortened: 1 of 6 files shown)


thumbor/handlers/__init__.py
```diff
@@ -436,9 +441,63 @@ def accepts_mime_type(self, mimetype=""):
...
+        accepts_mime_type_method = self.accepts_mime_type
+        if self._has_custom_accepts_mime_type():
+            if accept_attr == "accepts_jpeg":
+                return self._custom_accepts_jpeg(accepts_mime_type_method)
+
+            return any(
+                accepts_mime_type_method(mimetype) for mimetype in mimetypes
+            )
+
+        return bool(getattr(self.context.request, accept_attr, False))
```
marcelometal: `#1858` deliberately preserved `BaseHandler.accepts_mime_type()`'s header fallback for handlers that build legacy request objects without the new parsed `accepts_*` attributes. This path now bypasses that fallback and treats a missing attribute as `False`.

For example, with `SimpleNamespace(headers={"Accept": "image/avif"}, engine=...)`, `handler.accepts_mime_type("image/avif")` is `True`, but `handler.can_auto_convert_to_avif()` becomes `False`. The same regression applies to HEIF, JPEG, and PNG.

Please preserve the legacy fallback when `accept_attr` is absent, while also keeping the result-storage discriminator safe (for example, bypass result storage for this legacy request shape, or populate the parsed attributes before cache lookup). A regression test using a request object without `accepts_*` attributes would make this explicit.

marcelometal (changes requested):
Thanks for splitting this out. The cache namespace/versioning, conservative legacy migration, and read/write cache bypass for custom `accepts_mime_type` overrides all look solid.

I found one backward-compatibility regression that needs to be addressed before merging; I left the details inline.

Please also restore the SonarCloud quality gate. It currently reports 36 duplicated new lines (4.45%, with a 3% limit) in `tests/handlers/test_base_handler_compatibility.py`, in the similar setup blocks beginning around lines 214 and 253. Sharing that setup or combining the cases into one parametrized test should resolve it.

I re-ran the 25 compatibility tests and exercised an additional cache-key matrix; the behavior otherwise looks correct.

thumbor/handlers/__init__.py
scorphus: ~Yeah... I left that whole refactor as part of the last PR, but I can definitely pull it forward form that branch.~

Actually, that's a nice catch — this bug was hidden in the original #1746 PR. Thanks, @marcelometal 👍

scorphus:
Updated! Rebuilt on top of current master, now three commits:

1. `refactor: walk the automatic formats in a fixed order` — pulled forward from the preference branch, as promised. Bonus: kills the duplicated bodies SonarCloud was crying about.
2. `feat: decide automatic formats from the parsed Accept values` — with the fix for the regression you caught, @marcelometal: no parsed `accepts_*` on the request object? Fall back to reading the `Accept` header, exactly like master. Your regression test is in, covering all four formats. The repeated literals are constants now, too. Thanks for the catch!
3. `feat: separate result storage entries by accepted image formats` — same fix, storage side: legacy-shaped requests skip result storage, since a cache key computed from missing attributes can't be trusted. Unit-tested both ways (bypassed with `AUTO_AVIF` on, kept with `AUTO_WEBP`-only).

PTAL 🙏

marcelometal (approved):

PR #1861, opened 2026-08-10
feat: add AUTO_IMAGE_FORMAT_PREFERENCE

### LINKED ISSUE #1688: Question: Why is webp priorised over avif in AUTO_*?
Hello, I just wanted to know what is the rationale (if there is one) behind serving webp instead of avif if the browser supports both and both are enabled in AUTO_* options?

Relevant code: 
https://github.com/thumbor/thumbor/blob/252c5b10ff4766b8a09e0a017a11cb9912bf557a/thumbor/handlers/__init__.py#L498-L510

From what I can tell this looks like it's mostly because of historical reasons. The order of the if statements with avif-after-webp was added in #1476  : https://github.com/thumbor/thumbor/pull/1476/files#diff-1a881ee8d8f859902e12d41fa442ce25d70161ef9b66ac04df32444900323a6f

But I'd like to know if there's more to it (avif maturity? Implementation maturity? no clear benefit for avif over webp?), since I'm no image format expert.
I don't think any browser currently supports avif but not webp, so I think avif is only served here if AUTO_WEBP is set to false (so that the order is then AVIF, JPG, HEIF, PNG). Is that intended?

Thanks for your time!

scorphus:
👉 Note: Fifth and last of five PRs splitting up #1746.

The automatic formats are walked in a fixed order, WebP first. Browsers that accept AVIF also accept WebP, so they always get WebP and `AUTO_AVIF = True` does nothing — that's #1688, and @peleteiro hit it in production at Bíblia Online ([his report on #1746](https://github.com/thumbor/thumbor/pull/1746#issuecomment-4957420967), with AVIF at −27% bytes for the same quality). Fixes #1688.

This adds `AUTO_IMAGE_FORMAT_PREFERENCE`: an ordered list of preferred automatic output formats. Valid values are `webp`, `avif`, `jpg`, `heif` and `png`; values are trimmed, lowercased and deduplicated, and invalid values are ignored with one warning. A list with at least one valid value overrides the individual `AUTO_*` format settings; an empty or fully invalid list keeps them, so existing setups are unaffected.

```python
# For example:
AUTO_IMAGE_FORMAT_PREFERENCE = ["avif", "webp", "jpg", "png", "heif"]
```

With a preference configured, thumbor goes through the list in order and picks the first format that the client says it accepts (its `Accept` header) and that the engine can produce. When no entry matches, `AUTO_PNG_TO_JPG` still applies as an independent fallback, and the engine extension is kept otherwise.

The same URL can now return different formats to different clients, and a cache that stores by URL alone would hand one client's format to another — e.g. a cached AVIF to a browser that can't show it. So thumbor sends [`Vary: Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) whenever any automatic format is active. Result storage is separated the same way (#1859): files created with a preference are stored apart from files created with the `AUTO_*` settings, so turning the preference on or off never serves a file created the other way.

⚠️ Upgrade note: turning the preference on (or off) starts with an empty result-storage namespace — every cached result is regenerated once. Old entries are not removed automatically.

@peleteiro, taking you up on the offer: this is the PR to test in production.

Big thanks, @marcelometal 🤘

### DIFF
--- a/docs/autojpg.rst
+++ b/docs/autojpg.rst
@@ -6,12 +6,15 @@ Usage: `autojpg(enabled)`
 Description
 -----------
 
-This filter overrides ``AUTO_PNG_TO_JPG`` config variable.
+This filter overrides only the ``AUTO_PNG_TO_JPG`` fallback for the current
+request. It does not enable or disable a ``jpg`` entry explicitly configured
+in ``AUTO_IMAGE_FORMAT_PREFERENCE``.
 
 Arguments
 ---------
 
--  enabled - Passing ``True``, which is the default value, you will override the ``AUTO_PNG_TO_JPG`` config variable and False to keep the default behavior of thus config.
+- ``enabled`` - ``True`` enables PNG-to-JPEG fallback and ``False`` disables
+  it for this request. The default is ``True``.
 
 Example
 -------
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -397,6 +397,12 @@ Images with texts, for example, the result image maybe will be distorted.
 Dark images, for example, the size of result image maybe will be bigger.
 You have to evaluate the majority of your use cases to take a decision about the usage of this conf.
 
+This conversion remains independent of ``AUTO_IMAGE_FORMAT_PREFERENCE``.
+When a preference list is configured and none of its entries can be used,
+thumbor still applies ``AUTO_PNG_TO_JPG`` as a fallback. The ``autojpg()``
+filter enables or disables this fallback for an individual request; it does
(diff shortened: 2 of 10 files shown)


marcelometal (approved):
Non-blocking note: The statement that "turning the preference off starts with an empty namespace" is guaranteed for the first rollout. If an operator later switches back to a policy that was used before, the previous namespace for that policy may already contain entries. This remains safe because preference and flags entries are never mixed; subsequent toggles may simply not start from a completely cold namespace.

The implementation looks good to me. Approved.

PR #1774, opened 2025-11-09
Fix #1711 — Preserve transparency when resizing transparent PNGs (PIL engine) and add regression test

HarshAggarwal1:
## 🧩 Summary
Fixes **#1711** — Transparent PNGs could lose or corrupt their alpha channel when resized (especially for very wide images with small height).  
This patch updates the **PIL engine** to correctly detect and preserve transparency during resizing.

---

## 🛠️ Changes
- Detect transparency **before** conversions and optimizations  
- Convert **palette/indexed** images (`P`, `1`) to safe RGB(A) before resampling  
- Force **RGBA** mode when alpha is present  
- Skip **JPEG-specific `draft()`** optimizations when alpha is present  
- Verify transparency **after resize** and restore RGBA if it was lost  
- Added regression test: `test_pil_engine_transparent_resize.py`

---

### Automated
- Regression test covers:
  - Transparent and semi-transparent PNGs  
  - Very wide aspect ratios  
  - Palette and RGBA image modes

### Manual Verification
1. Run only the new test:
   ```bash
   pytest -v tests/engine/test_pil_engine_transparent_resize.py
   ```
2. Manually resize a transparent PNG:
   ```python
   from PIL import Image
   img = Image.open("input.png").convert("RGBA")
   img.resize((2000, 50), Image.LANCZOS).save("out.png")
   ```
3. Verify `out.png` retains transparency in any viewer with a checkered background.

---

## ✅ CI & Review Checklist
- [x] All tests (including new regression) pass on CI  
- [x] Transparent PNGs retain alpha correctly  
- [x] No change in behavior for JPEG or opaque images  
- [x] No public API changes introduced

---

## 🧾 Changelog
**Fix:** Preserve transparency when resizing transparent PNGs with extreme aspect ratios.  
(*Fixes #1711*)

PR #1863, opened 2026-08-12
fix: serve grayscale over indexed PNGs instead of raising 500

### LINKED ISSUE #1862: [Bug] filters:grayscale() returns 500 on indexed PNGs: quantize() rejects LA mode
### Thumbor Request URL

```
https://example.com/unsafe/800x/filters:grayscale()/https://example.com/logo.png
```

where `logo.png` is a **palette PNG** (mode `P` with a `tRNS` chunk), which is what
logos exported from most design tools look like.

### Expected Behaviour

A grayscale PNG with its transparency preserved, the same as for an RGBA PNG or a JPEG.

### Actual Behaviour

HTTP **500 with an empty body**. In the logs:

```
File "thumbor/engines/pil.py", line 272, in read
    self.image = self.image.quantize(method=quantize_method)
File "PIL/Image.py", line 1346, in quantize
    im = self._new(self.im.quantize(colors, method, kmeans))
ValueError: image has wrong mode
```

### Operating System

Ubuntu 26.04, and Debian 12 inside the `python:3.12-slim` image.

### Your `thumbor.conf`

```python
ALLOW_UNSAFE_URL = True
LOADER = 'thumbor.loaders.https_loader'
FILTERS = ['thumbor.filters.grayscale']
# PILLOW_PRESERVE_INDEXED_MODE left at its default (True)
```

---

### Minimal reproduction

No network and no fixture needed:

```python
import io

from PIL import Image
from thumbor.config import Config
from thumbor.context import Context
from thumbor.engines.pil import Engine
from thumbor.importer import Importer

buffer = io.BytesIO()
Image.new("P", (120, 90)).save(buffer, format="PNG", transparency=0)

config = Config()
importer = Importer(config)
importer.import_modules()
engine = Engine(Context(config=config, importer=importer))

engine.load(buffer.getvalue(), ".png")
engine.resize(400, 300)        # P -> RGBA
engine.convert_to_grayscale()  # RGBA -> LA
engine.read(".png")            # ValueError: image has wrong mode
```

### Root cause

1. `resize()` widens indexed images to RGB(A) — `pil.py:190`.
2. `convert_to_grayscale()` returns mode **`LA`** whenever the image carries alpha —
   `pil.py:493`.
3. `read()` restores the original indexed mode when `PILLOW_PRESERVE_INDEXED_MODE` is
   enabled (the default), through `Image.quantize()` — `pil.py:253-272`.
4. Pillow's `quantize()` accepts `L`, `P`, `RGB` and `RGBA` — for `RGBA` it picks
   FASTOCTREE on its own — but rejects **`LA`**.

The failure therefore needs all three of an indexed original, a filter that leaves the
image in `LA`, and PNG/GIF output. That explains the whole observed matrix:

| original | filters | result |
| --- | --- | --- |
| PNG `P` + tRNS | `grayscale()` | **500** |
| PNG `RGBA` | `grayscale()` | 200 — not indexed, the branch is skipped |
| JPEG | `grayscale()` | 200 — never reaches step 3 |
| PNG `P` + tRNS | `format(jpeg):grayscale()` | 200 — changes the requested extension |
| PNG `P` + tRNS | `blur(30)`, `rotate(90)`, `equalize()`, … | 200 — they leave `RGBA` |

Of the built-in filters, `grayscale` is the only one that leaves the image in `LA`
(checked one by one over an indexed PNG). Animated GIFs go through `GIF_ENGINE`
(gifsicle `--use-colormap gray`), a path that never calls `quantize()`, and are not
affected.

### Versions

- thumbor **7.8.0** (latest release) and **`master` @ 27a8afb5** — the code is identical
  in both, and the reproduction above fails on both.
- Pillow 12.3.0 (wheel, so no libimagequant: `quantize_method` ends up `None` →
  MEDIANCUT).
- Not covered by #1774: that one only touches `resize()`. Applied on top of `master`,
  the failure is unchanged.

### Workarounds

Either prepend `format(jpeg)` (loses the transparency) or set
`PILLOW_PRESERVE_INDEXED_MODE = False`, which stops the 500 but drops palette output
for **every** indexed PNG: on a 800px-wide logo, 63 KB → 378 KB here.

A PR with a three-line fix and a regression test follows.

diegosuarez:
Fixes #1862.

`filters:grayscale()` over a palette PNG answered **500 with an empty body**:

1. `resize()` widens indexed images to RGB(A).
2. `convert_to_grayscale()` returns mode **`LA`** when the image carries alpha.
3. `read()` restores the original indexed mode via `Image.quantize()` when
   `PILLOW_PRESERVE_INDEXED_MODE` is on (the default).
4. Pillow's `quantize()` takes `L`, `P`, `RGB` and `RGBA` — for `RGBA` it picks
   FASTOCTREE itself — but rejects `LA` with `ValueError: image has wrong mode`.

Widening `LA` to `RGBA` right before quantizing keeps the palette, the grey pixels and
the alpha channel, so the output is still the indexed PNG the setting asks for. No
flattening, no forced JPEG.

Only `grayscale` leaves the image in `LA` among the built-in filters, and GIFs use
`GIF_ENGINE`, so this is the whole surface. It applies to `master` and to 7.8.0 alike,
and #1774 does not cover it (it only touches `resize()`).

### Verification

The new test is red before the change and green after:

```
$ git stash && pytest tests/engines/test_pil.py -k after_grayscale
    self.image = self.image.quantize(method=quantize_method)
ValueError: image has wrong mode
FAILED (errors=1)

$ git stash pop && pytest tests/engines/test_pil.py -k after_grayscale
Ran 1 tests in 0.01s
OK
```

Full file, and the wider engine/filter selection with no new failures (the 10 that fail
do so identically on unmodified `master` here, for want of a local `gifsicle` and an SVG
converter):

```
$ pytest tests/engines/test_pil.py
Ran 30 tests in 0.54s
OK (skipped=5, expected failures=2)

$ pytest tests/engines/ tests/filters/     # this branch
Ran 180 tests ... errors=5, failures=5
$ pytest tests/engines/ tests/filters/     # master
Ran 179 tests ... errors=5, failures=5     # same names, pre-existing
```

Style gates:

```
$ black --check thumbor/engines/pil.py tests/engines/test_pil.py   # unchanged
$ isort --check-only ...                                           # clean
$ flake8 --config .flake8 ...                                      # clean
$ pylint --rcfile=pylintrc ...                                     # 10.00/10
```

The test reuses the existing `tests/fixtures/images/paletted-transparent.png` fixture
(mode `P` + `tRNS`), so no new binary is added, and it sits next to
`test_should_preserve_png_transparency`, which covers the same fixture without the
filter.

### DIFF
--- a/tests/engines/test_pil.py
+++ b/tests/engines/test_pil.py
@@ -15,7 +15,7 @@ from unittest import TestCase, mock
 
 import piexif
 import pytest
-from PIL import Image
+from PIL import Image, ImageChops
 from preggy import expect
 
 from tests.base import (
@@ -280,6 +280,12 @@ class PilEngineTestCase(TestCase):
 
         expect(img.mode).to_equal("P")
         expect(img.format.lower()).to_equal("png")
+
+        # Every pixel must be grey, so that neither a no-op grayscale nor
+        # colour reintroduced by the quantization goes unnoticed
+        red, green, blue = img.convert("RGB").split()
+        expect(ImageChops.difference(red, green).getextrema()[1]).to_equal(0)
+        expect(ImageChops.difference(green, blue).getextrema()[1]).to_equal(0)
 
         transparent_pixels_count = sum(
             img.convert("RGBA")
--- a/tests/engines/test_pil.py
+++ b/tests/engines/test_pil.py
@@ -262,6 +262,34 @@ class PilEngineTestCase(TestCase):
         # Image has total of 200x150=30000 pixels. Most of them should be transparent
         expect(transparent_pixels_count).to_be_greater_than(19000)
 
+    def test_should_preserve_png_transparency_after_grayscale(self):
+        engine = Engine(self.context)
+
+        with open(
+            join(STORAGE_PATH, "paletted-transparent.png"), "rb"
+        ) as image_file:
+            buffer = image_file.read()
+
+        engine.load(buffer, "png")
+        expect(engine.original_mode).to_equal("P")
+        engine.resize(200, 150)
+        # Leaves the image in LA mode, which quantize() does not accept
+        engine.convert_to_grayscale()
+
+        img = Image.open(BytesIO(engine.read(".png")))
+
+        expect(img.mode).to_equal("P")
+        expect(img.format.lower()).to_equal("png")
+
+        transparent_pixels_count = sum(
+            img.convert("RGBA")
+            .split()[3]
+            .point(lambda x: 0 if x else 1)
+            .getdata()
+        )
+
+        expect(transparent_pixels_count).to_be_greater_than(19000)
+
     @skip_unless_avif
     def test_convert_jpg_to_avif(self):
         engine = Engine(self.context)
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -262,10 +262,11 @@ class Engine(BaseEngine):
             if self.original_mode == "1":
                 self.image = self.image.convert("1")
             else:
-                # quantize() only takes L, P, RGB and RGBA. Filters that
-                # convert to grayscale leave images carrying alpha in LA
-                # mode, which it rejects with "image has wrong mode".
-                if self.image.mode == "LA":
+                # quantize() only takes L, P, RGB and RGBA, and rejects
+                # every other mode with "image has wrong mode". Filters are
+                # pluggable and can leave any of them behind: grayscale(),
+                # for one, leaves images carrying alpha in LA mode.
+                if self.image.mode not in ("L", "P", "RGB", "RGBA"):
                     self.image = self.image.convert("RGBA")
 
                 # libimagequant might not be enabled on compile time
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -262,6 +262,12 @@ class Engine(BaseEngine):
             if self.original_mode == "1":
                 self.image = self.image.convert("1")
             else:
+                # quantize() only takes L, P, RGB and RGBA. Filters that
+                # convert to grayscale leave images carrying alpha in LA
+                # mode, which it rejects with "image has wrong mode".
+                if self.image.mode == "LA":
+                    self.image = self.image.convert("RGBA")
+
                 # libimagequant might not be enabled on compile time
                 # but it's better than default octree for RGBA images
                 quantize_method = (


tests/engines/test_pil.py
```diff
@@ -262,6 +262,34 @@ def test_should_preserve_png_transparency(self):
...
+        expect(engine.original_mode).to_equal("P")
+        engine.resize(200, 150)
+        # Leaves the image in LA mode, which quantize() does not accept
+        engine.convert_to_grayscale()
+
+        img = Image.open(BytesIO(engine.read(".png")))
+
+        expect(img.mode).to_equal("P")
+        expect(img.format.lower()).to_equal("png")
+
```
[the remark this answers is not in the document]
diegosuarez: Good catch, added in 34109fd. The channel difference turns out to be exactly 0 (palette entries included), so the assertion is strict rather than approximate. Checked that it fails if the `convert_to_grayscale()` call is removed, so it does cover the no-op case.

thumbor/engines/pil.py
```diff
@@ -262,6 +262,12 @@ def read(  # noqa
             if self.original_mode == "1":
                 self.image = self.image.convert("1")
             else:
+                # quantize() only takes L, P, RGB and RGBA. Filters that
+                # convert to grayscale leave images carrying alpha in LA
+                # mode, which it rejects with "image has wrong mode".
+                if self.image.mode == "LA":
+                    self.image = self.image.convert("RGBA")
```
scorphus: `quantize()` takes `L`, `P`, `RGB` and `RGBA` and rejects every other mode with this same error. Today the only mode that can reach this line is the `LA` that `grayscale()` leaves behind, but filters are pluggable, and a custom one can leave any mode. I'd check against the four modes instead of the one:

```suggestion
                if self.image.mode not in ("L", "P", "RGB", "RGBA"):
                    self.image = self.image.convert("RGBA")
```

Please, feel free to adjust the comment above that `if` as you see fit.

scorphus (changes requested):
Thanks for the careful report and fix, @diegosuarez — and for strengthening the test on Copilot's nudge. The fix is in the right place: the indexed-output promise lives in `read()`, so the repair belongs at the `quantize()` call, not in the filter. Great find!

One request, inline. With that in, this is good to go — I'll squash when merging, so no need to rework the branch.

thumbor/engines/pil.py
diegosuarez: Thank you, that is a much better way to frame it — you are right that the pluggable filters make the narrow check a trap waiting to happen. Applied in 771c2ef, with the comment above reworded to match, and credited you as co-author. Tests and lint still green.

scorphus (approved):
Thanks for your contribution, @diegosuarez!

PR #1877, opened 2026-08-15
Validate security key paths before reading

marcelometal:
Resolve key file paths against an allowed directory before opening them, rejecting traversal and symlink escapes. Add regression coverage for relative traversal, valid absolute paths, and symlink targets.

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -7,6 +7,8 @@
 # http://www.opensource.org/licenses/mit-license
 # Copyright (c) 2011 globo.com thumbor@googlegroups.com
 
+from pathlib import Path
+from tempfile import TemporaryDirectory
 from unittest import TestCase, mock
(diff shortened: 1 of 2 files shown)


scorphus (changes requested):
Two things, both small:

- The new tests use preggy, and #1801 removes preggy from the whole suite — since both are yours, writing these with `pytest.raises` now saves the future conflict. Suggestions inline; `match=` is a regex search, so I kept exact message comparison instead.
- The symlink rejection is a behavior change worth one line in the PR body or release notes: a key file symlinked to outside its directory now fails at startup. Kubernetes-style mounts keep working (their symlinks stay inside the mount), but hand-made symlinks to elsewhere stop.

The validation logic itself looks right — traversal caught, absolute paths anywhere still fine.

PR #1879, opened 2026-08-15
fix(pil): detect tRNS transparency, not just alpha modes

scorphus:
Deciding on transparency by looking for "A" in the mode misses images that store it in a tRNS chunk instead of an alpha channel: palette PNGs, and grayscale or truecolor PNGs with a colorkey. Two spots did that:

- `convert_to_grayscale()` flattened those images to opaque `L` — `filters:grayscale()` over a transparent palette PNG, no resize involved, returned the image with its transparency gone, silently.
- `image_data_as_rgb()` handed filters `RGB` data with the colorkey dropped — and its old gate never let a truecolor image with a colorkey enter the conversion at all.

Both now decide with `has_transparency()`, which also checks tRNS.

Four tests, all comparing alpha bytes exactly: palette PNG through grayscale, gray and truecolor colorkey PNGs through `image_data_as_rgb()`, and the no-resize path through `read()`.

### DIFF
--- a/thumbor/engines/pil.py
+++ b/thumbor/engines/pil.py
@@ -483,13 +483,13 @@ class Engine(BaseEngine):
     def image_data_as_rgb(self, update_image=True):
         converted_image = self.image
 
-        if converted_image.mode not in ["RGB", "RGBA"]:
-            if "A" in converted_image.mode:
+        if converted_image.mode != "RGBA":
+            if self.has_transparency():
                 converted_image = converted_image.convert("RGBA")
             elif converted_image.mode == "P":
                 # convert() figures out RGB or RGBA based on palette used
                 converted_image = converted_image.convert(None)
-            else:
+            elif converted_image.mode != "RGB":
                 converted_image = converted_image.convert("RGB")
 
         if update_image:
@@ -498,7 +498,7 @@ class Engine(BaseEngine):
         return converted_image.mode, converted_image.tobytes()
 
     def convert_to_grayscale(self, update_image=True, alpha=True):
-        if "A" in self.image.mode and alpha:
+        if alpha and self.has_transparency():
             image = self.image.convert("LA")
         else:
             image = self.image.convert("L")
--- a/tests/engines/test_pil.py
+++ b/tests/engines/test_pil.py
@@ -290,6 +290,77 @@ class PilEngineTestCase(TestCase):
 
         assert transparent_pixels_count > 19000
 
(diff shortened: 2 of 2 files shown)


thumbor/engines/pil.py
```diff
@@ -483,13 +483,13 @@ def get_image_mode(self):
     def image_data_as_rgb(self, update_image=True):
         converted_image = self.image
 
-        if converted_image.mode not in ["RGB", "RGBA"]:
-            if "A" in converted_image.mode:
+        if converted_image.mode != "RGBA":
+            if self.has_transparency():
```
marcelometal, line 487: This check happens after `Transformer` has already resized the image. `Engine.resize()` only converts `1`/`P` inputs before resampling, so `L`/`RGB` PNGs with a tRNS colorkey are still resampled in their non-alpha mode. For example, an `L` image with pixels `[128, 255]` and `transparency=128` becomes `[106, 158, 225, 255]` when resized from 2 to 4 pixels; no pixel then matches the colorkey, `has_transparency()` returns false, and this returns opaque RGB. Converting to RGBA before resizing yields alpha `[0, 59, 196, 255]`. Could we handle `L`/`RGB` tRNS inputs in `resize()` and add resized cases to these tests?

thumbor/engines/pil.py
```diff
@@ -498,7 +498,7 @@ def image_data_as_rgb(self, update_image=True):
         return converted_image.mode, converted_image.tobytes()
 
     def convert_to_grayscale(self, update_image=True, alpha=True):
-        if "A" in self.image.mode and alpha:
+        if alpha and self.has_transparency():
```
marcelometal, line 501: `has_transparency()` is substantially more expensive than the old mode check: it converts the full image to RGBA, extracts and scans the alpha channel, and then this method converts the full image again. That scan is unnecessary when choosing an alpha-capable destination; checking for an alpha-bearing mode or `"transparency" in image.info` is sufficient and also preserves the previous behavior for fully opaque `LA`/`PA` images. On a 36 MP `LA` image, I measured direct RGBA conversion at 0.035 s / 302 MB max RSS versus 0.166 s / 337 MB for the scan plus conversion. Could we use a lightweight “has alpha representation” predicate here and in `image_data_as_rgb()`?

marcelometal (commented):
I found two issues around resized tRNS inputs and the cost/semantics of using `has_transparency()` as the conversion predicate. Details are inline.

Issue #1900, opened 2026-09-07
Docker : how to configure through env vars ?


franck-grenier:
Hello Thumbor,
I'm tring to setup my Thumbor service with Docker Compose and I would like to override configuration through env vars.

Unfortunately, it does not work.

Here is my Docker Compose config:

```
services:
  site-thumbor:
    image: ghcr.io/thumbor/thumbor:7-py-3.14
    ports:
      - 8888:8888
    environment:
      - UPLOAD_ENABLED=True
      - UPLOAD_DELETE_ALLOWED=True
      - UPLOAD_PUT_ALLOWED=True
```

The service starts correctly but upload is not enabled. I get 405 error when I POST on /image API. PUT does not work also. Env vars do exist in my container.

I tried `THUMBOR_` prefix on vars but no changes.

If I update directly `thumbor.conf`, it works.

Does ghcr.io/thumbor/thumbor:7-py-3.14 image allow overriding config with env vars ? 
What is the good syntax ?

Thanks

marcelometal:
Hi @franck-grenier,

Environment overrides are disabled by default. To enable them, add this to your Compose service:

```yaml
command: ["--use-environment=True"]
```

The current CLI requires a value for this option. Your variable names are correct: no additional `THUMBOR_` prefix is needed. See the [CLI definition](https://github.com/thumbor/thumbor/blob/master/thumbor/console.py).

However, the current environment override mechanism returns **strings**, without converting booleans, integers or lists. The [configuration documentation](https://github.com/thumbor/thumbor/blob/master/docs/configuration.rst) describes support for string settings. In particular, `"False"` remains a nonempty string, so it should not be used to disable boolean options.

For your upload settings, I recommend explicitly parsing the environment variables in a mounted `thumbor.conf`:

```python
import os

def env_bool(name):
    return os.environ.get(name, "false").strip().lower() == "true"

UPLOAD_ENABLED = env_bool("UPLOAD_ENABLED")
UPLOAD_DELETE_ALLOWED = env_bool("UPLOAD_DELETE_ALLOWED")
UPLOAD_PUT_ALLOWED = env_bool("UPLOAD_PUT_ALLOWED")
```

Keep your existing environment variables and add:

```yaml
command: ["-c", "/etc/thumbor/thumbor.conf"]
volumes:
  - ./thumbor.conf:/etc/thumbor/thumbor.conf:ro
```

[...]

PR #1882, opened 2026-08-25
docs: convert Sphinx sources to MyST Markdown

marcelometal:
## Summary

- convert the Sphinx documentation sources from reStructuredText to MyST Markdown
- configure myst-parser and update the locked documentation dependencies
- format Markdown sources and tables with an 80-column limit
- fix rendered escapes, internal references, wildcard markup, and API directives

## Validation

- strict Sphinx HTML build with warnings treated as errors
- audited all 79 generated HTML pages for visible RST and escaped Markdown artifacts
- verified local HTML targets and anchors
- checked line length, trailing whitespace, and final newlines

PR #1903, opened 2026-09-07
docs: add Compose examples and environment value conversion

marcelometal:
Add complete Docker Compose examples for string environment overrides and for a mounted `thumbor.conf` that converts environment values to booleans and integers. Explain why `"False"` does not disable boolean settings and why `--use-environment=False` still enables overrides. Preserve the configuration section anchor referenced in #1900.

This PR extends #1891 and targets its branch, `docs/fix-config-runtime-docs`. That PR depends on the Markdown migration in #1882. The diff contains only the additions to `docs/configuration.md` and `docs/hosting.md`; merge after those dependencies.

Validation:

- Sphinx HTML build passes with `-W --keep-going -E -a` (no warnings).
- Both Compose examples pass `docker compose config`; their CLI arguments and converted configuration values were checked against the runtime.
- Checked shell syntax, rendered anchors and cross-references, and Markdown line lengths.
- `make unit` passes (9 skips and 2 expected failures); `make flake isort pylint` and pre-commit checks pass.

Refs #1900.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,6 +25,7 @@ thumbor-config > ./thumbor.conf
 This command writes to standard output. Shell redirection creates or replaces
 the destination file; it does not merge with an existing configuration.
 
+(override-config-through-environment-variable)=
 ## Override configuration through environment variables
 
 Environment overrides are disabled by default. To enable the legacy derpconf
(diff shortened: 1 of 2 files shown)


franck-grenier:
Hi @marcelometal ,
you still have references to `THUMBOR_` prefix in current documentation: https://thumbor.readthedocs.io/en/latest/hosting.html#configuring-the-docker-image

marcelometal:
Thanks @franck-grenier. The page you linked is built from `master`, where `docs/hosting.rst` still carries the `THUMBOR_` prefix example. That text is already removed in #1891 (commit 50cf3f12, "docs: align Docker configuration with image"), which this PR is stacked on: the Docker section now states that the official image does not translate `THUMBOR_*` variables and points to `--use-environment` with unprefixed names. The published docs will update once #1882, #1891 and this PR are merged and Read the Docs rebuilds.

